PR #61, opened 2020-12-11
Fix typos: self -> this

facelessuser:


LeaVerou:
Thank you! But wouldn't `this` point to `Color` there and not the global object?
Perhaps `typeof CSS !== "undefined"` might be better?

facelessuser:
In that case, maybe so. After I found the first `self`, I just did a find and replace for the other, I honestly didn't look at the second case too close 😅.

LeaVerou:
Yes, this fixes the issue with Node, but it also completely changes the meaning of those loc. It's just that Node doesn't need those loc. 😄

facelessuser:
Looking closer, no one ever uses `this.Proxy` either. I'm not entirely sure what is being checked there either.

LeaVerou:
On the client-side, `self` is just another way to say `window`. The former exists in Web workers too, but not the latter, so I've gotten into the habit to prefer the former.
Checking for `self.Proxy` just checks for Proxy support :)

facelessuser:
Haha, well I dabble in some web stuff, but clearly, I'm not a web dev, so this was a little new to me. I made a very big assumption which is very much incorrect 😆.

facelessuser:
I've adjusted the fix based on your statements. If I am still off, feel free to let me know. I also won't feel bad if you find it easier to just fix it directly sidestepping this pull 😄.

LeaVerou:
Thank you for the fast turnaround! You need `===` instead of `!==` for actual equivalence with the previous code, now you're checking for the opposite :)

facelessuser:
🤦 Third times a charm!

LeaVerou:
Merged, thanks!

PR #62, opened 2020-12-11
chore: use conditional package.json exports

kripod:
As [suggested here](https://github.com/LeaVerou/color.js/issues/55#issuecomment-742808331), I've added CommonJS-specific conditional exports.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,12 +4,12 @@
 	"description": "Color space agnostic color manipulation library",
 	"main": "color.js",
 	"exports": {
-		"import": "./color.js",
+		"import": "./dist/color.esm.js",
 		"require": "./dist/color.cjs.js",
 		"default": "./dist/color.js"
 	},
 	"main": "./dist/color.cjs.js",
-	"module": "./color.js",
+	"module": "./dist/color.esm.js",
 	"directories": {
 		"test": "tests"
 	},
--- a/package.json
+++ b/package.json
@@ -3,6 +3,13 @@
 	"version": "0.0.3",
 	"description": "Color space agnostic color manipulation library",
 	"main": "color.js",
+	"exports": {
+		"import": "./color.js",
+		"require": "./dist/color.cjs.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3,6 +3,13 @@
 	"version": "0.0.3",
 	"description": "Color space agnostic color manipulation library",
 	"main": "color.js",
+	"exports": {
+		"import": "./color.js",
```
LeaVerou, line 7: Please use `"./dist/color.esm.js:` here, `color.js` on the root is primarily for use by the color.js website, and assigns Color to a global (so that people can play in the console) which kind of defeats the purpose of modules if used generally.

package.json
```diff
@@ -3,6 +3,13 @@
...
+		"require": "./dist/color.cjs.js",
+		"default": "./dist/color.js"
+	},
+	"main": "./dist/color.cjs.js",
```
LeaVerou, line 11: Shouldn't this be the ESM one?

package.json
```diff
@@ -3,6 +3,13 @@
...
+	"module": "./color.js",
```
LeaVerou, line 12: Same as above.

LeaVerou (changes requested):
Thank you! See comments.

package.json
kripod: No, as older non-ESM-compatible Node versions only look for this field when trying to resolve the module.

LeaVerou:
Merged, thanks!

LeaVerou:
Do you want to try the new code out manually so we can fix any remaining issues before I draft a new release?

kripod:
@LeaVerou I’m busy at work right now and it would be easier to try the new fixes as an npm release for me. Sorry for the inconveniences caused by this.

kripod:
@LeaVerou I’ve just tried the changes locally and things seems to work fine for now. May I ask if drafting a new npm release is possible anytime? (Take your time 😊)

PR #65, opened 2021-01-23
Fix algorithmic asymmetry

### LINKED ISSUE #64: Round trip of sRGB seems wrong
```
let color = new Color("color(srgb -0.3 0.5 1)").to('xyz').to('srgb')
```

result

```
[-0.41,0.5,1]
```

Maybe I'm missing what this change is trying to solve, but this doesn't seem right and seems to be causing the problem: https://github.com/LeaVerou/color.js/commit/efe60a84ee567e3fb7b8fb7343c45bac89de3085.

Without the change, it would round trip properly to `[-0.3,0.5,1]`.

facelessuser:
Fixes #64

### DIFF
--- a/src/spaces/srgb.js
+++ b/src/spaces/srgb.js
@@ -34,7 +34,7 @@ Color.defineSpace({
 			let abs = Math.abs(val);
 
 			if (abs > 0.0031308) {
-				return sign * 1.055 * Math.pow(abs, 1/2.4) - 0.055;
+				return sign * (1.055 * Math.pow(abs, 1/2.4) - 0.055);
 			}
 
 			return 12.92 * val;


facelessuser:
I assume this change would need to get updated in the CSS spec as well.

svgeesus (approved):

svgeesus:
> I assume this change would need to get updated in the CSS spec as well.

Yes!

svgeesus:
[also fixed in CSS Color 4 sample code](https://github.com/w3c/csswg-drafts/commit/dd272a51a4cafc44ed7cf574784ee7bffc1de5ec)

(Two of the four changes are for clarity, not correctness)

PR #68, opened 2021-01-29
Fix HSV to HSL conversion

facelessuser:
Do not scale lightness back to 0 - 100 until conversions are done.

Before fix:

```js
>  new Color("color(hsv 60 50 40)").to('srgb').to('hsv').toString()
'color(hsv 60.001 2.021 30.306)'
```

After fix:

```js
>  new Color("color(hsv 60 50 40)").to('srgb').to('hsv').toString()
'color(hsv 60 50 40)'
```

### DIFF
--- a/src/spaces/hsv.js
+++ b/src/spaces/hsv.js
@@ -44,12 +44,12 @@ Color.defineSpace({
 			s /= 100;
 			v /= 100;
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
LGTM, thanks!

Issue #72, opened 2021-03-02
inGamut check for HSL and other sRGB cylindrical spaces


facelessuser:
## Problem Description

I don't think the conversion is a bug, but I am more questioning the `inGamut` check, though I understand what it is doing, and technically it isn't wrong. I am more curious if this is known and accepted, or unknown and not yet considered. Consider the following:

```js
> let color = new Color("lab(100% 0 0)").to('hsl')
undefined
> color.toString()
'hsl(32.624 493.346% 99.998%)'
> color.inGamut()
true
```

We can see in this example that saturation, after conversion, is well past the accepted constraints, and yet, this HSL color is considered "in gamut".

Now, I know that HSL is not checked for "in gamut" by comparing its own coords, but is actually checked in the sRGB space with an allowable threshold. So, on the one hand, this makes sense:

```js
> let color2 = new Color("lab(100% 0 0)").to('srgb')
undefined
> color2.toString()
'rgb(100.01% 99.999% 99.989%)'
```

As can be noted above, the conversion directly to sRGB is far less jarring.

Now, I understand that the conversion algorithm for HSL is designed for values within the constraints of a color that is in gamut, and once colors are out of gamut (even if they are only slightly out of gamut) may not yield values a user expects, and that is not necessarily wrong. I am more interested in the idea of whether the HSL value should actually be considered "in gamut" even if it is technically not too far out of the sRGB gamut. In HSL, the color is wildly out of range.

The reason why this seems problematic is if I am trying to well form my cylindrical sRGB values, they do not actually register as out of bounds with the `inGamut` check. If you run `toGamut('srgb')`, the value is fit without the "tolerance" allowed before, and you will actually get a sane value. You just have no way of testing that it is out of bounds without manual introspection:

```js
> color.toString()
'hsl(32.624 493.346% 99.998%)'
> color.inGamut()
true
> color.toGamut('srgb').toString()
'hsl(54.251 100% 99.995%)'
```

Surprisingly, `toGamut('hsl')` does not seem to redirect to the `srgb` logic either.

```js
> color.toGamut('hsl').toString()
'hsl(32.624 493.346% 99.998%)'
```

## Question

So my question is: should HSL also evaluate its own coords in addition to the sRGB constraint check? Or is this a known and accepted behavior?

I know that HSL is simply a representation of the sRGB color, and the gamut is the same, but it also seems that values that are wildly out of range should also be flagged as "out of gamut". I also think that `toString` should gamut map the HSL color in this case.

[...]

PR #73, opened 2021-03-03
Pass epsilon to spaces that redirect the inGamut check

facelessuser:
This fixes an issue where when an epsilon threshold is passed to `inGamut`, it is not sent to spaces that implement their own `inGamut`.

Previously, when calling `toGamut`, colors like HSL would not actually be fit to the gamut:

```js
> color = new Color("lab(100% 0 0)").to('hsl')
Color$1 {
  _spaceId: 'hsl',
  toString: [Function: toString],
  coords: [ 32.6238164048059, 493.3464693494261, 99.99822243865972 ],
  alpha: 1
}
> color.toGamut()
Color$1 {
  _spaceId: 'hsl',
  toString: [Function: toString],
  coords: [ 32.6238164048059, 100, 99.99822243865972 ],
  alpha: 1
}
```

Though, I think this exposes another potential issue. When HSL has `inGamut` called on it, it uses the sRGB color space to confirm whether it is in gamut or not, but when `toGamut` is called on it, it then uses HSL. Shouldn't it also use sRGB to fit it in gamut?

```js
> color.toGamut('srgb')
Color$1 {
  _spaceId: 'hsl',
  toString: [Function: toString],
  coords: [ 54.25108421948211, 99.99999999986527, 99.99472645127346 ],
  alpha: 1
}
```

Should `toGamut` use sRGB or HSL? It seems the check and the fitting are inconsistent.

Address at least one issue found in Issue #72

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -258,7 +258,7 @@ export default class Color {
 		space = Color.space(space);
 
 		if (space.inGamut) {
-			return space.inGamut(coords);
+			return space.inGamut(coords, epsilon);
 		}
 		else {
 			if (!space.coords) {
--- a/src/spaces/hsl.js
+++ b/src/spaces/hsl.js
@@ -9,9 +9,9 @@ Color.defineSpace({
 		saturation: [0, 100],
 		lightness: [0, 100]
 	},
-	inGamut (coords) {
+	inGamut (coords, epsilon) {
 		let rgb = this.to.srgb(coords);
-		return Color.inGamut("srgb", rgb);
+		return Color.inGamut("srgb", rgb, {epsilon: epsilon});
 	},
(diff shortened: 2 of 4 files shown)


facelessuser:
Thinking about this more, I guess there might not be a problem. It is possible that I am making an assumption that HSL needs to be sized in the sRGB space simply because the gamut check is done in the sRGB space, but I guess maybe that isn't true. If it is important to someone, they can always feed in `srgb` as the space. I was probably overthinking this. Regardless, if it is a problem, it is separate from the issue being solved in this pull 🙂.

svgeesus:
Certainly, the epsilon if present should be propagated on.

PR #79, opened 2021-03-22
Fix conversion issues with HWB and HSL

### LINKED ISSUE #74: HSL algorithm can divide by zero
Is there a reason the HSL algorithm does not implement the part relating to when `L == 0 or L == 1` found on the wiki: https://en.wikipedia.org/wiki/HSL_and_HSV#From_RGB?

The `L == 1` is the part that is of particular concern as it causes a divide by zero:

```js
let color = new Color("srgb", [2, 0, -2]).to('hsl');
```

Will yield: 

```css
hsl(30 Infinity% 0%)
```

I'm not sure if this is an oversight or purposeful, and a clean methodology for handling the case has not been outlined yet.

EDIT: Correction, both these cases can cause division by zero.

facelessuser:
1. HSL: Avoid divide by zero cases
2. HSL: values are in range [0, 1] and scale to 100% at end
3. HWB: Avoid divide by zero cases in HWB
4. HWB: values are in range [0, 1] and scale to 100% at end
5. HWB: Fix case where achromatic HWB colors were not scaled correctly
   when returned as HSV.

Fixes #74

### DIFF
--- a/src/spaces/hwb.js
+++ b/src/spaces/hwb.js
@@ -78,12 +78,12 @@ Color.defineSpace({
 			let sum = w + b;
 			if (sum >= 1) {
 				 let gray = w / sum;
-				 return [h, 0, gray];
+				 return [h, 0, gray * 100];
 			}
 
-			let v = 1 - b;
-			let s = 100 - (100 * w) / (100 - b);
-			return [h, s, v * 100];
+			let v = (1 - b);
+			let s = (v === 0) ? 0 : 1 - w / v;
+			return [h, s * 100, v * 100];
 		},
 
 		hsl (hwb) {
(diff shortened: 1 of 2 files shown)


src/spaces/hwb.js
```diff
@@ -78,12 +78,12 @@ Color.defineSpace({
 			let sum = w + b;
 			if (sum >= 1) {
 				 let gray = w / sum;
-				 return [h, 0, gray];
+				 return [h, 0, gray * 100];
```
svgeesus, line 81: Oh wow yeah (embarrassed)

svgeesus (approved):
Yes this looks great, thanks!

PR #80, opened 2021-03-22
Fix comparison inconsistencies

facelessuser:
ProPhoto: https://en.wikipedia.org/wiki/ProPhoto_RGB_color_space#Encoding_function
Rec.2020: https://en.wikipedia.org/wiki/Rec._2020#Transfer_characteristics


Artoria2e5:
Rec. 2020 is probably fine since the value is chosen to be continuous. No idea about the other guy. (The human eye would be able to tell neither, but I agree sticking to the spec is a good idea.)

facelessuser:
Yeah, it is purely a "correctness" issue in regards to the spec. I don't imagine there would be any real world issues.

LeaVerou:
Thank you @facelessuser!

PR #83, opened 2021-03-28
Acquire luminance from XYZ D65 to match WCAG 2.0 more correctly

### LINKED ISSUE #82: Contrast ratio seems wrong
It seems that Color.js, when calculating relative luminance, uses the Y coordinate of the XYZ color space, but that XYZ color space uses D50.

```js
> let color = new Color("#343e58").contrast("#ca6c87");
3.095
```

This doesn't seem to match every other implementation. For instance, D3 has a whole page explaining contrast and goes through all the math: https://observablehq.com/@mbostock/wcag-contrast-ratio. They get `3.039602551762186`.

This seems to align closer to an XYZ color space using D65. I made an adjustment to my local library to use D65, and get something that aligns much better with every other WCAG contrast ration implementation.

```py
>>> Color("#343e58").contrast("#ca6c87")
3.039797627037577
```

Is this a mistake/bug in color.js?  Or is this a new direction that is purposefully being taken? It is very unclear as it states it calculates WCAG 2.1 color contrast, but it seems it actually does not. Or maybe I'm just confused?

facelessuser:
Fixes #82

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -194,11 +194,17 @@ export default class Color {
 
 	// Relative luminance
 	get luminance () {
-		return this.xyz.Y;
+		// Luminance should actually be retrieved from XYZ with a D65 white point.
+		return Color.chromaticAdaptation(Color.spaces.xyz.white, Color.whites.D65, this.xyz)[1];
 	}
 
 	set luminance (value) {
-		this.xyz.Y = value;
+		let xyz = Color.chromaticAdaptation(Color.spaces.xyz.white, Color.whites.D65, this.xyz);
+		xyz[1] = value;
+		xyz = Color.chromaticAdaptation(Color.whites.D65, Color.spaces.xyz.white, xyz);
+		this.xyz.X = xyz[0];
+		this.xyz.Y = xyz[1];
+		this.xyz.Z = xyz[2];
 	}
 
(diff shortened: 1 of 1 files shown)


facelessuser:
It didn't seem to make sense adding a XYZ D65 color space just for this. But this to match much better now.

Before:

```js
> new Color("#343e58").contrast("#ca6c87")
3.094994424224814
```

After:

```js
> new Color("#343e58").contrast("#ca6c87")
3.0397976890682963
```

External reference: https://observablehq.com/@mbostock/wcag-contrast-ratio.

svgeesus (approved):

src/color.js
```diff
@@ -194,11 +194,17 @@ export default class Color {
...
 	}
 
 	set luminance (value) {
-		this.xyz.Y = value;
+		let xyz = Color.chromaticAdaptation(Color.spaces.xyz.white, Color.whites.D65, this.xyz);
+		xyz[1] = value;
+		xyz = Color.chromaticAdaptation(Color.whites.D65, Color.spaces.xyz.white, xyz);
+		this.xyz.X = xyz[0];
+		this.xyz.Y = xyz[1];
+		this.xyz.Z = xyz[2];
```
LeaVerou, line 207: You should be able to just set `this.xyz = xyz`, did that not work?

LeaVerou (approved):

PR #88, opened 2021-04-15
docs: minor non-content fixes in color-difference.md

micgro42:
I noticed a typo and stray comment in color-difference.md and my IDE also noticed an unmatched bracket :slightly_smiling_face:


micgro42:
Also, it would seem that the two colors in the code block where the comment was removed are supposed to be different, but they are the same.
From the intro-text to that code block, it would seem that they are supposed to differ by ca. 2.3, but since all three algorithms produce different results, I'm hesitant to suggest a color here myself.

LeaVerou:
Thanks for spotting that!

@svgeesus What's up with the example?

PR #96, opened 2021-06-09
color.js: Do not use window when it does not exist

Artoria2e5:
Right now color.js seems to... import poorly in node via `esm` due to this use of `window`. I could change it to `global` so it works everywhere, but I figured it won't be much use in non-browser environments anyways.

### DIFF
--- a/color.js
+++ b/color.js
@@ -2,7 +2,9 @@
 // for easier testing and experimentation without building
 import Color, {util} from "./src/main.js";
 
-window.Color = window.Color || Color;
+if (window) {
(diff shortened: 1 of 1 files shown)


LeaVerou:
You are not meant to import that file, that's primarily for quick testing. 
The [actual entry point](https://github.com/LeaVerou/color.js/blob/master/src/main.js) does not use `window`.
We should definitely document this better!

Please note that `global` is not available in browser contexts, and thus would produce an error. You can confirm this by opening your dev tools console right here and typing `global` into it.

PR #97, opened 2021-06-09
package.json: Remove duplicate wrong "main"

Artoria2e5:
Turns out #96 was partially caused by a package.json issue -- there actually *is* a CJS version already available, it's just it got masked by this wrong entry!

After this commit `esm` is no longer needed to `require` this package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,6 @@
 	"name": "colorjs.io",
 	"version": "0.0.3",
 	"description": "Color space agnostic color manipulation library",
-	"main": "color.js",
 	"exports": {
 		"import": "./dist/color.esm.js",
 		"require": "./dist/color.cjs.js",


LeaVerou:
Hey, thanks for spotting that!
Instead of removing it entirely, we should probably set it to something else, as [otherwise it defaults to `index.js`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#main) which doesn’t exist.

Artoria2e5:
There is a correct `"main": "./dist/color.cjs.js",` a couple lines down :)

LeaVerou:
Why would `main` be CJS though? Why not ESM?

Artoria2e5:
That's how package.json is designed to work. The ESM thing is right there too: `"module": "./dist/color.esm.js",`

kripod:
The `"main"` field is used by legacy Node versions, which only support CommonJS modules. This PR looks good to me.

kripod (approved):

LeaVerou:
Merged, thanks!

PR #99, opened 2021-06-14
Add HyAB distance

Artoria2e5:
This commit adds an HyAB-style distance and a very half-asses system for determining whether a colour space is applicable. Test vectors from the paper if any are yet to be added.

A little JSDoc with types are added where convenient. Some editor plugins (like my VSCode) make use of them, so that might make writing code with cjsio a bit more pleasant. I can’t quite figure out how to describe the statify thing though.

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -155,7 +155,12 @@ export default class Color {
 		return ret;
 	}
 
-	// Euclidean distance of colors in an arbitrary color space
+	/**
+	 * Euclidean distance of colors in an arbitrary color space
+	 *
+	 * @param {Color} color   - The other color.
+	 * @param {string} space  - Color space to run in.
+	 */
 	distance (color, space = "lab") {
 		color = Color.get(color);
 		space = Color.space(space);
@@ -172,12 +177,23 @@ export default class Color {
 		}, 0));
 	}
 
+	/**
+	 * Color difference (delta E) according to any of the supported methods.
+	 *
+	 * @param {Color} color       - The other color.
+	 * @param {string | Object} o - String specifying the method, or an object specifying both the method and additional options.
+	 */
 	deltaE (color, o = {}) {
 		if (util.isString(o)) {
 			o = {method: o};
 		}
 
-		let {method = Color.defaults.deltaE, ...rest} = o;
+		if (util.isString(Color.defaults.deltaE)) {
+			console.warn("The data structure for Color.defaults.deltaE has changed.");
+			Color.defaults.DeltaE = {method: Color.defaults.DeltaE};
+		}
+
+		let {method, ...rest} = {...Color.defaults.DeltaE, ...o};
 		color = Color.get(color);
 
 		if (this["deltaE" + method]) {
@@ -188,11 +204,15 @@ export default class Color {
 	}
 
 	// 1976 DeltaE. 2.3 is the JND
-	deltaE76 (color) {
-		return this.distance(color, "lab");
+	deltaE76 (color, options = {space: 'lab'}) {
+		return this.distance(color, options.space);
 	}
 
-	// Relative luminance
+	/**
+	 * Relative luminance
+	 * 
+	 * @returns {number}
+	*/
 	get luminance () {
 		// Luminance should actually be retrieved from XYZ with a D65 white point.
 		return Color.chromaticAdaptation(Color.spaces.xyz.white, Color.whites.D65, this.xyz)[1];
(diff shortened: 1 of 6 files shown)


Artoria2e5:
There are, in fact, no test vectors in the said paper. So I guess my job is done here.

src/color.js
```diff
@@ -172,12 +177,47 @@ export default class Color {
...
+	/**
+	 * HyAB distance of colors in an Lab-like color space.
+	 * Best used for large distances. Claims to outperform even DE2000 at that.
+	 * Said to be okay for small distances.
+	 * 
+	 * @see http://markfairchild.org/PDFs/PAP40.pdf
+	 * @param {Color} color   - The other color.
+	 * @param {string} space  - Color space to run in.
+	 */
+	distanceHyAB (color, space = "lab") {
```
LeaVerou, line 189: This should definitely be a separate module.

I wonder if we should add an `options` argument to `distance()` so this can be invoked that way.

src/color.js
```diff
@@ -971,7 +1019,7 @@ Object.assign(Color, {
 	defaults: {
 		gamutMapping: "lch.chroma",
 		precision: 5,
-		deltaE: "76", // Default deltaE method
+		deltaE: { method: "76" }, // Default deltaE method
```
LeaVerou, line 1022: I'm not necessarily opposed to this change, but I would like some justification for this change (and ideally this should have been a separate PR as it does not seem to be a required change for HyAB distance).

src/spaces/ictcp.js
```diff
@@ -26,6 +26,7 @@ Color.defineSpace({
 		CT: [-0.5, 0.5],	// Full BT.2020 gamut in range [-0.5, 0.5]
 		CP: [-0.5, 0.5]
 	},
+	classification: ['labish'],
```
LeaVerou, line 29: I like the idea of classifying color spaces, but it needs broader discussion about the API to do so.

What possible classifications do we envision? Are they mutually exclusive? Which of those can be inferred and which need to be explicitly specified? etc 

And what are the requirements for `labish`? E.g. `oklab` and `jzazbz` have completely different ranges than `lab`.

LeaVerou (commented):
Hello there, thank you so much for you contributions!
I’ll leave the actual color science to @svgeesus but I added some comments on the API changes. This needs more discussion I'm afraid. It would help merge it faster if you were willing to break it into smaller PRs, but I understand if that's too much of a hassle.

In general, smaller, atomic PRs can be iterated on and merged faster, whereas PRs that introduce a number of different changes need agreement for all changes before they're merged.

src/spaces/ictcp.js
Artoria2e5: My current requirements for labish is... basically the lightness-chroma_a-chroma_b layout. Some other classifications I am thinking of are:
* Element order: `rgb`, `lchish`, `hslish`
* Maybe some other properties: `perceptual`

It's made into a list for the possibility of having many tags.

src/color.js
Artoria2e5: The similarity in argument normalisation does make me think that way.

src/spaces/ictcp.js
LeaVerou: @svgeesus thoughts?

src/color.js
LeaVerou: We can keep the current signature as well, branch if the second argument is an object and treat it as an options object (`{space, method}`). Plugins can add methods, like they do for `deltaE`.
We'd probably want to throw if HyAB distance is used with a non-labish space.

src/spaces/ictcp.js
svgeesus: So, OKLab is certainly `labish` with L, a and b axes. But the ranges are different, [0,1] for OKLab L for example. And certainly, one would not want to apply something like deltaE2000 to OKLab since the various corrections in deltaE 2000  (blue hue non-uniformity, notably) do not apply to OKLab.

src/color.js
svgeesus: Yes, this should be a module under [`src/deltaE`](https://github.com/LeaVerou/color.js/tree/master/src/deltaE)

src/spaces/ictcp.js
Artoria2e5: Only some deltaE methods will be able to take a space parameter, so I guess that is all fine and dandy.

Artoria2e5:
Did a big rebase. The method commit is still retained because I kinda want to keep the HyAB interface "unsurprising" in terms of its ability to accommodate different spaces, although I do doubt its 76 name. Either way it should be easily removable now.

The main HyAB stuff is now moved to DeltaEHyAB.

Artoria2e5:
I am sure there's a comment I missed a couple of days ago. Where did it go....

svgeesus:
I think we would rather add ths (very simple) HyAB functionality without the associated refactoring and `Labish` concept which makes up the majority of this PR. Thanks for bringing HyAB to our attention, but closing this PR.

PR #100, opened 2021-06-14
A nitpick on chromaticAdaptation and a bunch of JSDoc

Artoria2e5:
chromaticAdaptation currently has an unused `options` function. The code alludes to `env.M` somehow getting set, so the nature thing I figured was to destructure `options` into the `env`.

The bunch of JSDoc is there because I am finding it really hard to fall asleep. They do make sense to VSCode.

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -516,7 +516,7 @@ export default class Color {
 			return XYZ;
 		}
 
-		let env = {W1, W2, XYZ, options};
+		let env = {W1, W2, XYZ, ...options};
 
 		Color.hooks.run("chromatic-adaptation-start", env);
 
--- a/src/color.js
+++ b/src/color.js
@@ -11,7 +11,9 @@ export default class Color {
 	// new Color(coords, alpha) // defaults to sRGB
 	// new Color(CSS variable [, root])
 	constructor (...args) {
-		let str, color;
+		let str;
+		/** @type {Color | ColorJSON} */ 
+		let color;
 
 		// new Color(color)
 		// new Color({spaceId, coords})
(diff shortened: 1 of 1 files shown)


src/color.js
```diff
@@ -516,7 +572,7 @@ export default class Color {
 			return XYZ;
 		}
 
-		let env = {W1, W2, XYZ, options};
+		let env = {W1, W2, XYZ, ...options};
```
LeaVerou, line 575: This was intentional. Now plugins cannot get the originally provided options doing `env.options` and cannot branch accordingly. Also, any existing references to `env.options` will break.

This is not a biggie for this particular case (I doubt there are that many plugins using `env.options`, though I think there are some official ones), but we do use `env.options` across the codebase, and it would break that convention.

LeaVerou (changes requested):
Hi there, thanks for sending this PR!

I'm afraid I do not agree with the change on `env.options`, I provided some reasoning in my comment.

I’m fine with adding jsdoc style comments, though I'm not sure about types like `Vec3` and `Mat33`, can we avoid abbreviations? e.g. `Vector_3x1` or `Matrix_3x3` would be far more readable, whereas I had to look up the definition to understand what `Vec3` was. Also, given that these are used across the codebase, I wonder if there's a more global way to define them?

src/color.js
Artoria2e5: Oh! I guess I will only keep the JSDoc changes then. Cheers.

LeaVerou: Thanks! Ping me when this is ready to merge!

Artoria2e5: I have reworked the injection of `env.M` to be explicit from `options`. I understand that adding a new param would be more consistent with the other place where `env` is used, but putting in another positional parameter itself is a compatibility break I do not wish to go for.

LeaVerou: Why do you need `M` on `env`?

LeaVerou:
@Artoria2e5 Are you still interested in working on this PR? We'd love to have more JSDoc in our codebase.

LeaVerou:
> @Artoria2e5 Are you still interested in working on this PR? We'd love to have more JSDoc in our codebase.

Closing due to lack of response (also a PR from that long ago would take a lot of effort to rebase)

PR #106, opened 2021-06-27
[RFC] Make eslint do something

### LINKED ISSUE #105: Eslint is not being run... but I think something else is wrong too!
I was trying to use eslint to fix my code style once and for all -- so I don't need to manually change them. But it first turns out that we don't really have a test set up to run it, not even in `package.json`. So I manually ran `node_modules/.bin/eslint src`, and sure enough the entire repo has been a bit sloppy too:

```
.....\color.js\src\deltaE\deltaEITP.js
  24:2  warning  Missing semicolon  semi

.....\GH\color.js\src\deltaE\deltaEJz.js
  32:2  warning  Closing curly brace appears on the same line as the subsequent
block  brace-style
  36:2  warning  Closing curly brace appears on the same line as the subsequent
block  brace-style
```

I figured re-running with `--fix` would solve the issue, but check out what I got instead:

```diff
diff --git a/src/deltaE/deltaEJz.js b/src/deltaE/deltaEJz.js
index 80f116e..5ad170b 100644
--- a/src/deltaE/deltaEJz.js
+++ b/src/deltaE/deltaEJz.js
@@ -29,11 +29,13 @@ Color.prototype.deltaEJz = function (sample) {
                // both undefined hues
                Hz1 = 0;
                Hz2 = 0;
-       } else
+       }
+ else
        if (Number.isNaN(Hz1)) {
                // one undefined, set to the defined hue
                Hz1 = Hz2;
-       } else
+       }
+ else
        if (Number.isNaN(Hz2)) {
                Hz2 = Hz1;
        }

```

Yep, it broke the indentation. I think.

Artoria2e5:
This commit enables eslint's `indent` check to produce more useful `--fix`
results, and also turns on the recommended set and fixes the resulting
(non-indent) complaints when it makes sense and disables them otherwise.

I think there should be a CI thing done too, but that's not something I can change. Anyway, the commit is mostly pointless in that it does not change anything in the program's behavior.

PS: the package-lock.json feels outdated, but I don't quite have the
energy to figure out what changed.

Close #105.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+	"extends": "eslint:recommended",
 	"parser": "babel-eslint",
 	"parserOptions": {
 		"ecmaVersion": 6,
@@ -9,7 +10,8 @@
 	},
 	"env": {
 		"browser": true,
-		"node": true
+		"node": true,
+		"es6": true
 	},
 	"rules": {
 		"semi": 1,
@@ -29,6 +31,8 @@
 		}],
 		"arrow-spacing": 1,
 		"comma-spacing": 1,
-		"keyword-spacing": 1
+		"keyword-spacing": 1,
+		"indent": ["warn", "tab"],
+		"no-unused-vars": ["error", { "argsIgnorePattern": "^_" }]
 	}
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,3 +32,6 @@ Here are a few other guideliines that cannot be enforced via ESLint:
 - Prefer single-word names over multi-word names. 3+ word names are especially frowned upon.
 - camelCase over underscore_case, with few exceptions.
 - Don't be afraid of unicode characters when appropriate, except on user-facing names. E.g. use ε over epsilon internally, but not ΔΕ over deltaE in the public-facing method name.
+
+If you prefer to run ESLint from the command-line, a `lint` script is provided so you can use `npm run lint`. To use the auto-fix functionality, try `npm run lint -- --fix`.
+Do note that the current code base is not very consistent on the indents, so you may see many unwanted results.
--- a/src/color.js
+++ b/src/color.js
@@ -79,7 +79,7 @@ export default class Color {
 
 	set space (value) {
 		// Setting spaceId works with color space objects too
-		return this.spaceId = value;
+		this.spaceId = value;
 	}
 
 	get spaceId () {
@@ -99,7 +99,7 @@ export default class Color {
 
 			// b) Remove instance properties from previous color space
 			for (let prop in this.space.instance) {
-				if (this.hasOwnProperty(prop)) {
+				if (Object.prototype.hasOwnProperty.call(prop)) {
 					delete this[prop];
 				}
 			}
@@ -255,7 +255,7 @@ export default class Color {
 	/**
 	 * @return {Boolean} Is the color in gamut?
 	 */
-	inGamut (space = this.space, options) {
+	inGamut (space = this.space, options = {}) {
 		space = Color.space(space);
 		return Color.inGamut(space, this[space.id], options);
 	}
@@ -473,7 +473,8 @@ export default class Color {
 
 			let fallbacks = Array.isArray(fallback)? fallback.slice() : Color.defaults.fallbackSpaces;
 
-			for (let i = 0, fallbackSpace; fallbackSpace = fallbacks[i]; i++) {
+			for (let i = 0; fallbacks[i]; i++) {
+				let fallbackSpace = fallbacks[i];
 				if (Color.spaces[fallbackSpace]) {
 					let color = this.to(fallbackSpace);
 					ret = color.toString({precision});
@@ -913,7 +914,7 @@ export default class Color {
 				return util.value(this, Color.shortcuts[prop]);
 			},
 			set (value) {
-				return util.value(this, Color.shortcuts[prop], value);
+				util.value(this, Color.shortcuts[prop], value);
 			},
 			configurable: true,
 			enumerable: true
@@ -924,7 +925,7 @@ export default class Color {
 	static statify(names = []) {
 		names = names || Object.getOwnPropertyNames(Color.prototype);
 
-		for (let prop of Object.getOwnPropertyNames(Color.prototype)) {
+		for (let prop of names) {
 			let descriptor = Object.getOwnPropertyDescriptor(Color.prototype, prop);
 
 			if (descriptor.get || descriptor.set) {
@@ -942,7 +943,7 @@ export default class Color {
 			}
 		}
 	}
-};
+}
 
 Object.assign(Color, {
 	util,
@@ -985,7 +986,7 @@ Color.defineSpace({
 		Z: []
 	},
 	white: Color.whites.D50,
-	inGamut: coords => true,
+	inGamut: _coords => true,
 	toXYZ: coords => coords,
 	fromXYZ: coords => coords
 });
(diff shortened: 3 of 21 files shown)


src/color.js
```diff
@@ -473,7 +473,8 @@ export default class Color {
 
 			let fallbacks = Array.isArray(fallback)? fallback.slice() : Color.defaults.fallbackSpaces;
 
-			for (let i = 0, fallbackSpace; fallbackSpace = fallbacks[i]; i++) {
+			for (let i = 0; fallbacks[i]; i++) {
+				let fallbackSpace = fallbacks[i];
```
LeaVerou, line 477: Why?

src/hooks.js
```diff
@@ -6,8 +6,8 @@ export default class Hooks {
 	add (name, callback, first) {
 		if (typeof arguments[0] != "string") {
 			// Multiple hooks
-			for (var name in arguments[0]) {
-				this.add(name, arguments[0][name], arguments[1]);
+			for (var one_name in arguments[0]) {
```
LeaVerou, line 9: This should just be a `let`

src/spaces/hsl.js
```diff
@@ -79,6 +79,7 @@ Color.defineSpace({
 	},
 
 	instance: {
+		// eslint-disable-next-line no-unused-vars
```
LeaVerou, line 82: In general I'm not a huge fan of these linting directives scattered throughout the codebase. It couples code and tooling too much, and if they are needed, it may indicate that the rule is too restrictive.
In this case, it looks like [`ignoreRestSiblings`](https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings) is what we need.

src/spaces/ictcp.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
LeaVerou, line 1: Where are the unused vars here?

src/util.js
```diff
@@ -83,6 +83,7 @@ export function parseCoord(coord) {
 	if (coord.indexOf(".") > 0) {
 		// Reduce a coordinate of a certain color space until the color is in gamut
 		let [spaceId, coordName] = coord.split(".");
+		// eslint-disable-next-line no-undef
 		let space = Color.space(spaceId);
```
LeaVerou, line 87: Where is the undeclared variable here?

.eslintrc.json
```diff
@@ -1,4 +1,5 @@
 {
+	"extends": "eslint:recommended",
```
LeaVerou, line 2: I would really prefer we explicitly list rules than import a black box set like this.

LeaVerou (changes requested):
Hi there, thank you for your time! See comments and questions.

src/color.js
```diff
@@ -99,7 +99,7 @@ export default class Color {
 
 			// b) Remove instance properties from previous color space
 			for (let prop in this.space.instance) {
-				if (this.hasOwnProperty(prop)) {
+				if (Object.prototype.hasOwnProperty.call(prop)) {
```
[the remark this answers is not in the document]
Artoria2e5: It's from eslint's [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins). The idea is that on untrusted objects this may be trouble. Since this is our own object this is more of a personal choice to not use a disable-next-line

src/util.js
Artoria2e5: `Color`, which is not quite a false positive...

src/spaces/ictcp.js
Artoria2e5: `rec2020`, but I think the intent of the TODO is pretty clear.

src/color.js
Artoria2e5: This is from [`no-cond-assign`](https://eslint.org/docs/rules/no-cond-assign), and personally I agree with its judgement here. I am not a big fan of doing assignments in conditions.

.eslintrc.json
```diff
@@ -29,6 +31,8 @@
 		}],
 		"arrow-spacing": 1,
 		"comma-spacing": 1,
-		"keyword-spacing": 1
+		"keyword-spacing": 1,
+		"indent": ["warn", "tab"],
```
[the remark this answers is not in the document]
Artoria2e5: It does not appear to generate any false positives from that, so I guess not. The warnings it gives are mostly 4 space vs tab things. The exception lies in hsl and hwb, where the toString is indeed indented wrong.

CONTRIBUTING.md
```diff
@@ -32,3 +32,6 @@ Here are a few other guideliines that cannot be enforced via ESLint:
 - Prefer single-word names over multi-word names. 3+ word names are especially frowned upon.
 - camelCase over underscore_case, with few exceptions.
 - Don't be afraid of unicode characters when appropriate, except on user-facing names. E.g. use ε over epsilon internally, but not ΔΕ over deltaE in the public-facing method name.
+
+If you prefer to run ESLint from the command-line, a `lint` script is provided so you can use `npm run lint`. To use the auto-fix functionality, try `npm run lint -- --fix`.
+Do note that the current code base is not very consistent on the indents, so you may see many unwanted results.
```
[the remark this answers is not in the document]
Artoria2e5: I am a little bit concerned with how it might interact with `git blame`, so...

src/color.js
```diff
@@ -255,7 +255,7 @@ export default class Color {
 	/**
 	 * @return {Boolean} Is the color in gamut?
 	 */
-	inGamut (space = this.space, options) {
+	inGamut (space = this.space, options = {}) {
```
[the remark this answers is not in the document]
Artoria2e5: That's from [`default-param-last`](https://eslint.org/docs/rules/default-param-last). The idea is sound, but it *is* unnecessary.

src/color.js
```diff
@@ -79,7 +79,7 @@ export default class Color {
 
 	set space (value) {
 		// Setting spaceId works with color space objects too
-		return this.spaceId = value;
+		this.spaceId = value;
```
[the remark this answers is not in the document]
Artoria2e5: That's [`no-setter-return`](https://eslint.org/docs/rules/no-setter-return). The return value does not do anything, so ESLint thinks it's best to not write it.

src/util.js
LeaVerou: Whoa, it looks like this file does not import `Color` anywhere! Does it depend on having `Color` as a global variable? If so this will break in most bundles! We should be fixing this instead of adding comments to get ESLint to ignore it…

src/spaces/ictcp.js
LeaVerou: I don’t see a TODO?
I think we should just remove or comment out that line, @svgeesus what do you think?

src/color.js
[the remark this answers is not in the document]
LeaVerou: See this is why I prefer explicitly listing rules rather than extending `eslint:recommended`. 😉

src/color.js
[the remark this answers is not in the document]
LeaVerou: We should add this rule to the ESLint config then!
(see other comments about extending `eslint:recommended`)

src/color.js
LeaVerou: I have no strong opinion about this particular case, but I wouldn't want to blanket forbid assigning in conditionals.

src/util.js
Artoria2e5: It... does! The tricky thing is that `util` is imported by `color`, so I guess it needs to go into `color.js`??

LeaVerou: Yikes. I can't remember what happens if two ES modules import each other, does it create a loop or does the engine do something clever?

LeaVerou:
Hi there, what is the status of this? Are you still interested in working on it or should I close this PR? Thanks!

PR #114, opened 2021-11-26
Gamut mapping: use the appropriate deltaE method, JND and epsilon based on mapSpace

svgeesus:
Also sets the default method to oklch.chroma in line with CSS Color 4

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -311,30 +311,44 @@ export default class Color {
 
 		if (method.indexOf(".") > 0 && !this.inGamut(space)) {
 			let clipped = color.toGamut({method: "clip", space});
-			if (this.deltaE(clipped, {method: "2000"}) > 2) {
-				// Reduce a coordinate of a certain color space until the color is in gamut
-				let [mapSpace, coordName] = util.parseCoord(method);
+
+			// Reduce a coordinate of a certain color space until the color is in gamut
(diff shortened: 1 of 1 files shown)


src/color.js
```diff
@@ -311,30 +311,44 @@ export default class Color {
...
-				// Reduce a coordinate of a certain color space until the color is in gamut
-				let [mapSpace, coordName] = util.parseCoord(method);
+
+			// Reduce a coordinate of a certain color space until the color is in gamut
+			let [mapSpace, coordName] = util.parseCoord(method);
+
+			let delta = "2000";
+			let ε = .01; // for deltaE
+			let JND = 2;
+			if (mapSpace.id === "oklch") {
```
LeaVerou, line 321:
```suggestion

			if (mapSpace.id === "oklch") {
```

nit: blank line before control flow blocks

src/color.js
```diff
@@ -963,7 +977,7 @@ Object.assign(Color, {
 
 	// Global defaults one may want to configure
 	defaults: {
-		gamutMapping: "lch.chroma",
+		gamutMapping: "oklch.chroma",
```
LeaVerou, line 980: Not so sure about using this as a default, what happens if oklch is not available?

LeaVerou: (not saying we shouldn't have it as a default if oklch is, but that it shouldn't be the *only* default)

Issue #123, opened 2022-01-26
Selecting text on https://colorjs.io/apps/convert/ behaves like a click


smfr:
Trying to select color functions on https://colorjs.io/apps/convert/ is rather frustrating; whenever the mouse up happens, the page treats it like a click, and changes the page content, which destroys the selection. So you can't just select and Copy a color function.

PR #124, opened 2022-01-28
[apps/convert] fix issue #123

Mohammed785:


LeaVerou:
Hey, thank you for working on this!

However, UX wise I don't think this is a good solution.
1. It is quite surprising to overwrite the user's clipboard just because they clicked on a color format. Writing on the user's clipboard should always be opt-in, through UI that clearly indicates what it does.
2. It still doesn't solve the actual UI problem which is that colors cannot be selected with the mouse without actually clicking on them.

cc @smfr @svgeesus for input

LeaVerou (changes requested):
See comments

Mohammed785:
please review the new updates

svgeesus:
#123 has been closed, meanwhile, so closing this PR

PR #126, opened 2022-02-17
Reduce npm package size

ai:
Current `colorjs.io` package is 1.7 MB which is a lot.

By removing website. development configs, etc we can reduce it to 1.2 MB.

If this PR will be accepted, I can find a way to reduce it even further. But it is more complicated, since Rollup creates huge source map files.


LeaVerou:
Oh nice, thanks!

PR #129, opened 2022-02-18
Remove `.min` files from npm package

ai:
- [x] `npm` ignores `node_modules/` by default
- [x] We do not use `.min.js` files in npm package. Removing it will reduce size `1.2 MB` → `0.7 MB`.


LeaVerou:
Thanks!

PR #130, opened 2022-02-18
Fix .min cleaning and remove IIFE builds

ai:
What do you think of removing IIFE build?

I think use case for it is very low:

1. Most of the users use bundlers
2. Quick hacks without bundlers is better to do with ESM build since it is an official way (IIFE was a hack)
3. Removing IIFE will drop other 300 KB on users’s machine and CI. Not to big, but seems like unnecessary resource spending since IIFE use case is extremely low.

In my open source, I do not have IIFE build and had no problems with users asking about it.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,0 @@
-dist/*.min.js
-dist/*.min.js.map
--- a/package.json
+++ b/package.json
@@ -3,12 +3,14 @@
 	"version": "0.0.4",
 	"description": "Color space agnostic color manipulation library",
 	"files": [
-		"dist/"
+		"dist/color.cjs.js",
+		"dist/color.cjs.js.map",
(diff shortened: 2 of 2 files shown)


LeaVerou:
IIFE is more friendly to novices who haven't learned ESM yet so we don't want to remove it. 
That is also why you haven't heard people asking for it, because those people won't even visit the project's repo.

However, perhaps we could remove it from npm, since I imagine most people who use IIFE would use it either directly from the website or after downloading the file themselves?

LeaVerou:
Merged, since this was just removing IIFE from the npm module

ai:
This is the final npm package size reduction PR (which can be made without API changes). We can release changes to npm.

LeaVerou:
Done!

PR #131, opened 2022-02-19
Move acorn to devDependencies

ai:
@LeaVerou  I found another way to reduce npm package size.

`dist/` files don’t use `acorn`. It was used only in notebook which is not part of npm package.

So we can move it from `dependencies` to `devDependencies`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,4733 +1,10578 @@
 {
-  "name": "color.js",
-  "version": "0.0.1",
-  "lockfileVersion": 1,
-  "requires": true,
-  "dependencies": {
-    "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
-      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
-      "dev": true,
-      "requires": {
-        "@babel/highlight": "^7.8.3"
-      }
-    },
-    "@babel/generator": {
-      "version": "7.10.2",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.10.2.tgz",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,4733 +1,10578 @@
 {
```
ai: npm updated lock file according to `.editorconfig`. Should we keep it since new version of npm will do it for every contributor to the project?

LeaVerou:
Oh I thought devDependencies were for deps related to the build process, whereas dependencies were for deps actually used by the project. No?

Not sure I understand your followup comment. Keep what?

ai:
> Oh I thought devDependencies were for deps related to the build process, whereas dependencies were for deps actually used by the project.

The problem is that we are using the same repo for website and for npm package. So we technically have two sets of dependencies: `dependencies`/`devDependencies` for npm package and `dependencies`/`devDependencies` for website.

There is no simple solution for this problem. I suggest to keep `dependencies` for npm package dependencies and `devDependencies` for website/build tools.

> Not sure I understand your followup comment. Keep what?

When I called `npm install` it updated `package-lock.json` from old version 1 to new version 2. I have npm 8, which have new lockfile format.

LeaVerou (approved):
I guess moving acorn to devDependencies is fine, can't think of how it may affect anything negatively, so I'm gonna go ahead and approve this.
Then you can merge it whenever you're happy with `package-lock.json`? Not sure whether you want to keep it or remove it from this PR. Up to you!

ai:
> Then you can merge it whenever you're happy with package-lock.json? Not sure whether you want to keep it or remove it from this PR. Up to you!

Let’s keep it. It will make contribution simpler for new developer (who will have npm >7 too).

LeaVerou:
Sure. Are you able to merge it yourself with the approved review or should I merge it?

ai:
Nope, I do not have Merge button

LeaVerou:
I can add you as a contributor if you plan to work more on the project, what do you think?

ai:
> I can add you as a collaborator if you plan to work more on the project, what do you think?

I am not sure about my contribution in month later.

Let’s keep PR flow for now.

Issue #55, opened 2020-10-05
How to run and test locally?


facelessuser:
This issue is created in response to https://github.com/LeaVerou/color.js/issues/54#issuecomment-703844158.

I'm not sure it is a bug as much as maybe a question. How can one quickly run and evaluate this project locally? Coding up a web environment is surely possible, and one could import the code and test, but when it mentioned you could install via npm, I found that an attractive option as I was looking more to debug issue #50 

1. Tried to install via the documentation's recommendation for node, and the package isn't found in the registry, so maybe that is outdated.

2. Tried to install by using the git repo, seemed to work, but couldn't figure out how to import it. Seems it is an ESM module as it complained about `import`? Struggled to figure out when I couldn't load it. Googled some things and tried to load it as an ESM module, but only ran into other complaints.

Maybe all that is needed is a little documentation for the less knowledgable like me 🙂.

I'm on node.js v14.4.0.

LeaVerou:
Ah, I see what's going on here: Because the library is pre-release, we never published on npm, since you need a release for that. 🤦🏽‍♀️
We could make a release with a 0.0.x version or something, just so we can publish on npm. @svgeesus are you ok with that? 

We'll need to find a good package name too. All the following are taken 😕

http://npmjs.com/package/color
https://www.npmjs.com/package/color-js
https://www.npmjs.com/package/color.js
https://www.npmjs.com/package/colorjs

Maybe https://www.npmjs.com/package/colorjs.io or Maybe https://www.npmjs.com/package/colorjsio (which is what's in the docs)

Regarding import: yes, it's an ESM module, so you need to import it with a regular `import` statement, not `require()` (which also means you need Node v13.2 or later)

facelessuser:
> Ah, I see what's going on here: Because the library is pre-release, we never published on npm, since you need a release for that

I completely understand if it isn't ready. The docs mentioned it was still a prerelease, I assumed, incorrectly, that the prereleases were available as the docs mentioned you could use node.

> Regarding import: yes, it's an ESM module, so you need to import it with a regular import statement, not require() (which also means you need Node v13.2 or later)

This was also the tricky part. So I figured out this much and tried to import it with `import`, but it complained about that. Googling led me to someone saying you should add `"type": "module"` to the package JSON, which then gave me new issues about trying to use an ESM module. It is unclear to me how you can get the package running as a node module. With all of that said, maybe it is an issue with the latest node, I'm not sure as I haven't tried out other node releases.

[...]

PR #132, opened 2022-02-19
Fix ESM support in Node.js

ai:
Current package doesn’t work in Node.js with `import` (I need `color.js` to generate image in Node.js script).

```
(node:79791) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
/home/ai/Dev/oklch-picker/node_modules/.pnpm/colorjs.io@0.0.5/node_modules/colorjs.io/dist/color.esm.js:2970
export default Color$1;
```

It happens because Node.js need explicit mark that the file is ESM file (with `package.type = "module"` or with `.mjs` extension).

I decided to add `.esm.mjs` files to npm package (and keep old `.esm.js` on website for URL compatibility).

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -188,4 +188,19 @@ gulp.task("bundle", async function () {
 	}
 });
 
-gulp.task("default", gulp.parallel("css", "bundle", "html", "md"));
+gulp.task("mjs", function () {
+	return gulp
+		.src(["./dist/color.esm.js*"])
+		.pipe(
+			rename((path) => {
+				if (path.extname === ".js") {
+					path.extname === ".js.mjs";
+				} else {
+					path.basename = "color.esm.mjs";
+				}
+			})
+		)
+		.pipe(gulp.dest("./dist/"));
+});
+
+gulp.task("default", gulp.parallel("css", "bundle", "mjs", "html", "md"));
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -194,7 +194,7 @@ gulp.task("mjs", function () {
 		.pipe(
 			rename((path) => {
 				if (path.extname === ".js") {
-					path.extname === ".js.mjs";
+					path.extname = ".mjs";
 				} else {
 					path.basename = "color.esm.mjs";
 				}
--- a/package.json
+++ b/package.json
@@ -5,15 +5,15 @@
 	"files": [
 		"dist/color.cjs.js",
 		"dist/color.cjs.js.map",
-		"dist/color.esm.js",
-		"dist/color.esm.js.map"
+		"dist/color.esm.mjs",
+		"dist/color.esm.mjs.map"
 	],
 	"exports": {
-		"import": "./dist/color.esm.js",
+		"import": "./dist/color.esm.mjs",
 		"require": "./dist/color.cjs.js"
 	},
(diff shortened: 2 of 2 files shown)


LeaVerou:
Not a huge fan of `.mjs`, why not just use `"type": "module"` in package.json?

ai:
> why not just use "type": "module" in package.json?

It will force renaming all `.js` files to `.cjs`. For instance, `gulpfile.js` need to be renamed to `gulpfile.cjs` (or convert it to ESM, but it could take a while).

LeaVerou:
But at least those are internal. 
Won't renaming the ESM files to `.mjs` break existing code, including code that links to them from `colorjs.io`?

ai:
> Won't renaming the ESM files to .mjs break existing code, including code that links to them from colorjs.io?

`dist/` will have both: `color.esm.js` and `color.esm.mjs`. I think it should not break anything, but I am not 100% sure.

gulpfile.js
```diff
@@ -188,4 +188,19 @@ gulp.task("bundle", async function () {
...
+					path.extname = ".mjs";
+				} else {
+					path.basename = "color.esm.mjs";
+				}
+			})
+		)
+		.pipe(gulp.dest("./dist/"));
+});
+
+gulp.task("default", gulp.parallel("css", "bundle", "mjs", "html", "md"));
```
romainmenke, line 206: I might be wrong, but having `bundle` and `mjs` in parallel might result in an older version being renamed before rollup is done creating the new one.

romainmenke:
What seems to work for users of `postcss-preset-env` :

https://github.com/csstools/postcss-plugins/blob/main/plugins/postcss-base-plugin/package.json#L11-L23

```json
  "main": "dist/index.cjs",
  "module": "dist/index.mjs",
  "exports": {
    ".": {
      "import": "./dist/index.mjs",
      "require": "./dist/index.cjs",
      "default": "./dist/index.mjs"
    }
  },
```

We know they do these things :

- used in commonjs scripts
- used in esmodule scripts
- bundled with webpack and used in browsers
- bundled with rollup

So the change by @ai should work.
Only the gulp order might need a slight change.

LeaVerou:
I don't understand what's the downside with just using `"type": "module"`. Why not just rename gulpfile? It's one file.

ai:
> I don't understand what's the downside with just using "type": "module".

It will force all `.js` files to be ESM files (`module.exports`, `require`, `__dirname` will stop working).

1. Move all Node.js files (like `gulpfile`) to ESM. It is not a big problem, but I do not know the project to do it on my own.
2. We will need to rename `dist/color.cjs.js` to `dist/color.cjs.cjs` (`.cjs` extension will switch file from ESM to CJS). It could theoretically break something, but I am not sure.
3. Test that everything is still working.

I agree that it is the best solution for the project. But it requires much more work.

LeaVerou:
We don't really use any CJS files besides a few for build tools and the color.js CJS export, so the impact should be minimal (I think `color.cjs` should be fine, no need for the repetition).

LeaVerou:
This was fixed recently, see more details in #55.

The gist is:

- We added `"type": "module"
- All CJS config files we were using now have a `.cjs` extension

PR #136, opened 2022-02-27
Fix issue where longer logic does not match CSS spec

facelessuser:
It appears that Color.js "longer" hue adjust logic does not match the CSS specification: https://www.w3.org/TR/css-color-4/#hue-longer.

The important difference is that while the spec states in pseudo code:

```js
if (0 < θ₂ - θ₁ < 180) {
  θ₁ += 360;
}
else if (-180 < θ₂ - θ₁ < 0) {
  θ₂ += 360;
}
```

It appears Color.js is using something similar to:

```js
if (0 < θ₂ - θ₁ < 180) {
  θ₂ += 360;
}
else if (-180 < θ₂ - θ₁ <= 0) {
  θ₁ += 360;
}
```

### DIFF
--- a/src/angles.js
+++ b/src/angles.js
@@ -27,10 +27,10 @@ export function adjust (arc, angles) {
 	else if (arc === "longer") {
 		if (-180 < angleDiff && angleDiff < 180) {
 			if (angleDiff > 0) {
-				a2 += 360;
-			}
-			else {
 				a1 += 360;
+			}
+			else if (angleDiff < 0) {
+				a2 += 360;
 			}
 		}
 	}


LeaVerou:
Thanks for spotting that! Which behavior is better? I wrote both that part of the spec and that part of color.js, so either could be changed :P

facelessuser:
> Which behavior is better?

I guess that depends on the intention 🙂.

The handling of equivalent angles is what threw me off and made me take a closer look. Any time we have a scenario where two colors with exact hues (or equivalents, think 0 and 360 or 180 and 540), the unpatched implementation gives you an interpolation of all hues while the patched version keeps the interpolated hue constant:

Unpatched:
Screen Shot 2022-02-27 at 3 04 46 PM

Patched:
Screen Shot 2022-02-27 at 3 04 57 PM

In a way, I can see how both could be interpreted as valid. You either interpret the long way around the two identical hues, or you interpret the long way around the two identical hues. I honestly don't know which is right 🙃.

LeaVerou:
The latter seems more correct to me. @svgeesus?

Is this the only difference?

facelessuser:
I kind of agree that I think the latter is the intended.

facelessuser:
I'll see if I can get a CSS Color 4 issue raised later today.

facelessuser:
Since the CSS spec is not properly describing the intent, and the intent is being followed here, I will close this PR.

svgeesus:
CSS [now properly describes the intent](https://github.com/w3c/csswg-drafts/commit/17d0417ceea0eb97ea971a689844d7513f9ec426), so color.js continues to be correct

PR #162, opened 2022-06-27
Tree-shakeable, faster, module-based, procedural API

### LINKED ISSUE #159: Tree-shaking ready API
Right now library exports one big class with everything inside.

It is bad for website performance because webapp JS bundle will contain all methods of Color.js even if user use only a few of them.

Modern JS bundler has tree-shaking feature. If you will move rare (or all) methods to separated functions with separated export, JS bundler will keep only used functions in the JS bundle.

I am talking about something like:

```js
import { contrast } from 'colorjs.io'

contrast(color)
```

I understand that this feature requires a full refactoring and will not be implemented soon. Just an idea for better web performance.

Color spaces could be tree-shaked as well. Here is [an example from `culori`](https://culorijs.org/guides/tree-shaking/).

LeaVerou:
Will close #159 once merged

The idea is:

- Functions defined in modules. 
- Functions are standalone and can work on either plain objects (with the right properties) or  actual `Color` objects.
- Functions return plain objects
- We only use these plain functions internally
- `color.js` becomes optional
- Plain functions added on `Color` with a simple method call and then they return `Color` objects instead of plain objects (staying compatible with the current API)
- `index.js` imports everything, including `Color`
- `index-fn.js` imports and re-exports all functions

This means there is both a tree-shakeable, procedural API that is faster and easier to debug, as well as a slower OO API that can be more readable in certain cases.

The new API resulted in a 4x speed increase in [this benchmark inspired by @ai's use case](https://colorjs.io/benchmarks/ai.html).


LeaVerou:
I still need to add docs but I think I fixed all the regressions and this should be ready for review. 
@ai feel free to weigh in too (it won't let me add you as a reviewer).

benchmarks/ai.html
```diff
@@ -67,13 +68,15 @@
...
+import inGamut from "../src/inGamut.js";
+import serialize from "../src/serialize.js";
 
 test("Color.js", colorjs_canvas, {
-	create: (l, c, h) => new Color('oklch', [l, c, h]),
-	to_srgb: color => color.to("srgb"),
-	in: color_srgb => color_srgb.inGamut(undefined, {epsilon: 0}),
-	str: color_srgb => color_srgb.toString()
+	create: (l, c, h) => ({space: oklch, coords: [l, c, h]}),
+	to_srgb: color => to(color, "srgb"),
```
ai, line 77: Does `srgb` color space register in `to()`?

We can do `to(color, srgb)` to avoid registration.

src/index-fn.js
```diff
@@ -0,0 +1,14 @@
+export {default as getColor} from "./getColor.js";
```
ai, line 1: If we replace `export default` to `export` in `./*.js` this file will be smaller.

LeaVerou: Yes but default exports are nicer for the individual modules.

src/variations.js
```diff
@@ -0,0 +1,14 @@
+import ColorSpace from "./space.js";
+import set from "./set.js";
+
+export function lighten (color, amount = .25) {
+	let space = ColorSpace.get("oklch", "lch");
```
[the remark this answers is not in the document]
LeaVerou: `ColorSpace.get("oklch", "lch")` loads oklch if it's available, or lch if it's not, i.e. loading oklch progressively enhances how `lighten()` works.

src/interpolation.js
```diff
@@ -143,25 +148,29 @@ export function range (color1, color2, options = {}) {
 
 	outputSpace = outputSpace? ColorSpace.get(outputSpace) : space;
 
-	color1 = color1.to(space).toGamut();
-	color2 = color2.to(space).toGamut();
+	color1 = to(color1, space);
+	color2 = to(color2, space);
+
+	// Gamut map to avoid areas of flat color
```
svgeesus, line 154: Should document that GM happens before interpolation here, especially as this is a change from CSS Color 4

svgeesus (approved):
Just one comment on need for documentation; no changes requested

LeaVerou:
Changes are in! Time to do a new release on npm and see how it goes.

manuelmeister:
Is it correct that the named contrast functions e.g. `contrastAPCA` are not included in `colorjs.io/fn`?

LeaVerou:
> Is it correct that the named contrast functions e.g. `contrastAPCA` are not included in `colorjs.io/fn`?

They should be. If not, it's a bug. But I believe they are exported en masse from `contrast/index-fn.js`, so if you're searching the source for `contrastAPCA` it won't find anything.

manuelmeister:
Okay I'll pull the last version again and look at the typings again.

MysteryBlokHed:
There's a file `src/contrast/index.js` that exports all the contrast functions, but it doesn't seem to be re-exported in `src/index-fn.js`:

<https://github.com/LeaVerou/color.js/blob/0b3cacae4af0b93fd0d21ff88b21efd3be27109d/src/index-fn.js#L1-L30>

PR #170, opened 2022-06-29
Rework subpath and conditional export declaration

sgomes:
Hey folks! 👋

It looks like `package.json` is mixing conditional and subpath export declarations at the top level. This is not [the recommended approach as per the `node` documentation](https://nodejs.org/api/packages.html#conditional-exports).

Specifically, it recommends the following approach:

```js
{
  "exports": {
    ".": "./index.js",
    "./feature.js": {
      "node": "./feature-node.js",
      "default": "./feature.js"
    }
  }
}
```

The current approach is not well supported with common tools like `webpack`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,10 @@
 		"src/"
 	],
 	"exports": {
-		"import": "./dist/color.js",
-		"require": "./dist/color.cjs",
+		".": {
+			"import": "./dist/color.js",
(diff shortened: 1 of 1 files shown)


sgomes:
Oh, looks like someone created something similar while I was working on this 🙂 Closing in favour of #169.

Issue #171, opened 2022-06-29
Name collision in `/fn`


sgomes:
Looks like `chromaticity.js` and `interpolation.js` both have an export named `register`. This becomes a problem because they both get re-exported with `*` from `index-fn.js`:

```js
export *                      from "./luminance.js";
export *                      from "./chromaticity.js";
```

Sample webpack error:

```
ERROR in ./node_modules/colorjs.io/src/index-fn.js 20:0-55
The requested module './chromaticity.js' contains conflicting star exports for the name 'register' with the previous requested module './luminance.js'
 @ ./src/index.js 1:0-52

ERROR in ./node_modules/colorjs.io/src/index-fn.js 24:0-56
The requested module './interpolation.js' contains conflicting star exports for the name 'register' with the previous requested module './luminance.js'
 @ ./src/index.js 1:0-52
```

LeaVerou:
Fixed with #172

PR #172, opened 2022-06-29
Fix index-fn reexport collisions

sgomes:
This PR renames some re-exports in `index-fn.js`, so that there are no name collisions.

See #171.

### DIFF
--- a/src/index-fn.js
+++ b/src/index-fn.js
@@ -16,9 +16,15 @@ export {default as distance}  from "./distance.js";
 export {default as equals}    from "./equals.js";
 export {default as contrast}  from "./contrast.js";
 export {default as clone}     from "./clone.js";
-export *                      from "./luminance.js";
-export *                      from "./chromaticity.js";
+export {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thank you so much! I don't think we need these exports at all, the whole point of `index-fn` is to be tree-shakeable and `register()` is used to add these methods to the `Color` object (see how it's used in `index.js`). So I think we can just remove these exports, do you agree?

LeaVerou:
Oops, didn't mean to close this, sorry! (puzzled how Cmd + enter did that)

sgomes:
> So I think we can just remove these exports, do you agree?

@LeaVerou Yes, that sounds good! Done 🙂

LeaVerou (approved):

LeaVerou:
Thanks, just did so! I guess I'll wait an hour or so before other low hanging fruit emerges

LeaVerou:
Actually scratch that, if people are playing with it, we shouldn't have a broken package. I've merged all open PRs and will release now.

PR #181, opened 2022-06-30
Improve handling of angular units for hues

### LINKED ISSUE #180: Only `deg` is supported as an angular unit for hues
Currently, `src/util.js` has a [TODO note](https://github.com/LeaVerou/color.js/blob/main/src/util.js#L73) to handle more (angular) units. These are `rad, `grad` and `turn` in CSS. 

I believe it would also be useful to accept the degree sign `°` U+00B0 (i.e. outside US-ASCII) as a synonym for `deg`, which is currently possible due to lax parsing that ignores punctuation characters after the number. 

I've also argued before that `pi` or `pirad` should be a valid angular unit in CSS. This would be trivial to add. However, since hues are almost always provided in degrees, this is probably not the best place to make that case.

By the way, the regular expressions used in `parseFunction()` do also not accept all numbers allowed by CSS Values and Units or CSS Syntax, e.g. scientific notation, and they are not case-insensitive.

### LINKED ISSUE #311: Support grad, rad, and turn angle units for the hue value of hsl()
I'd like to be able to pass any of the 4 hue angle units (`deg`, `grad`, `rad`, and `turn`) to the `new Color(str)` constructor, but at the moment only `deg` works.

```js
new Color('hsl(90deg 40% 50%)') // works
new Color('hsl(100grad 40% 50%)') // throws
new Color('hsl(1.57rad 40% 50%)')  // throws
new Color('hsl(0.25turn 40% 50%)') // throws
```

Crissov:
fixes #180 
fixes #311 (dupe)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -74,35 +74,23 @@ export function parseFunction (str) {
 				arg.type = "<angle>";
 				arg.unit = "deg";
 			}
-			else if (/°$/.test(arg)) { // this notation is not valid in CSS
-				// Drop degree sign from degrees and convert to number
-				arg = new Number(+arg.slice(0, -1));
-				arg.type = "<angle>";
-				arg.unit = "deg";
-			}
 			else if (/grad$/.test(arg)) { // check for 'grad' needs to come before check for 'rad' 
 				// Drop grad from gradians (gon) and convert to number in degrees
 				arg = new Number(+arg.slice(0, -4) / 10 * 9);
 				arg.type = "<angle>";
-				arg.unit = "deg";
+				arg.unit = "grad";
 			}
 			else if (/rad$/.test(arg)) {
 				// Drop rad from radians and convert to number in degrees
 				arg = new Number(+arg.slice(0, -3) * 180 / Math.PI);
 				arg.type = "<angle>";
-				arg.unit = "deg";
-			}
-			else if (/pi$/.test(arg)) {// this unit is not valid in CSS
-				// Drop pi from π radians and convert to number in degrees
-				arg = new Number(+arg.slice(0, -2) * 180);
-				arg.type = "<angle>";
-				arg.unit = "deg";
+				arg.unit = "rad";
 			}
 			else if (/turn$/.test(arg)) {
 				// Drop turn from τ radians and convert to number in degrees
 				arg = new Number(+arg.slice(0, -4) * 360);
 				arg.type = "<angle>";
-				arg.unit = "deg";
+				arg.unit = "turn";
 			}
 			else if (isNumberRegex.test(arg)) {
 				// Convert numerical args to numbers
--- a/src/util.js
+++ b/src/util.js
@@ -70,11 +70,12 @@ export function parseFunction (str) {
 			}
 			else if (/(?:deg|g?rad|turn)$/.test(arg)) {
 				// Drop the unit symbol and convert to number in degrees
+                                arg = new Number(+arg.slice(0, -unit.length));
                                 switch (unit) {
-				  case "deg": arg = new Number(+arg.slice(0, -3)); break;
-				  case "grad": arg = new Number(+arg.slice(0, -4) / 10 * 9); break;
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -55,26 +55,55 @@ export function parseFunction (str) {
...
 				arg.type = "<percentage>";
 			}
 			else if (/deg$/.test(arg)) {
 				// Drop deg from degrees and convert to number
-				// TODO handle other units too
 				arg = new Number(+arg.slice(0, -3));
 				arg.type = "<angle>";
 				arg.unit = "deg";
 			}
+			else if (/°$/.test(arg)) { // this notation is not valid in CSS
```
LeaVerou, line 77: We are aiming for eventual parity with CSS, so we will not accept `°` unless CSS does, sorry.

src/util.js
```diff
@@ -55,26 +55,55 @@ export function parseFunction (str) {
...
+				arg.type = "<angle>";
+				arg.unit = "deg";
+			}
+			else if (/rad$/.test(arg)) {
+				// Drop rad from radians and convert to number in degrees
+				arg = new Number(+arg.slice(0, -3) * 180 / Math.PI);
+				arg.type = "<angle>";
+				arg.unit = "deg";
+			}
+			else if (/pi$/.test(arg)) {// this unit is not valid in CSS
```
LeaVerou, line 95: Same as `°`, CSS only units please.

src/util.js
```diff
@@ -55,26 +55,55 @@ export function parseFunction (str) {
...
 				arg.type = "<angle>";
 				arg.unit = "deg";
 			}
+			else if (/°$/.test(arg)) { // this notation is not valid in CSS
+				// Drop degree sign from degrees and convert to number
+				arg = new Number(+arg.slice(0, -1));
+				arg.type = "<angle>";
+				arg.unit = "deg";
+			}
+			else if (/grad$/.test(arg)) { // check for 'grad' needs to come before check for 'rad'
```
LeaVerou, line 83: This repeated regex testing is inefficient. You can have a regex that just reads the unit into a group and then does simple string comparison. This will also help with the repetitive code.

src/util.js
```diff
@@ -55,26 +55,55 @@ export function parseFunction (str) {
...
+				// Drop degree sign from degrees and convert to number
+				arg = new Number(+arg.slice(0, -1));
+				arg.type = "<angle>";
+				arg.unit = "deg";
+			}
+			else if (/grad$/.test(arg)) { // check for 'grad' needs to come before check for 'rad' 
+				// Drop grad from gradians (gon) and convert to number in degrees
+				arg = new Number(+arg.slice(0, -4) / 10 * 9);
+				arg.type = "<angle>";
+				arg.unit = "deg";
```
LeaVerou, line 87: `arg.unit` also becomes pointless if you only ever set it to `deg`, I'd do the conversion like you're doing, but preserve the info about which unit was used in `arg.unit`.

LeaVerou (changes requested):
Hi @Crissov, thank you for working on this! I left some comments.

src/util.js
Crissov: I agree, just didn’t want to change too much.

LeaVerou:
Hi @crissov, is this PR ready for re-review? Please ping me when it is. Thanks for also adding tests!

Crissov:
@LeaVerou I’ve written and committed the code already, but did so on my phone and haven’t had the chance yet to actually try it.

LeaVerou:
Ok, please ping me when it's ready!

Crissov:
@LeaVerou Okay, if I’m not overlooking anything this has long been ready for re-review. Also fixes #311.

Crissov:
superseded by #324

(I still like named capture groups, though)

LeaVerou:
> (I still like named capture groups, though)

Me too — happy to merge a PR that adapts the current code to use named capture groups :)

PR #182, opened 2022-06-30
Don't import Color class in keywords.js

sgomes:
The `Color` class is being imported in `keywords.js`, without being used anywhere in the file.

This is causing the full compositional API to be included as long as keywords are used, even if the user is making use of just `/fn` functionality.

### DIFF
--- a/src/keywords.js
+++ b/src/keywords.js
@@ -2,7 +2,6 @@
  * Note that this does not include currentColor, transparent,
  * or system colors
  */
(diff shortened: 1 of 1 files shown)


LeaVerou:
Oh shit, thanks man!!

LeaVerou:
This will likely be in v0.2.5 that I will release by tomorrow (though there's a chance it may be 0.3.0 if I finish some changes soon enough).

Issue #183, opened 2022-06-30
Individual color spaces not exported


sgomes:
There is currently no mechanism for a consumer to pull in a single color space, as they're not exposed via `/src/index-fn.js`.

We can't expose `spaces/index.js`, since that depends on every color space and expects them all to load. As such, the only feasible option for now is to map them individually in `package.json`, e.g.:

```json
"exports": {
  ".": {
    "import": "./dist/color.js",
    "require": "./dist/color.cjs"
  },
  "./fn": "./src/index-fn.js",
  "./spaces/srgb.js": "./src/spaces/srgb.js",
  "./spaces/p3.js": "./src/spaces/p3.js",
  [...]
}
```

[Node also supports wildcard exports](https://nodejs.org/api/packages.html#subpath-patterns) (which could be a good fit here), but popular bundlers like `webpack` apparently do not, which limits their use.

The alternative would be to work towards dropping side-effects in color space files, so that we could have an index file that would re-export them individually, together with a separate file that would load and register all of them.

I'm happy to contribute the work here, but I'm not sure which of the two options (if either) you think is best!

LeaVerou:
Yeah, right now color spaces cannot be tree-shaken, one would need to import `spaces/index.js`, which as you point out expects them all to load. So this is basically the same issue as #163, just a different facet of it.

I think probably the best way forwards is:
- Color spaces drop side effects
- `spaces/index-fn.js` imports and re-exports all of them
- `index-fn.js` imports and re-exports `spaces/index-fn`
- `spaces/index.js` imports everything from `spaces/index-fn.js` and registers it calling `ColorSpace.register()`

Wondering if there's a better suffix than `fn` for this sort of thing.

sgomes:
> So this is basically the same issue as #163, just a different facet of it.

You're right, this is fundamentally the same issue!

> I think probably the best way forwards is:
> 
> * Color spaces drop side effects
> * `spaces/index-fn.js` imports and re-exports all of them
> * `index-fn.js` imports and re-exports `spaces/index-fn`
> * `spaces/index.js` imports everything from `spaces/index-fn.js` and registers it calling `ColorSpace.register()`

I'm happy to prototype that! I'll post a PR once I have something.

> Wondering if there's a better suffix than `fn` for this sort of thing.

I'm coming up short of alternatives, but I think `fn` is okay, in the same vein as `date-fns` vs more monolithic libraries like moment.

LeaVerou:
[...]

Issue #163, opened 2022-06-28
Color spaces are not tree-shakeable


LeaVerou:
(Remaining part of #159)

While the new procedural API is tree-shakeable wrt functions, since color spaces register themselves onto `ColorSpace.registry`, this means they are not tree-shakeable.

To make them tree-shakeable, we'd need to eliminate side effects from the color space modules. Then, the importing code would register them.

Something like:

```js
import ColorSpace from "COLORJS_ROOT/src/space.js";
import oklch from "COLORJS_ROOT/src/spaces/oklch.js";
ColorSpace.register(oklch);
```

We could make this less painful by offering pre-generated indices that do this for all existing color spaces, but I'm still struggling with the idea of requiring more plumbing to use Color.js, which makes it easier to make mistakes. E.g. someone could easily forget to register their color space and do something like:

```js
import oklch from "COLORJS_ROOT/src/spaces/oklch.js";
let color = new Color("oklch", [.5, .2, 340]); // TypeError: No color space found with id = "oklch"
```

Note that registration is needed only to use string ids instead of color space objects. The API is perfectly usable with just passing identifiers around and never registering any color space, like so:

```js
import srgb from "COLORJS_ROOT/src/spaces/srgb.js";
import oklch from "COLORJS_ROOT/src/spaces/oklch.js";
let color = new Color(oklch, [.5, .2, 340]);
color.to(srgb);
```

Thoughts?

sgomes:
Thank you for spending so much time working on a procedural API 🙇 That massively helps with keeping bundle sizes down!

> Note that registration is needed only to use string ids instead of color space objects.

Yes, that's what I see as the greatest difficulty in using the current color space approach for everything. Would it be feasible to make parsing and serialization self-sufficient within each colour space?

The way the top-level `parse` and `serialize` functions are generated is already really modular, so it may be a case of making a couple of utils that are just a bit more generic than that, and using them in every colour space to generate standalone versions there?

sgomes:
> Perhaps it could accept an optional list of color spaces that will be used instead of ColorSpace.all / ColorSpace.registry?

Yeah, an optional list of color spaces would likely do the trick here 👍 That would allow for use of the function while having an empty registry.

`serialize` doesn't need it, as you mention, but perhaps it could do the same, for symmetry? Though I'm really not sure. I can't decide whether it would indeed make things easier to use or just unnecessarily introduce more complexity 🙂

lydell:
[...]

PR #184, opened 2022-06-30
Draft: remove side-effects from spaces

sgomes:
Following the discussion in #163 and #183, this (starts to) implement an approach laid out by @LeaVerou to remove the usage of side-effects in spaces, and make them available externally in a tree-shakeable way:

> I think probably the best way forwards is:
> 
> * Color spaces drop side effects
> * `spaces/index-fn.js` imports and re-exports all of them
> * `index-fn.js` imports and re-exports `spaces/index-fn`
> * `spaces/index.js` imports everything from `spaces/index-fn.js` and registers it calling `ColorSpace.register()`

Please let me know what you think!

Aside: my editor seems to be ending blank lines at the end of files. Please let me know whether I should change my config to stop that from happening!


src/spaces/index.js
```diff
@@ -1,22 +1,11 @@
...
-import "./rec2020.js";
-import "./oklab.js";
-import "./oklch.js";
+import ColorSpace from "../space.js";
 
-import "./index-hdr.js";
\ No newline at end of file
+import * as spaces from './index-fn.js';
+
+ColorSpace.register(spaces.XYZ_D65);
```
sgomes, line 5: Question: do we need to use `RGBColorSpace.register` instead of `ColorSpace.register` when registering RGB color spaces, or can we use the latter for everything?

LeaVerou: We can use `ColorSpace.register()` for everything. But I don't recommend listing them out like that, just loop over `spaces`

LeaVerou:
> Aside: my editor seems to be adding newlines at the end of files. Please let me know whether I should change my config to stop that from happening!

I'm fine with that!

package.json
```diff
@@ -14,7 +14,8 @@
 			"import": "./dist/color.js",
 			"require": "./dist/color.cjs"
 		},
-		"./fn": "./src/index-fn.js"
+		"./fn": "./src/index-fn.js",
+		"./fn/spaces": "./src/spaces/index-fn.js"
```
LeaVerou, line 18: I'd expect `index-fn` to also import `spaces/index-fn`, so this shouldn't be needed. But not a strong opinion.

LeaVerou:
Looking good so far!

src/spaces/index.js
```diff
@@ -1,22 +1,11 @@
...
+// ...
+
+// Register xyz alias
+ColorSpace.register("xyz", XYZ_D65);
+XYZ_D65.formats.color.ids ||= [];
+XYZ_D65.formats.color.ids.push("xyz");
```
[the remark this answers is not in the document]
sgomes: That's a very good point, yes! I suppose that would mean adding support for aliases to the options object?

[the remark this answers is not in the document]
LeaVerou: I suppose. Can't think of a better way.

src/space.js
```diff
@@ -271,6 +272,15 @@ export default class ColorSpace {
 
 		space = this.get(space);
 		this.registry[id] = space;
+
+		if (this.aliases) {
+			for (let alias of this.aliases) {
+				this.registry[alias] = space;
```
LeaVerou, line 278: A few things here: 
a) `this` is the `ColorSpace` instance here, so `this.registry` will be `undefined`
b) The constructor should be side effect free. It should be `register()` that adds stuff to `ColorSpace.registry`
c) The id for the `color()` function should be decoupled from aliases. In this case it happens that the alias is also an id for `color()`, but not sure this should be enforced by the API

sgomes: Oh, right, that's a good point, thank you! Sounds like space aliases need to be separate from color ID aliases 👍

LeaVerou: Also, what should be the expected behavior if one registers two spaces with the same id or alias? Right now each one will overwrite the previous one.

src/spaces/index.js
LeaVerou: Ha, you can count on me for finding aaaalll the ways to reduce duplication 😆

src/space.js
sgomes: Throwing a runtime error seems like the best bet to me. Having a unit test that checks to make sure there aren't any such errors on the full lib should then be enough to make sure the build doesn't break.

How does that sound?

package.json
sgomes: Done in 05bdd32631c2100ea7d1362584f340cb4e7a4b12.

I also added a `sideEffects` declaration to `package.json` to ensure that tree-shaking works correctly.

src/space.js
LeaVerou: Not sure about a runtime error, users should be able to call `register()` multiple times from different parts of the code to register the same color space. Though perhaps we could only throw the error if the color space object is different.

package.json
```diff
@@ -56,5 +56,6 @@
 		"postcss-nesting": "latest",
 		"rollup": "^2.10",
 		"rollup-plugin-terser": "^7"
-	}
+	},
+	"sideEffects": ["./src/spaces/index.js"]
```
LeaVerou, line 60: Should we also add `src/index.js` here?

src/space.js
```diff
@@ -46,6 +47,10 @@ export default class ColorSpace {
 			this.formats.color.id = this.id;
 		}
 
+		if (options.idAliases) {
+			this.formats.color.ids = options.idAliases;
```
LeaVerou, line 51: I don't think we need this. They can just be defined in the format itself.

src/space.js
```diff
@@ -263,20 +268,46 @@ export default class ColorSpace {
...
-	static create (options) {
-		let space = new this(options);
-		return this.register(space);
+		if (space.aliases) {
+			for (let alias of space.aliases) {
+				this.addToRegistry(alias, space);
+			}
+		}
+
+		this.validateIds();
```
LeaVerou, line 308: I like detailed errors as much as the next person, but checking that deeply this seems a little too much.

src/space.js
```diff
@@ -263,20 +268,46 @@ export default class ColorSpace {
...
 		if (arguments.length === 1) {
 			space = arguments[0];
 			id = space.id;
 		}
 
 		space = this.get(space);
-		this.registry[id] = space;
-		return space;
-	}
+		this.addToRegistry(id, space);
```
LeaVerou, line 300: Do we need a separate method? When somebody is using the `register(id, space)` signature, we should not be registering any additional things like aliases. So all we need to do is preserve what signature was used. Then we can recursively call it to register a color space with each alias, and still only have the checking code in one place.

src/space.js
sgomes: Yes, I totally agree that it shouldn't error out if registering the same space multiple times. The code should be doing that already, but do let me know if I missed something!

package.json
sgomes: Yes, it does seem to have side effects 👍

src/space.js
sgomes: Good point! Removed.

src/space.js
sgomes: Do you mean we should forgo the ID check altogether? I think that's okay. As far as I can tell, if there is a collision, it'll simply use the first match, which is probably an acceptable behaviour?

src/space.js
sgomes: Sorry, I'm not entirely sure I understand 😕

Do you mean that we should't have the aliases as part of the color space definition (`space.aliases`), and we should instead delegate the responsibility of registering the alias to the consumer? Or do you mean that we should simply call `register` recursively from within itself, but keep the responsibility of registering aliases within the `ColorSpace` class?

src/spaces/xyz-d65.js
```diff
@@ -10,8 +10,9 @@ export default new ColorSpace({
 	},
 	white: "D65",
 	formats: {
-		color: {}
+		color: {
+			ids: ['xyz'],
```
LeaVerou, line 14: We should probably add both, for clarity

src/space.js
LeaVerou: I mean, there is a semantic difference between the two signatures here. If someone calls `ColorSpace.register("foobar", mySpace)` they intend to register `mySpace` under the `"foobar"` alias, and not do any other registration. E.g. someone may import a color space and want to register it with a different id altogether. If they do `ColorSpace.register(mySpace)`, then it makes sense to register it with its id and aliases.

This also means that *we* can use it that way internally: When `ColorSpace.register()` is used with a single argument, we can internally call `ColorSpace.register(id, space)` as well as `ColorSpace.register(alias, space)` for every alias, thus not needing a separate function for this.

Does it make sense now?

sgomes: Oh, I see, thank you for clarifying that! Yes, that sounds like a good approach 👍

sgomes: Done!

src/spaces/xyz-d65.js
sgomes: Done!

package.json
```diff
@@ -56,5 +56,6 @@
 		"postcss-nesting": "latest",
 		"rollup": "^2.10",
 		"rollup-plugin-terser": "^7"
-	}
+	},
+	"sideEffects": ["src/index.js", "./src/spaces/index.js"]
```
LeaVerou, line 60: Nitpick: Can we format it in multiple lines?

src/spaces/index.js
```diff
@@ -1,22 +1,9 @@
...
-import "./oklab.js";
-import "./oklch.js";
+import ColorSpace from "../space.js";
 
-import "./index-hdr.js";
\ No newline at end of file
+import * as spaces from './index-fn.js';
+
+for (let key of Object.keys(spaces)) {
+  ColorSpace.register(spaces[key]);
```
LeaVerou, line 6: Nitpick: indentation. We use smart tabs, not spaces

sgomes: Oops! I prefer tabs as well, but it looks like my IDE defaults to spaces when there's no prettier config. Fixed!

package.json
sgomes: Good idea. Done!

src/spaces/index.js
```diff
@@ -1,22 +1,9 @@
...
 
-import "./index-hdr.js";
\ No newline at end of file
+import * as spaces from './index-fn.js';
+
+for (let key of Object.keys(spaces)) {
+	ColorSpace.register(spaces[key]);
+}
+
+// ...
```
LeaVerou, line 9: Is there a way to re-export en masse?

sgomes:
Do you think this is ready to move forward to full implementation across all color spaces, @LeaVerou?

src/spaces/index.js
LeaVerou, line 9: Don't forget to re-export!

LeaVerou:
Yes! 😄

src/spaces/index.js
sgomes: Oh, do we want to re-export from this file as well? No problem, can do 🙂

sgomes:
I've applied this to all color spaces, and rebased on `main`. I think this is ready for review!

LeaVerou:
Thank you! The testsuite is in a state of disarray right now, but have you checked that all passing tests still pass? (I just pushed some fixes so you'd need to rebase first)

sgomes:
> Thank you! The testsuite is in a state of disarray right now, but have you checked that all passing tests still pass? (I just pushed some fixes so you'd need to rebase first)

I did try to run the tests, but even after rebasing from `main` again I'm still getting 404s on resources from `raw.githubusercontent.com`. Perhaps I'm missing something? I'm not entirely sure how to actually go about running the tests; all I tried was to open `tests/index.html` locally.

LeaVerou:
> > Thank you! The testsuite is in a state of disarray right now, but have you checked that all passing tests still pass? (I just pushed some fixes so you'd need to rebase first)
> 
> I did try to run the tests, but even after rebasing from `main` again I'm still getting 404s on resources from `raw.githubusercontent.com`. Perhaps I'm missing something? I'm not entirely sure how to actually go about running the tests; all I tried was to open `tests/index.html` locally.

Oh shit, you're right, the test harness is broken. I'm on it, but it will take a while. Meanwhile you can try loading the different HTML pages individually. Sorry this is such a PITA. 😢

sgomes:
> Oh shit, you're right, the test harness is broken for users who are not logged in to GitHub. I'm on it, but it will take a while. Meanwhile you can try loading the different HTML pages individually (or log in). Sorry this is such a PITA. 😢

No worries at all 🙂

Looks like there are some issues to figure out on my end, namely with ordering; registering `xyz-abs-d65` before `xyz-d65` results in an error, because the former depends on the latter. We may have to go back to explicitly listing them (instead of looping over them) to maintain ordering, or instead being smarter about it to reorder at registration time.

LeaVerou:
Wait, does `xyz-abs-d65` refer to `xyz-d65` by id? It should import the color space and use that reference.

sgomes:
> Wait, does `xyz-abs-d65` refer to `xyz-d65` by id? It should import the color space and use that reference.

The issue isn't something that `xyz-abs-d65` specifically is doing, but rather the `ColorSpace` constructor:

```js
this.base = options.base ? ColorSpace.get(options.base) : null;
```

Since `base` is provided through an ID, and since `ColorSpace.get` relies on the color space registry, this fails during construction of the `xyz-abs-d65` space.

One way of solving this would be by having `base` provide not the ID but the ColorSpace instance itself. What do you think, @LeaVerou?

LeaVerou:
> > Wait, does `xyz-abs-d65` refer to `xyz-d65` by id? It should import the color space and use that reference.
> 
> The issue isn't something that `xyz-abs-d65` specifically is doing, but rather the `ColorSpace` constructor:
> 
> ```js
> this.base = options.base ? ColorSpace.get(options.base) : null;
> ```
> 
> Since `base` is provided through an ID, and since `ColorSpace.get` relies on the color space registry, this fails during construction of the `xyz-abs-d65` space.
> 
> One way of solving this would be by having `base` provide not the ID but the ColorSpace instance itself. What do you think, @LeaVerou?

Yeah, base should always point to the ColorSpace instance, not an id. I think I've done that with most spaces, but I guess not that one.
That line should still be there because other people should be able to provide base through an id, but we should never do that internally.

sgomes:
> Yeah, base should always point to the ColorSpace instance, not an id.

Oh, I hadn't noticed there was code to handle that possibility, thanks!

I've updated all the existing IDs to instances, and that seems to have fixed things 👍

sgomes:
I rebased and tested `main` vs this branch, and I get the exact same number of passes/failures for all tests 👍

Some of the tests still fail with 404s, this time from `projects.verou.me`, but everything seems to be the same before and after these changes.

LeaVerou:
I just fixed the tests to eliminate the 404, but the failing requests were for cosmetic things, so I don't think this affects anything here. I think I'm gonna go ahead and merge this! Thanks again!

sgomes:
Woohoo, thank you @LeaVerou! 🎉

After these changes, the procedural API is looking quite good. To give you an example, here is a map for what gets included on a simple test case that just makes use of `parse`, `serialize`, and the `sRGB` color space:

image

21 KB uncompressed / 8.4KB compressed is quite good 🙂

LeaVerou:
This should be in the upcoming v0.3.0 release, hopefully later today, but failing that tomorrow.

LeaVerou:
Btw what did you use to generate the graphic above? It's really cool!

PR #197, opened 2022-07-17
[parse] Trim input, fixes #196

### LINKED ISSUE #196: `parse()` should trim input strings
This leads to a TypeError:

```js
let c = " #bada55"; // Notice the leading space
let color = new Color(c);
```
There's the same problem with CSS named colors, leading or trailing spaces produce errors. 
So it might be a good idea to trim the inputs before parsing, to make the library idiot proof ? 

Otherwise fantastic library !

AlainGourves:
Fixes #196 
Parse() should trim input strings to avoid edge cases where leading and trailing spaces cause type error.

```js
let color1 = new Color("red"); // works
let color2 = new Color(" red"); // this doesn't
```

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -4,7 +4,7 @@ import ColorSpace from "./space.js";
 
 // CSS color to Color object
 export default function parse (str) {
-	let env = {str};
+	let env = {"str": str?.trim()};
 	hooks.run("parse-start", env);
 
 	if (env.color) {
--- a/src/parse.js
+++ b/src/parse.js
@@ -4,7 +4,7 @@ import ColorSpace from "./space.js";
 
 // CSS color to Color object
 export default function parse (str) {
-	let env = {"str": str?.trim()};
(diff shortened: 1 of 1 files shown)


src/parse.js
```diff
@@ -4,7 +4,7 @@ import ColorSpace from "./space.js";
 
 // CSS color to Color object
 export default function parse (str) {
-	let env = {str};
+	let env = {"str": str?.trim()};
```
LeaVerou, line 7: Actually, we may want to coerce to a string here as well (`str + ""`) before trimming.

LeaVerou (changes requested):
Thanks! See comment!

src/parse.js
JoaoAMarinho: Something like this?
```diff
-let env = {"str": str?.trim()};
+let env = {"str": (str+"")?.trim()};
```
The parentheses could also be removed (just preference).

PR #209, opened 2022-08-20
[contrast / deltaPhi]

Myndex:
New contrast module, deltaPhiStar 𝜟𝜱✴︎

https://github.com/Myndex/deltaphistar

This version is approximately aligned with APCA light mode.

deltaPhiStar 𝜟𝜱✴︎ is symmetric and does not consider polarity. It is intended as a general purpose perceptual contrast method, which is similar to APCA, and useable for high spatial frequency stimuli on self-illuminated monitors.

May be particularly useful for dataviz elements.

### DIFF
--- a/src/contrast/deltaPhi.js
+++ b/src/contrast/deltaPhi.js
@@ -6,7 +6,7 @@
 
 import getColor from "../getColor.js";
 import {getLuminance} from "../luminance.js";
-import lab-d65 from "../spaces/lab-d65.js";
+import lab_d65 from "../spaces/lab-d65.js";
 
 const phi = Math.pow(5, 0.5) * 0.5 + 0.5; // Math.phi can be used if Math.js
 
@@ -14,8 +14,8 @@ export default function contrastDeltaPhi (color1, color2) {
 	color1 = getColor(color1);
 	color2 = getColor(color2);
 
-	let Lstr1 = get(color1, [lab-d65, "l"]);
-	let Lstr2 = get(color2, [lab-d65, "l"]);
+	let Lstr1 = get(color1, [lab_d65, "l"]);
+	let Lstr2 = get(color2, [lab_d65, "l"]);
 
 	let deltaPhiStar = Math.abs(Math.pow(Lstr1,phi) - Math.pow(Lstr2,phi));
 
--- a/src/contrast/deltaPhi.js
+++ b/src/contrast/deltaPhi.js
@@ -1,0 +1,34 @@
+// Delta Phi Star perceptual lightness contrast
+// The (difference between two Lstars each raised to phi) raised to (1/phi)
+// Symmetric, does not matter which is foreground and which is background
+
+
+import getColor from "../getColor.js";
+import {getLuminance} from "../luminance.js";
+
+// from CIE standard, which now defines these as rational fractions
+const epsilon = 216/24389;  // 6^3/29^3
+const kappa = 24389/27;    // 29^3/3^3
+
+const phi = Math.pow(5, 0.5) * 0.5 + 0.5; // Math.phi can be used if Math.js
+
+// CIE Y to Lstar
+function YtoLstar(Y) {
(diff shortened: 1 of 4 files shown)


LeaVerou (approved):
LGTM

LeaVerou:
Thank you!

src/contrast/deltaPhi.js
```diff
@@ -0,0 +1,34 @@
+// Delta Phi Star perceptual lightness contrast
+// The (difference between two Lstars each raised to phi) raised to (1/phi)
+// Symmetric, does not matter which is foreground and which is background
+
+
+import getColor from "../getColor.js";
+import {getLuminance} from "../luminance.js";
+
+// from CIE standard, which now defines these as rational fractions
```
svgeesus, line 9: There is no need to re-implement CIE L*, it is already available in color.js.As an example usage, see

https://github.com/LeaVerou/color.js/blob/main/src/contrast/Lstar.js

src/contrast/deltaPhi.js
```diff
@@ -0,0 +1,34 @@
...
+	// Assuming Y is relative to D65 ref = 1, convert to CIE Lstar
+	return Y > epsilon ? 116 * Math.cbrt(Y) - 16 : kappa * Y;
+	// L* in range [0,100]
+}
+
+export default function contrastDeltaPhi (color1, color2) {
+	color1 = getColor(color1);
+	color2 = getColor(color2);
+
+	let Lstr1 = YtoLstar(Math.max(getLuminance(color1), 0));
```
svgeesus, line 26: Suggest instead the simpler:

let Lstr1 = get(color1, [lab, "l"]);

I see that negative luminance is clamped to zero; is there a particular risk of negative luminance?

src/contrast/deltaPhi.js
svgeesus: But it will need to be imported:

import lab from "../spaces/lab.js";

src/contrast/deltaPhi.js
```diff
@@ -0,0 +1,34 @@
+// Delta Phi Star perceptual lightness contrast
+// The (difference between two Lstars each raised to phi) raised to (1/phi)
+// Symmetric, does not matter which is foreground and which is background
```
svgeesus, line 3: Add a link to the source of this:

// See https://github.com/Myndex/deltaphistar

svgeesus (changes requested):

src/contrast/deltaPhi.js
Myndex: Hi @svgeesus -- I didn't use the lab.js as it appeared to me that it does a transform to  D50, and DeltaPhiStar requires D65 for sRGB. I thought of adding a commit to adjust lab.js to have a D65 option, but I didn't want to open that can of worms...

src/contrast/deltaPhi.js
Myndex: >  _...is there a particular risk of negative luminance?_

No, I was literally just following the pattern of the other contrast modules here. I started this module based on Weber.js, which had the same clamp. I am not familiar enough with the total color.js library to know what is an important pattern...

src/contrast/deltaPhi.js
Myndex: I added lab-d65.js in commit [fe321fa], and updated deltaPhi.js to match

src/contrast/deltaPhi.js
Myndex: fe321fa updates deltaPhi.js to ~match Lstar.js, except using lab-d65.js

src/contrast/deltaPhi.js
Myndex: Added with commit fe321fa

src/contrast/deltaPhi.js
svgeesus: Aha yes, good point. Thanks for adding the D65 variant.

svgeesus (approved):

PR #211, opened 2022-08-26
Add Carbon

sayzlim:
This is a new PR with the same purpose as this PR: https://github.com/LeaVerou/color.js/pull/195

### DIFF
--- a/_includes/plain.njk
+++ b/_includes/plain.njk
@@ -20,6 +20,11 @@
 
 {{ content | safe }}
 
+
+<aside class="carbon">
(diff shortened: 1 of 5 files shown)


sayzlim:
@LeaVerou this one should be okay now. Please review and let me know if I need to change anything else.

LeaVerou:
Hi @sayzlim, thank you! I think I would prefer it closer to the right side of the viewport on large screens, rather than right next to the content. Would you be able to make this change?

sayzlim:
@LeaVerou all set! Please review again.

LeaVerou:
Fantastic, thank you!

PR #212, opened 2022-08-26
Fix procedural examples

ajlende:
Color spaces need to be registered for `parse`, `convert`, `toGamut`, and `serialize` to work in the procedural example. I was a bit surprised that I had to register LCH for `toGamut` to work, but if it uses LCH by default, I suppose it can't be avoided.

### DIFF
--- a/docs/procedural.md
+++ b/docs/procedural.md
@@ -25,21 +25,30 @@ but everything applies to importing from URLs too.
 import {
 	to as convert,
 	toGamut,
-	serialize
+	serialize,
+	ColorSpace,
+	sRGB,
+	P3,
+	LCH
 } from "colorjs.io/fn";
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hey, thanks for pointing this out! That sounds like a bug we need to fix in the code, not the documentation.

LeaVerou:
I mean, color spaces do need to be registered for `parse()` to work. For `convert()`, you can pass in a reference instead of a string. `toGamut` should import its own dependencies, I suspect the reason it doesn't right now is that you can choose to reduce via a different coordinate. `serialize()` should not need any color spaces to be registered?

LeaVerou:
So I'm inclined to merge this, and we can always tweak in the future :) Thanks!

Issue #206, opened 2022-08-16
Add CI to run tests for old Node.js versions


ai:
We need to support at least the current Node.js versions: 14, 16, 18.

It will be easier to add `color.js` if we will support Node.js 12 for a some time.

I suggest using Github Action since it is free and has a good integration to Github.

LeaVerou:
I fully support this.

colorjs is actually hosted on Netlify, which is even more powerful than Github Actions :)

ai:
Am I right that Netlify is a hosting, not CI? Can we run tasks there and see the tasks status in PR?

LeaVerou:
Yes, absolutely! That's one of the main things Netlify does!

PR #215, opened 2022-09-09
Add Babel with support for Node v12.

### LINKED ISSUE #205: Avoid Nullish coalescing operator to support old Node.js
We are adding `color.js` to Stylelint for cool [`color-no-out-gamut-range`](https://github.com/stylelint/stylelint/pull/6271) rule.

But `??` operator from `color.js` sources breaks Node.js 12 and 14 support:

```
    /home/runner/work/stylelint/stylelint/node_modules/colorjs.io/dist/color.cjs:328
    		let coords = options.coords ?? this.base.coords;
    		                             ^

    SyntaxError: Unexpected token '?'
```

jgerigmeyer:
Fixes #205.

This is related to #206 as well, but does not add in CI to test Node version support.


rollup.config.js
```diff
@@ -1,44 +1,54 @@
+import { babel } from "@rollup/plugin-babel";
+import commonjs from "@rollup/plugin-commonjs";
```
LeaVerou, line 2: I'm not sure why we need this one?

LeaVerou:
Hey, thanks! See one question, apart from that LGTM.
In future PRs please do try to separate style changes from the actual PR, as it makes reviewing harder.

rollup.config.js
jgerigmeyer: Because this is currently setup to use [@babel/preset-env](https://babeljs.io/docs/en/babel-preset-env) with [`useBuiltIns: 'usage'`](https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage), it automatically pulls in any necessary polyfills from [core-js](https://github.com/zloirock/core-js). Since we want to include those polyfills in the distributed bundles, we need to use this plugin to convert them from CommonJS modules to ES6, so Rollup can include them. See: https://github.com/rollup/plugins/tree/master/packages/node-resolve#using-with-rollupplugin-commonjs

jgerigmeyer (commented):
> In future PRs please do try to separate style changes from the actual PR, as it makes reviewing harder.

Reverted style changes and squashed commits. 👍

rollup.config.js
LeaVerou: I see, thanks. Does it only pull in polyfills that are actually needed or everything? Do you have any sense of how much this increases the size of the bundles?

jgerigmeyer: > Does it only pull in polyfills that are actually needed or everything?

It only pulls in the polyfills that are necessary for the target (node v12) based on the code actually used. In this case, I think it's these two small polyfills:

- https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.error.cause.js (used throughout)
- https://github.com/zloirock/core-js/blob/master/packages/core-js/modules/es.reflect.to-string-tag.js (only used by `space-accessors.js`)

> Do you have any sense of how much this increases the size of the bundles?

It increases the bundle size by about 44% (non-minified) or 30% (minified):

Before:

```
Created bundle color.global.js: 107.94 kB → 34.67 kB (gzip)
Created bundle color.js: 104.35 kB → 34.23 kB (gzip)
Created bundle color.cjs: 104.42 kB → 34.27 kB (gzip)
Created bundle color.min.js: 41.59 kB → 16.48 kB (gzip)
Created bundle color.global.min.js: 41.6 kB → 16.52 kB (gzip)
Created bundle color.min.cjs: 41.66 kB → 16.5 kB (gzip)
```

After:

```
Created bundle color.global.js: 155.44 kB → 44.57 kB (gzip)
Created bundle color.js: 151.04 kB → 44.22 kB (gzip)
Created bundle color.cjs: 151.12 kB → 44.25 kB (gzip)
Created bundle color.min.js: 54.88 kB → 21.45 kB (gzip)
Created bundle color.global.min.js: 54.89 kB → 21.55 kB (gzip)
Created bundle color.min.cjs: 54.95 kB → 21.48 kB (gzip)
```

If you're concerned about that, I suppose another option would be to add new bundles that have support for older versions of Node, and not run everything through Babel?

LeaVerou: Hey, so sorry for the delay, I didn't get a notification and only saw you had responded when I manually visited the PR to see if there were any updates. Ouch, that's quite an increase. Yes, I think the way to go would be to have separate bundles. In that case we could even support somewhat older browsers too in those bundles, what do you think?
Is there a good naming convention for the transpiled bundles that we could follow?

jgerigmeyer: > Yes, I think the way to go would be to have separate bundles. In that case we could even support somewhat older browsers too in those bundles, what do you think? Is there a good naming convention for the transpiled bundles that we could follow?

@LeaVerou I agree, that seems like a good way to go. Are you thinking multiple new bundles with different levels of browser/node support? Or one "legacy" bundle that has support for older platforms?

I think I'd lean toward the latter, and probably I'd propose something like `color-legacy.js`? 🤷 

I'll do a quick update to try that out with Node 12 -- let me know if there are other platforms you think should be targeted (see: https://babeljs.io/docs/en/options#targets).

jgerigmeyer: (The [most recent commit](https://github.com/LeaVerou/color.js/pull/215/commits/fb3c8102e9bc2d25be453f364dede9f45536a26f) doesn't work -- I'll push a fixed version shortly.)

jgerigmeyer: @LeaVerou Adjusted in https://github.com/LeaVerou/color.js/pull/215/commits/862425c7466dbf1034f2015ef01e8446d80cc00f so that the original bundles are unchanged.

jgerigmeyer:
Hi @LeaVerou, any thoughts on this? Thanks!

jgerigmeyer:
Hi @LeaVerou, sorry to bother you, but I'm happy to wrap up this PR if you have any feedback? Thanks again!

LeaVerou:
Sorry @jgerigmeyer, I had a deadline, then travel, and then got covid 🤦🏽‍♀️ I'm still in quarantine, but will try to review this soon. So sorry about the delay, please don't take this as meaning anything about how interested we are in merging this PR!

jgerigmeyer:
Oh no! Please take your time, and feel better soon!

LeaVerou:
Thank you!
I took a look and LGTM, my only comment would be since we're producing separate versions anyway, and doing all this work with Babel etc, we may as well add some more targets (older browsers that people may want to support) and produce a `color.global.legacy.js` build as well (or would it be `color.legacy.global.js`?)

jgerigmeyer:
> Thank you! I took a look and LGTM, my only comment would be since we're producing separate versions anyway, and doing all this work with Babel etc, we may as well add some more targets (older browsers that people may want to support)

@LeaVerou Sure! I added the browserslist `defaults` in https://github.com/LeaVerou/color.js/pull/215/commits/d008c585f650bc313b0b79a61e98b7e7950aecdf -- this is the current list that it's supporting: https://browsersl.ist/#q=defaults%2C+node%3E%3D+12

Do you have particular browsers in mind? I wasn't sure where to make the cutoff, and if we go much further back (e.g. Android 4 or iOS 12) then the bundles roughly double in size.

> and produce a `color.global.legacy.js` build as well (or would it be `color.legacy.global.js`?) (Not crazy old like IE11 though)

Done -- here's the full list:

```
dist/color.global.js
dist/color.global.min.js
dist/color.js
dist/color.min.js
dist/color.cjs
dist/color.min.cjs
dist/color.global.legacy.js
dist/color.global.legacy.min.js
dist/color.legacy.js
dist/color.legacy.min.js
dist/color.legacy.cjs
dist/color.legacy.min.cjs
```

LeaVerou:
LGTM, thank you for working on this! Just merged. :)

PR #219, opened 2022-09-17
support the premultiplied option in mix

redneb:
`range` supports the `premultiplied` option which is not supported by `mix`, even though `mix` is implemented on top of `range`. I think it makes sense to support it in `mix` as well.


svgeesus (approved):
LGTM

PR #231, opened 2022-10-28
Add alpha blending

### LINKED ISSUE #230: Add alpha blending
We should have a method for [alpha blending](https://en.wikipedia.org/wiki/Alpha_compositing). 

Things to decide:
- **Naming**: Perhaps `overlay()` or `over()`? Or `overlayOn()` or just `on()` which is more clear as a `Color` method, but not so much in the procedural API (See `overlayOn(color1, color2)`).
- Is there any reason for it to take more than 2 arguments? My understanding is that the operator is associative.

The [math](https://en.wikipedia.org/wiki/Alpha_compositing#Description) seems pretty straightforward.

This is required to be able to do contrast calculations properly (see https://github.com/w3c/csswg-drafts/issues/7358 )

LeaVerou:
Closes #230 

Currently WIP, opening a draft PR so we can iterate.

### DIFF
--- a/src/index-fn.js
+++ b/src/index-fn.js
@@ -15,6 +15,7 @@ export {default as inGamut}   from "./inGamut.js";
 export {default as toGamut}   from "./toGamut.js";
 export {default as distance}  from "./distance.js";
 export {default as equals}    from "./equals.js";
+export {default as over}  from "./over.js";
 export {default as contrast}  from "./contrast.js";
 export {default as clone}     from "./clone.js";
 export {
--- a/src/index.js
+++ b/src/index.js
@@ -15,6 +15,9 @@ Color.extend(deltaE);
 import * as variations from "./variations.js";
 Color.extend(variations);
 
+import over from "./over.js";
+Color.extend(over);
+
 import contrast from "./contrast.js";
 Color.extend(contrast);
 
--- a/src/over.js
+++ b/src/over.js
@@ -1,0 +1,38 @@
+import getColor from "./getColor.js";
+import xyz_d65 from "./spaces/xyz-d65.js";
+import to from "./to.js";
+
+export default function over (source, backdrop, {
+	space = xyz_d65,
+	outputSpace = source.space,
+} = {}) {
+	source = getColor(source);
+	backdrop = getColor(backdrop);
+
+	let result;
+
+	if (source.alpha === 0) {
+		result = backdrop;
+	}
+	else if (source.alpha === 1 || backdrop.alpha === 0) {
+		result = source;
+	}
+	else {
+		let source_xyz = to(source, space);
+		let backdrop_xyz = to(backdrop, space);
+		result = {
+			space,
+			coords: source_xyz.map((s, i) => {
+				let b = backdrop_xyz[i];
+				return s + b * (1 - source.alpha);
+			}),
+			alpha: source.alpha + backdrop.alpha * (1 - source.alpha)
+		}
+	}
+
+	return to(result, outputSpace);
+}
+
+export function register (Color) {
+	Color.defineFunction("over", over, {returns: "color"});
+}
--- a/src/over.js
+++ b/src/over.js
@@ -1,4 +1,5 @@
 import getColor from "./getColor.js";
+import ColorSpace from "./space.js";
 import xyz_d65 from "./spaces/xyz-d65.js";
 import to from "./to.js";
 
@@ -8,6 +9,13 @@ export default function over (source, backdrop, {
 } = {}) {
 	source = getColor(source);
 	backdrop = getColor(backdrop);
+
+	space = ColorSpace.get(space);
+	outputSpace = ColorSpace.get(outputSpace);
+
+	if (space.isPolar) {
+		throw new Error("Compositing in polar color spaces is not supported.");
+	}
 
 	let result;
 
@@ -20,6 +28,7 @@ export default function over (source, backdrop, {
 	else {
 		let source_xyz = to(source, space);
 		let backdrop_xyz = to(backdrop, space);
+
 		result = {
 			space,
 			coords: source_xyz.map((s, i) => {
--- a/types/src/over.d.ts
+++ b/types/src/over.d.ts
@@ -1,0 +1,12 @@
+import Color, { ColorTypes } from "./color";
+
+export interface OverOptions {
+	space?: string | ColorSpace | undefined;
+	outputSpace?: string | ColorSpace | undefined;
+}
+
+export function over(
+	source: ColorTypes,
+	backdrop: ColorTypes,
+	options?: OverOptions
+) : ColorTypes


facelessuser:
I don't know if it was intentional, but pre-multiplication was left out. This is actually what browsers do when overlaying colors. If this is not desired, feel free to disregard it.

If another compositing is desired in the future, additionally you would likely want to clamp the resultant alpha, but source over shouldn't yield an alpha greater than 1 or less than zero unless the input alpha were somehow out of range (I assume Colorjs.io doesn't allow this, but maybe I'm wrong).

```diff
diff --git a/src/over.js b/src/over.js
index 20b4a1e..40851b6 100644
--- a/src/over.js
+++ b/src/over.js
@@ -20,13 +20,14 @@ export default function over (source, backdrop, {
        else {
                let source_xyz = to(source, space);
                let backdrop_xyz = to(backdrop, space);
+               let alpha = source.alpha + backdrop.alpha * (1 - source.alpha)
                result = {
                        space,
-                       coords: source_xyz.map((s, i) => {
-                               let b = backdrop_xyz[i];
-                               return s + b * (1 - source.alpha);
+                       coords: source_xyz.coords.map((s, i) => {
+                               let b = backdrop_xyz.coords[i];
+                               return (s * source.alpha + b * backdrop.alpha * (1 - source.alpha)) / (alpha ? alpha : 1);
                        }),
-                       alpha: source.alpha + backdrop.alpha * (1 - source.alpha)
+                       alpha: alpha
                }
        }
```

Additionally, `over` seems to not return a `Color` object which I imagine was a mistake. I think it is due to how `to()` returns stuff, but I didn't dig much deeper than that.

Makes chaining `over` impossible currently.

LeaVerou:
This is why this PR is draft :) Yes, there should absolutely be a `premultiplied` option, not sure whether it should be the default though. I agree with your proposed patch, but it doesn't show me the button to commit it, did you post it as a suggestion or just as a diff?

facelessuser:
> but it doesn't show me the button to commit it, did you post it as a suggestion or just as a diff?

I honestly don't know how to submit a diff that is auto-committable 😅. That is just a code block, sorry.

>  Yes, there should absolutely be a premultiplied option, not sure whether it should be the default though.

Everything I've read about compositing usually suggests premultiplying, but I see no harm in making it optional.

---

On a side note, there probably should be some accounting for `none` values as well. I imagine just translating them to zero should suffice. I wouldn't think there would be a need to preserve them across the alpha blending, but if that is desired, then I guess you'd need some additional logic to do that as well.

LeaVerou:
> I honestly don't know how to submit a diff that is auto-committable 😅. That is just a code block, sorry.

This button:
image

> Everything I've read about compositing usually suggests premultiplying, but I see no harm in making it optional.

👍🏼 @svgeesus ?

> On a side note, there probably should be some accounting for none values as well.

Good point!

svgeesus:
There should be code for premultiplication, and it should be the default because [CSS Color 4 requires premultiplication when interpolating non-opaque colors](https://drafts.csswg.org/css-color-4/#interpolation-alpha).

svgeesus:
> On a side note, there probably should be some accounting for none values as well. 

Handling interpolation with missing values is [also defined in CSS Color 4](https://drafts.csswg.org/css-color-4/#interpolation-missing)

> I imagine just translating them to zero should suffice.

No, because

>> Thus, the first stage in interpolating two colors is to classify any [missing components](https://drafts.csswg.org/css-color-4/#missing-color-component) in the input colors, and compare them to the components of the [interpolation color space](https://drafts.csswg.org/css-color-4/#interpolation-color-space). If any analogous missing components are found, they will be carried forward and re-inserted in the converted color before linear interpolation takes place.

svgeesus:
@LeaVerou why is this re-implementing (a subset of) interpolation instead of calling the interpolation function?

@svgeesus why didn't you finish implementing premultiplication?? :)

```js
if (premultiplied) {
		// not coping with polar spaces yet
		color1.coords = color1.coords.map(c => c * color1.alpha);
		color2.coords = color2.coords.map(c => c * color2.alpha);
	}
```

LeaVerou:
@svgeesus It’s implementing alpha blending, not interpolation (or at least trying to).

facelessuser:
@LeaVerou  @svgeesus  I've had some time to revisit this to see if we can push this conversation forward.

Locally, I checked out this branch, rebased it, and made the following changes.

```diff
diff --git a/src/color.js b/src/color.js
index d78f6cf..cc25c80 100644
--- a/src/color.js
+++ b/src/color.js
@@ -17,6 +17,7 @@ import {
 	set,
 	setAll,
 	display,
+	over,
 } from "./index-fn.js";
 
 
@@ -185,6 +186,7 @@ Color.defineFunctions({
 	inGamut,
 	toGamut,
 	distance,
+	over,
 	toString: serialize,
 });
 
diff --git a/src/over.js b/src/over.js
index 22193fb..36e4874 100644
--- a/src/over.js
+++ b/src/over.js
@@ -1,12 +1,32 @@
 import getColor from "./getColor.js";
 import ColorSpace from "./space.js";
+import * as util from "./util.js";
 import xyz_d65 from "./spaces/xyz-d65.js";
 import to from "./to.js";
 
+function porterDuffSourceOver (cba, csa) {
+	// The normal way in which two colors are overlaid.
+
+	let fa = 1;
+	let fb = 1 - csa;
+
+	return {
+		co: (cb, cs) => {
+			return csa * fa * cs + cba * fb * cb;
+		},
+		ao: () => {
+			return csa * fa + cba * fb;
+		}
+	}
+}
+
+
 export default function over (source, backdrop, {
 	space = xyz_d65,
 	outputSpace = source.space,
 } = {}) {
+	// Colors should be an RGB like space.
+
 	source = getColor(source);
 	backdrop = getColor(backdrop);
 
@@ -17,31 +37,32 @@ export default function over (source, backdrop, {
 		throw new Error("Compositing in polar color spaces is not supported.");
 	}
 
-	let result;
-
-	if (source.alpha === 0) {
-		result = backdrop;
-	}
-	else if (source.alpha === 1 || backdrop.alpha === 0) {
-		result = source;
-	}
-	else {
-		let source_xyz = to(source, space);
-		let backdrop_xyz = to(backdrop, space);
-
-		result = {
-			space,
-			coords: source_xyz.map((s, i) => {
-				let b = backdrop_xyz[i];
-				return s + b * (1 - source.alpha);
-			}),
-			alpha: source.alpha + backdrop.alpha * (1 - source.alpha)
-		}
-	}
+	// This could be changed to use other methods if desired.
+	let compositor = porterDuffSourceOver(backdrop.alpha, source.alpha);
+	let cra = util.clamp(0.0, compositor.ao(), 1.0);
+	let source_xyz = to(source, space);
+	let backdrop_xyz = to(backdrop, space).coords;
+	let result = {
+		space,
+		coords: source_xyz.coords.map((s, i) => {
+			// Only normal blending is supported, so the result is the source.
+			// A blend function for different blend modes could be used.
+			// ```
+			// let cr = util.clamp(0.0, blend(backdrop_xyz[i], s), 1.0);
+			// ```
+			// The result of the mixing formula should be clamped, per
+			// https://www.w3.org/TR/compositing/#blending
+			let cr = util.clamp(0.0, s, 1.0);
+			// Apply compositing
+			cr = compositor.co(backdrop_xyz[i], cr);
+			if (cra !== 0 || cra !== 1) {
+				cr /= cra;
+			}
+			// Alpha blended color
+			return cr;
+		}),
+		alpha: cra
+	};
 
 	return to(result, outputSpace);
 }
-
-export function register (Color) {
-	Color.defineFunction("over", over, {returns: "color"});
-}
\ No newline at end of file
```

This should give results like what I have here, but we are only implementing source over compositing with normal blending, this provides simple alpha blending. This is expected to be performed in RGB gamuts. This would give results like I have here: https://facelessuser.github.io/coloraide/compositing/.

I'm willing run with this if this is what is desired, if not , we would need to clearly define what is wanted for me to push this forward. I've left it open to providing other blend modes and even other compositing approaches, but realize none of that may be wanted, but it is written in such a way that it could be if desired in the future.

Anyways, at the very least, hopefully, this can get discussions moving again.

facelessuser:
I assume it may be reasonable to enforce gamut mapping of colors entering the function as well. I assume if this was extended beyond RGB spaces, then some of the channel clamping rules may not apply in those scenarios. Also, if expanded beyond RGB gamuts, not all blending modes and composition operators would make sense, but as mentioned earlier, these may not be desired.

svgeesus:
Earlier I chided myself:

> @svgeesus why didn't you finish implementing premultiplication?? :)
> 
> ```js
> if (premultiplied) {
> 		// not coping with polar spaces yet
> 		color1.coords = color1.coords.map(c => c * color1.alpha);
> 		color2.coords = color2.coords.map(c => c * color2.alpha);
> 	}
> ```

Would be fixed by

 - https://github.com/color-js/color.js/pull/754

PR #233, opened 2022-10-31
Add prepack script to build files before release.

### LINKED ISSUE #232: Legacy builds are not included in NPM package.
New legacy builds were added in #215 to support older versions of Node etc, but it appears that the legacy built files are missing in the published NPM package for v0.4.1: https://unpkg.com/browse/colorjs.io@0.4.1/dist/

I'm not sure what the build/release process is for colorjs.io, but when I `npm pack` locally the files are included correctly -- perhaps `npm run build` wasn't run before the new release? In case that's the issue, I'll open a PR to address that with a `prepack` script.

jgerigmeyer:
Possibly fixes #232.

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -184,7 +184,7 @@ body > header nav > .menu, body > footer nav > .menu {
 
 body > header nav > .menu > ul, body > footer nav > .menu > ul {
 				position: absolute;
-				z-index: 2;
+				z-index: 1;
 				padding: 0;
 				margin: 0;
 				list-style: none;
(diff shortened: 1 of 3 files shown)


assets/css/style.css
```diff
@@ -184,7 +184,7 @@ body > header nav > .menu, body > footer nav > .menu {
 
 body > header nav > .menu > ul, body > footer nav > .menu > ul {
 				position: absolute;
-				z-index: 2;
+				z-index: 1;
```
LeaVerou, line 187: This seems unrelated?

jgerigmeyer: Yes, sorry. This (and the change in `package-lock.json`) were updated when I ran `npm run build` locally. I can revert them if you prefer?

PR #221, opened 2022-10-02
[types] Add type definitions from DefinitelyTyped/DefinitelyTyped#62510

MysteryBlokHed:
Closes #119

Moves types from the original DefinitelyTyped PR and adds some dependencies and workflows to test them.

This PR is still very much a work in progress, and I'm pretty sure there are some parts of the API that I haven't typed yet.

PR #236, opened 2022-11-03
[ci] Update conditions for dtslint workflow run

MysteryBlokHed:
The workflow added in #221 only runs on pull requests, not on pushes. This means that anyone directly pushing to the main branch without creating a pull request skips the job to run `dtslint`.

This PR updates the conditions so that it runs on both pushes and pull requests, as well as changes the conditions so that it only runs when files in the types/ folder are modified.


LeaVerou:
Thanks!

PR #237, opened 2022-11-04
Adjust format type and add serialize type test.

jgerigmeyer:
This adds a type test for the change in https://github.com/LeaVerou/color.js/commit/d0ab5d1fe61e43f37386f7634a0cc4aee15e2302, and also adjusts the `Format` type to account for the change in https://github.com/LeaVerou/color.js/commit/667c19a60d5a6c8c81e9d07bd217ebda385a10b2.

### DIFF
--- a/types/src/serialize.d.ts
+++ b/types/src/serialize.d.ts
@@ -1,5 +1,5 @@
 import { ColorTypes } from "./color";
-import { Format } from "./space"
+import { Format } from "./space";
 
 export interface Options {
 	precision?: number | undefined;
--- a/types/src/space.d.ts
+++ b/types/src/space.d.ts
@@ -13,6 +13,7 @@ export interface Format {
 	toGamut?: boolean | undefined;
 	commas?: boolean | undefined;
 	lastAlpha?: boolean | undefined;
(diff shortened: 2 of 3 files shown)


LeaVerou:
Thanks! LGTM, though I'm thinking of renaming `noAlpha` to `opaque` or `opaqueOnly` since negative names are an API design antipattern. Though "opaque" is a more obscure word, and that's also an antipattern. 🤔

jgerigmeyer:
> Thanks! LGTM, though I'm thinking of renaming `noAlpha` to `opaque` or `opaqueOnly` since negative names are an API design antipattern. Though "opaque" is a more obscure word, and that's also an antipattern. 🤔

Maybe `includeAlpha` and default to `true`?

LeaVerou:
> > Thanks! LGTM, though I'm thinking of renaming `noAlpha` to `opaque` or `opaqueOnly` since negative names are an API design antipattern. Though "opaque" is a more obscure word, and that's also an antipattern. 🤔
> 
> Maybe `includeAlpha` and default to `true`?

Booleans that default to `true` is another API Design antipattern, possibly a bigger one…

PR #239, opened 2022-11-05
Fix bundler compatibility

### LINKED ISSUE #235: Bundle compatibility: .contrast is not a function, .deltaE is not a function
## Bug

Using vite@3.2.2 or esbuild@0.15.12

## Reproduce

input.js
```js
import Color from 'colorjs.io'

const red = new Color('red')
const blue = new Color('blue')

// ERROR: .contrast is not a function
red.contrast(blue, 'WCAG21')
// ERROR: .deltaE is not a function
red.deltaE(blue, '2000')

// PASSED
// red.contrastWCAG21(blue)
// red.deltaE2000(blue)
```

bundle:

```bash
esbuild --bundle --minify input.js --outfile=bundle.js && node bundle.js
```

ambar:
Fixes #235 

It's not safe to access function name in bundled apps, such as (reproduced in):
- vite@3.2.2 / esbuild@0.15.12
- cra@5.0.1 / terser@5.15.1 
- next@13.0.1 / swc@1.2

This bug could also reproduced in `dist/color.global.min.js` (main branch, bundled by terser):

```
// eval dist/color.global.min.js 
Color.prototype.contrast -> undefined
```

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -119,10 +119,6 @@ export default class Color {
 	}
 
 	static defineFunction (name, code, o = code) {
-		if (arguments.length === 1) {
-			[name, code, o] = [arguments[0].name, arguments[0], arguments[0]];
-		}
-
 		let {instance = true, returns} = o;
 
 		let func = function (...args) {
@@ -167,12 +163,6 @@ export default class Color {
 	static extend (exports) {
 		if (exports.register) {
 			exports.register(Color);
-		}
-		else if (exports.default) {
-			Color.defineFunction(exports.default.name, exports.default);
-		}
-		else if (typeof exports === "function") {
-			Color.defineFunction(exports);
 		}
 		else {
 			// No register method, just add the module's functions
--- a/src/index.js
+++ b/src/index.js
@@ -5,18 +5,18 @@ import Color from "./color.js";
 import "./spaces/index.js";
 
 // Import all DeltaE methods
-import * as deltaE from "./deltaE.js";
+import deltaE from "./deltaE.js";
 import * as deltaEMethods from "./deltaE/index.js";
 
 Color.extend(deltaEMethods);
-Color.extend(deltaE);
+Color.extend({deltaE});
 
 // Import optional modules
 import * as variations from "./variations.js";
 Color.extend(variations);
 
 import contrast from "./contrast.js";
-Color.extend(contrast);
+Color.extend({contrast});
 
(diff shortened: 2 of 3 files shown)


LeaVerou:
I'm fine to make the change, though I think it's a bug in these tools if they mess up `func.name`. But are you sure the only calls affected are these two? I'd expect it's a lot more. I need some time to look through the codebase to make sure there aren't more.

ambar:
I've checked all prototype properties, there're only two properties that got renamed:

image

LeaVerou:
@jgerigmeyer what do you think? Fine to merge if it LGTY.

jgerigmeyer:
Looks good on a quick skim, but I'd be happy to take a closer look tomorrow. 👍🏻

src/color.js
```diff
@@ -119,10 +119,6 @@ export default class Color {
 	}
 
 	static defineFunction (name, code, o = code) {
-		if (arguments.length === 1) {
-			[name, code, o] = [arguments[0].name, arguments[0], arguments[0]];
-		}
```
jgerigmeyer, line 124: Remove the corresponding type definition: https://github.com/LeaVerou/color.js/blob/main/types/src/color.d.ts#L93

src/color.js
```diff
@@ -168,12 +164,6 @@ export default class Color {
 		if (exports.register) {
 			exports.register(Color);
 		}
-		else if (exports.default) {
-			Color.defineFunction(exports.default.name, exports.default);
-		}
-		else if (typeof exports === "function") {
-			Color.defineFunction(exports);
-		}
```
jgerigmeyer, line 176: Remove the corresponding type definitions: https://github.com/LeaVerou/color.js/blob/main/types/src/color.d.ts#L105 and https://github.com/LeaVerou/color.js/blob/main/types/src/color.d.ts#L107

jgerigmeyer (changes requested):
@ambar Looks good! Thanks for this contribution. 🚀

One minor request, then this is ready to merge.

src/color.js
ambar: Done!

jgerigmeyer (approved):

carlobeltrame:
Thanks for the fix! Can we get a release which includes this?

LeaVerou:
Will try to do that tomorrow!

LeaVerou:
Done! Sorry I dropped the ball last week.

PR #240, opened 2022-11-08
[contrast] Add re-export of `src/contrast/index.js` to `src/index-fn.js`

MysteryBlokHed:
Adds a missing re-export of the contrast functions to `src/index-fn.js`.

From <https://github.com/LeaVerou/color.js/pull/162#issuecomment-1306232139>:

> > Is it correct that the named contrast functions e.g. `contrastAPCA` are not included in `colorjs.io/fn`?
> 
> They should be. If not, it's a bug. But I believe they are exported en masse from contrast/index-fn.js, so if you're searching the source for contrastAPCA it won't find anything.

I've updated the TypeScript tests to make sure the contrast methods are importable through `src/index-fn.js`, but I'm not sure if there's anything to change for the tests in `tests/`.

### DIFF
--- a/src/index-fn.js
+++ b/src/index-fn.js
@@ -21,6 +21,7 @@ export {
   getLuminance, setLuminance
 }                             from "./luminance.js";
 export {uv, xy}                             from "./chromaticity.js";
+export *                      from "./contrast/index.js";
 export {default as deltaE}    from "./deltaE.js";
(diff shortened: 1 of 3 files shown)


LeaVerou (approved):

manuelmeister:
@LeaVerou When can we expect a new release, with all the changes?

PR #241, opened 2022-11-12
Update outdated wiki links in README.md

MysteryBlokHed:
Changes a couple of documentation links in the README that use relative links instead of links to the website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@ Modules for <a href="https://colorjs.io/docs/spaces.html">a wide variety of colo
 including Lab/LCh, OKLab/OKLCh,
 sRGB and friends (HSL/HSV/HWB), Display P3,
 J<sub>z</sub>a<sub>z</sub>b<sub>z</sub>, REC.2100 and many <a href="https://colorjs.io/docs/spaces.html">more</a>.
-- **Doesn't gloss over color science**: Actual <a href="docs/gamut-mapping.html">gamut mapping</a> instead of naïve clipping,
+- **Doesn't gloss over color science**: Actual <a href="https://colorjs.io/docs/gamut-mapping.html">gamut mapping</a> instead of naïve clipping,
 multiple <a href="https://colorjs.io/docs/color-difference.html">DeltaE</a> methods (76, CMC, 2000, J<sub>z</sub>),
 multiple <a href="https://colorjs.io/docs/adaptation.html">chromatic adaptation</a> methods (von Kries, Bradford, CAT02, CAT16),
 all with sensible defaults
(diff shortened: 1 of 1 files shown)


LeaVerou:
What if we use links that start with `/`? I'd rather avoid absolute links if possible.

MysteryBlokHed:
I could change the links to point to the repository’s `.md` files in the docs folder, if that’s preferred? i.e.:

```diff
- **Doesn't gloss over color science**: Actual <a href="docs/gamut-mapping.html">gamut mapping</a> instead of naïve clipping,
+ **Doesn't gloss over color science**: Actual <a href="docs/gamut-mapping.md">gamut mapping</a> instead of naïve clipping,
```

LeaVerou:
Wouldn't that make them point to `.md` files in the website as well?

MysteryBlokHed:
Does the built documentation include the main README file? If it does, then I’m not sure if there’s any way to make relative links work properly, since they’d have to go to `.md` files on the repository, and `.html` files on the website.

It might be possible to use relative links if the website redirected URLs ending in `.md` to `.html`?

Myndex:
> Wouldn't that make them point to `.md` files in the website as well?

If _no_ extension is put on the file name, then the HTML server should serve the file with the .html, and GitHub will serve the file with .md, so 

`docs/gamut-mapping` 

Should work for both, it does on the Jekyll-built sites I'm hosting here, and for colorjs would seem just a matter of configuring Eleventy to use the extension that Apache is configured for (i.e. htm or html or php or whatever)

LeaVerou:
Some ways to address this:
- We could instead follow the convention of having `README.md` in its own folder, compiled to `index.html` and just point links to folders (with trailing `/`)
- We could add Netlify redirects that rewrite `foo.md` to `foo/index.html`, but that means we now have no way to actually request MD files (which may be ok?)

Myndex:
I don't know Netifly, but it is seamless with Jekyll.

All my files use .md and Jekyll turns all the .md into .html to serve (with the code inside also being converted of course). The way the GitHub pages server is setup here (and the way my own Apache servers are setup other places) only the file name, with no extension, is needed, and in the case of multiple file names with different extensions, the conf file lists the priority order.

E.g.` .php, .htm, .html, .txt ... `

All the _links_ in my pages, or in other pages linking here, therefore do not have an extension at all on them, and the server(s) determine which to serve per context.

**_So this:_**
https://myndex.github.io/colorparsley/versionHistory

**_Is automatically displaying this, without displaying the extension:_**
https://myndex.github.io/colorparsley/versionHistory.html

But if you add .md
https://myndex.github.io/colorparsley/versionHistory.md

Then you still get the unparsed markdown file raw.... 

Which you may or may not want, but you may want to access your .txt or other files from the web interface, which is possible at least in the GitHub environment. My point being that you don't need a single file in a folder trick, because you don't need extensions on _any_ URL used as a link, the way Jekyll-build pages are configured.

For the server, it's trivial to setup in Apache. I have to think that Netifly could be so configured to do it's voodoo similarly...

PR #250, opened 2022-12-02
Allow to extend the deltaE methods

latin-1:
```js
Color.deltaEMethods.deltaECustom = (color, sample) => { ... };
color1.deltaE(color2, {method: "custom"});
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -10,6 +10,7 @@ import * as deltaEMethods from "./deltaE/index.js";
 
 Color.extend(deltaEMethods);
 Color.extend({deltaE});
+Object.assign(Color, {deltaEMethods});
 
 // Import optional modules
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hey, sorry for the delay but we just hadn't gotten to this, it's not that we were against it! Could you please reopen?

latin-1:
I simply forgot I had this PR open 🤣

LeaVerou (changes requested):
Good job spotting the issue that the deltaE methods cannot be extended right now! We definitely need to fix this.
However, this way would only work for the OO API, and not the procedural one.
I'd prefer to find a different solution, possibly an exported object from `deltaE/index.js` that people would add their methods to?

latin-1:
Closing in favor of #280

PR #251, opened 2022-12-02
Fix parameter type of Color.defineFunction

latin-1:
The function may return something else. For example: `deltaE` returns a number.


MysteryBlokHed (approved):
LGTM! Functions passed to `Color.defineFunction` don't have anything preventing them from returning any value.

LeaVerou:
Thank you both!

PR #252, opened 2022-12-02
Fix Color constructor parameter types

latin-1:


types/src/color.d.ts
```diff
@@ -22,9 +22,10 @@ import {
 export type Coords = [number, number, number];
 
 export interface ColorObject {
-	space: ColorSpace;
+	spaceId?: string | ColorSpace | undefined;
+	space?: string | ColorSpace | undefined;
 	coords: Coords;
-	alpha?: number;
+	alpha?: number | undefined;
```
LeaVerou, line 28: Does it need `| undefined` if it's marked optional?

LeaVerou:
LGTM, @MysteryBlokHed @jgerigmeyer do you know if optional keys still need `| undefined`?

MysteryBlokHed:
It's a good idea to explicitly add `| undefined` to optional properties because of TypeScript's `exactOptionalPropertyTypes` config option, which treats optional properties differently than properties with a value of `undefined`: <https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types>

LeaVerou:
I see, thanks! And thanks @latin-1 for the PR. Merging then.

PR #259, opened 2022-12-05
[types] Merge `Color` types

### LINKED ISSUE #258: There's two type definitions of Color instance
```ts
import Color from "colorjs.io";

const color1 = new Color("black"); // Color from "colorjs.io/types/src/index.d.ts"
const color2 = color1.to("srgb"); // Color "from colorjs.io/types/src/color.d.ts"

// @ts-expect-error Type 'Color' is missing the following properties from type 'Color': uv, xy, contrast, contrastWCAG21, and 18 more.
const color3: Color = color2;
```

MysteryBlokHed:
Closes #258

The types originally had two definitions for the `Color` type which weren't interchangeable. This PR changes the types to more accurately reflect the original JS by augmenting the existing `Color` type when `src/index.d.ts` is imported.

There is a problem with how I've done this right now: the module augmentation is happening even when the normal `src/color.d.ts` version of the class is imported. I'm not sure how this is happening without `src/index.d.ts` even being referenced, so I'll have to debug that when I get the chance. Any help to make sure the augmentation only happens when `src/index.d.ts` is imported would be appreciated.

You can look at the failing test to see the problem: <https://github.com/LeaVerou/color.js/pull/259/files#diff-5bb8dd015c2660f438205df14840aa404517154744bcf054146734f0d5f8ad26>

### DIFF
--- a/types/src/index.d.ts
+++ b/types/src/index.d.ts
@@ -1,6 +1,6 @@
-import { uv, xy } from "../src/chromaticity";
-import ColorImport, { ToColorPrototype } from "../src/color";
-import contrast from "../src/contrast";
+import { uv, xy } from "./chromaticity";
+import Color from "./color";
+import contrast from "./contrast";
 import {
 	contrastWCAG21,
 	contrastAPCA,
@@ -8,8 +8,8 @@ import {
 	contrastWeber,
 	contrastLstar,
 	contrastDeltaPhi,
-} from "../src/contrast/index";
-import deltaE from "../src/deltaE";
+} from "./contrast/index";
+import deltaE from "./deltaE";
 import {
 	deltaE76,
 	deltaECMC,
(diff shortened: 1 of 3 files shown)


MysteryBlokHed:
I completely forgot I opened this PR 😅
From my understanding of TypeScript, it’s not possible to only augment types when a certain file is imported, which is what I was trying to do since the `src/index.js` file modifies the original `Color` import.

The only solution I can think of is to merge this as-is. The only problem with that is it will cause properties that only exist when `src/index.js` is imported to always be present in the types, even when they don’t actually exist in the current context.

latin-1 (approved):

LeaVerou:
I’m about to release a new version, should I merge this in?

MysteryBlokHed:
I think it's ready to be merged. The only problem I noted is that some properties that only exist after importing `src/index.js` instead of `src/color.js` are always present in the types, but I don't see any way around that 🤷

LeaVerou:
> I think it's ready to be merged. The only problem I noted is that some properties that only exist after importing `src/index.js` instead of `src/color.js` are always present in the types, but I don't see any way around that 🤷

Yeah, TS doesn't do that well with very dynamic JS. I guess it's better if they always exist than if they never exist.

jgerigmeyer (approved):

PR #265, opened 2023-01-04
Fix missing value parameter in setLuminance

xiegeo:


LeaVerou:
Wow, how did this get undetected for so long! Thank you!

Issue #190, opened 2022-07-02
Our rec2100-hlg is wrong


svgeesus:
I see from the convert app that the results for rec2100-hlg are very wrong.

I thought I knew how to fix it, looking at BT.2390, but I tried it and I don't. 

_(Note that BT.2390 expects the linear-light intensity range Edash to be 0 to 1 over the HDR range, while our code puts 0-1 over the SDR range so media white is at 1. Even accounting for that, I still couldn't fix it)._

It seems from [this email](https://lists.w3.org/Archives/Public/public-colorweb/2021Sep/0008.html) that things are more complex than I thought.

Explorations meanwhile are [in a separate repo](https://github.com/svgeesus/HLG-playground/blob/main/HLG.js) and I want to get some expert review from BBC folks before that comes back to color.js.

In particular there will probably need to be an options object with

 - Lw, the display monitor peak luminance (default 1000)
 - Lb, the display monitor black luminance (default 0.05)
 - Lamb, the ambient luminance (default 5)

to cope with non-reference monitor and non-reference ambient illumination.

**Meanwhile here be dragons and don't use our rec2100-hlg for anything important.**

LeaVerou:
You might want to add a note to the JSON that generates [our color space docs](https://colorjs.io/docs/spaces.html).

svgeesus:
Now [fixed](https://github.com/LeaVerou/color.js/commit/f3791da5ccdb3865c5b9cffd893d111faf21bb69) and gives the results expected from ITU-R BT.2390-10 page 25:

> Report ITU-R BT.2408 indicates that, for HLG HDR, diffuse white should be set at a signal level of 75%. This can be configured by making the output from an 18% grey card correspond to a signal level of 38%, rather than the 42.5% stated above.

 - [white gives `color(rec2100-hlg 0.75 0.75 0.75)`](https://colorjs.io/apps/convert/?color=white&precision=5)
 - [`color(rec2020-linear 0.18 0.18 0.18)` gives `color(rec2100-hlg 0.3782 0.3782 0.3782)`](https://colorjs.io/apps/convert/?color=color(rec2020-linear%200.18%200.18%200.18)&precision=5)

svgeesus:
The scaling factor assumes an overall system gamma of 1, and should be a default which is adjustable for different viewing conditions. For now, it is correct for standard viewing conditions.

PR #271, opened 2023-01-16
Correct rec2100-hlg, #190

svgeesus:
Fix wildly incorrect REC.2100 HLG encoding and decoding:

 - Parts of the code assumed [0,1] was SDR range, others that it was HDR range
 - The EOTF and OEFT from BT.2390-10 were the wrong way round
 - implemented a scaling factor to place 18% grey at 38% and SDR white at 75%
 - also fixed parse error in `color(rec2020-linear ...)`


LeaVerou (approved):
LGTM. Obviously I haven't really checked the math, but the actual code changes look good.

PR #279, opened 2023-02-03
Fix Rec 2100 HLG algorithm

facelessuser:
As you can see, the round trip is busted:

```js
> new Color('srgb', [0.6, 0.6, 0.6]).to('rec2100hlg').to('srgb').toString()
'rgb(58.775% 58.775% 58.775%)'
```

With the proposed fix:

```js
> new Color('srgb', [0.6, 0.6, 0.6]).to('rec2100hlg').to('srgb').toString()
'rgb(60% 60% 60%)'
```


LeaVerou:
Thank you!!

PR #289, opened 2023-02-25
[types] Fix clone signature

### LINKED ISSUE #288: Incorrect type definition for clone()
I think the type definition for the clone function should be changed from 
```ts
export default function clone(color: Color): Color;
```
to
```ts
export default function clone(color: ColorObject): ColorObject;
```

The clone method doesn't copy the spaceId property so ColorObject isn't 100% technically correct but it's more correct than Color.

I would have created a PR but I couldn't figure out how to get the type tests to run.

MysteryBlokHed:
Closes #288

Fixes the types for the `clone` function. The function only returns a few keys from the passed object, not a new class:

<https://github.com/LeaVerou/color.js/blob/e939c7cc70b473a9bfe07fe9d8c4c1c46ef3f754/src/clone.js#L2-L6>

### DIFF
--- a/types/src/clone.d.ts
+++ b/types/src/clone.d.ts
@@ -1,3 +1,3 @@
-import Color from "./color";
+import Color, { ColorObject } from "./color";
 
-export default function clone(color: Color): Color;
+export default function clone(color: Color | ColorObject): ColorObject;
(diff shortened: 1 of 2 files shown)


LeaVerou:
Thanks!

LeaVerou (approved):

MysteryBlokHed:
Note: This PR won’t be relevant if #291 is merged

PR #290, opened 2023-03-01
[toGamut] Ensure toGamut always returns a color object

lloydk:
Calling ```toGamut()``` with a color that is already in gamut for the specified color space would return the original color parameter which may or may not be a color object.

For example calling ```toGamut("red", "oklch")``` would return "red" instead of a color object.

This change ensures that a color object will always be returned.


LeaVerou:
Nice catch, thanks!

PR #291, opened 2023-03-01
Add a PlainColorObject type for functions that return color objects

lloydk:
The ```ColorObject``` type has optional properties and doesn't accurately reflect the color object type returned by functions such as ```getColor()``` which ensures that the object returned has space, coords and alpha properties that are not undefined.

A new interface called ```PlainColorObject``` has been introduced to represent fully formed color objects.

The ```Color``` class implements the ```PlainColorObject``` interface so the ```Color``` class was removed from list of types in the ```ColorTypes``` type.

This PR also fixes the return type for functions in the interpolation module that were returning the incorrect type.

### DIFF
--- a/types/src/color.d.ts
+++ b/types/src/color.d.ts
@@ -28,13 +28,19 @@ export interface ColorObject {
 	alpha?: number | undefined;
 }
 
+export interface PlainColorObject {
+	space: ColorSpace;
+	coords: Coords;
+	alpha: number;
+}
+
 export interface ColorConstructor {
 	spaceId: string;
 	coords: Coords;
 	alpha: number | undefined;
 }
 
-export type ColorTypes = Color | ColorObject | ColorConstructor | string;
+export type ColorTypes = ColorObject | ColorConstructor | string | PlainColorObject;
 
 export type DefineFunctionCode = (...args: any[]) => any;
 
@@ -79,7 +85,7 @@ declare namespace Color {
 	export const spaces: typeof ColorSpace["registry"];
 }
 
-declare class Color {
+declare class Color implements PlainColorObject {
 	constructor(color: ColorTypes);
 	constructor(space: string | ColorSpace, coords: Coords, alpha?: number);
 
--- a/types/src/clone.d.ts
+++ b/types/src/clone.d.ts
@@ -1,3 +1,3 @@
-import Color from "./color";
+import { PlainColorObject } from "./color";
 
-export default function clone(color: Color): Color;
+export default function clone(color: PlainColorObject): PlainColorObject;
--- a/types/src/display.d.ts
+++ b/types/src/display.d.ts
@@ -1,7 +1,7 @@
-import Color, { ColorObject, ColorTypes } from "./color";
+import { PlainColorObject, ColorTypes } from "./color";
 import ColorSpace from "./space";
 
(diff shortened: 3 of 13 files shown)


LeaVerou:
Hey, thanks for the contribution! I've requested review from some more TS-savvy folks than me. Meanwhile, you could rebase this PR, as it now shows changes from your other, merged, PR.

lloydk:
I managed to remove the extra commit from my original pull request but I think I ended up creating a different mess. I have no idea what I'm doing when it comes to git commands. For now I'll just wait for the review and if everything is ok we can figure out what to do next.

jgerigmeyer (approved):
I like this! I think there may be a few more places where we can replace `ColorObject` with `PlainColorObject` (e.g. `Display.color` in `display.d.ts`). I haven't looked through all of them, but I expect there might be other places too?

types/src/color.d.ts
```diff
@@ -28,13 +28,19 @@ export interface ColorObject {
...
+}
+
 export interface ColorConstructor {
 	spaceId: string;
 	coords: Coords;
 	alpha: number | undefined;
 }
 
-export type ColorTypes = Color | ColorObject | ColorConstructor | string;
+export type ColorTypes = ColorObject | ColorConstructor | string | PlainColorObject;
```
MysteryBlokHed, line 43: Was it intended to remove `Color` from `ColorTypes` here?

lloydk: Yes, ```Color``` implements the ```PlainColorObject``` interface so a ```Color``` is always a ```PlainColorObject```.

lloydk:
> I like this! I think there may be a few more places where we can replace `ColorObject` with `PlainColorObject` (e.g. `Display.color` in `display.d.ts`). I haven't looked through all of them, but I expect there might be other places too?

I tried to find all of the functions that could be switched but it's quite possible I missed a few. 

The ```display``` function calls ```serialize``` which calls ```getColor``` so in the case of the ```display``` function ```ColorTypes``` is the correct type for the ```color``` parameter.

jgerigmeyer:
> The `display` function calls `serialize` which calls `getColor` so in the case of the `display` function `ColorTypes` is the correct type for the `color` parameter.

I meant the return value `Display`. Shouldn't it be defined as `export type Display = string & { color: PlainColorObject };`?

lloydk:
> > The `display` function calls `serialize` which calls `getColor` so in the case of the `display` function `ColorTypes` is the correct type for the `color` parameter.
> 
> I meant the return value `Display`. Shouldn't it be defined as `export type Display = string & { color: PlainColorObject };`?



> > The `display` function calls `serialize` which calls `getColor` so in the case of the `display` function `ColorTypes` is the correct type for the `color` parameter.
> 
> I meant the return value `Display`. Shouldn't it be defined as `export type Display = string & { color: PlainColorObject };`?

Sorry, misread your comment. You're correct and I've made the appropriate change to the `Display` type.

jgerigmeyer (approved):

LeaVerou:
Just realized this PR had not been merged because it was referenced by another issue. Sorry @lloydk! FWIW @jgerigmeyer @MysteryBlokHed if I tag you for review, I fully trust you to merge the PR when/if ready. 😊

PR #292, opened 2023-03-06
Fix color temperature of D65

nico:
See last paragraph of https://en.wikipedia.org/wiki/Standard_illuminant#Illuminant_series_D


svgeesus (approved):

svgeesus:
Thanks! Yes, that is correct due to the value of Plank's constant changing meanwhile.

PR #297, opened 2023-03-28
Wrong serialization result for rgba/hsla spaces, fixes #296

### LINKED ISSUE #296: Wrong serialization result for rgba/hsla spaces
There is this bug:
```js
const c = new Color("rgba(255,0,0,1)");
c.toString({format: "rgba"}); // rgba(100%, 0%, 0%), expected rgba(100%, 0%, 0%, 1)
```
the fix is to check `format.lastAlpha` flag in [the serialize.js function](https://github.com/LeaVerou/color.js/blob/cf124a4931914a2ac3bf33988fdf8e73170bd5bb/src/serialize.js#L80):
```js
let strAlpha = (format.lastAlpha || (alpha < 1 && !format.noAlpha))
    ? `${format.commas? "," : " /"} ${alpha}`
    : "";
```

However, is it necessary to have two flags to determine whether a given format should include alpha?
To my knowledge, there are only 3 cases:
1) alpha is required (`rgba`, `hsla`)
2) alpha is optional (`alpha < 1` `rgb`/`hsl` for CSS Color 4)
3) alpha is ignored (`rgb_number`)

So the final flag should have 3 values, this can be achieved by, for example: 
```ts
interface Format {
    /**
     * true - alpha always present in serialized string
     * undefined - default, alpha present when < 1
     * false - no alpha in serialized string
     */
    withAlpha?: boolean | undefined;
    // or more descriptively by string union:
    withAlpha?: "always" | "auto" | "never";
}
```
```
{ noAlpha: true } => { withAlpha: false }
{ lastAlpha: true } => { withAlpha: true }
```

Is there something I don't know about or can I make PR to merge `noAlpha` and `lastAlpha` flags into one?

kajkal:
#296 fix

```js
const c = new Color("rgba(255,0,0,1)");

// before:
c.toString({format: "rgba"}); // rgba(100%, 0%, 0%)
// after:
c.toString({format: "rgba"}); // rgba(100%, 0%, 0%, 1)
```

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -56,7 +56,7 @@ export default function parse (str) {
 				if (format && format.type === "function") {
 					let alpha = 1;
 
-					if (format.lastAlpha || util.last(env.parsed.args).alpha) {
+					if (format.withAlpha || util.last(env.parsed.args).alpha) {
 						alpha = env.parsed.args.pop();
 					}
 
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -2,7 +2,6 @@ import * as util from "./util.js";
 import ColorSpace from "./space.js";
 import defaults from "./defaults.js";
 import getColor from "./getColor.js";
-import to from "./to.js";
 import checkInGamut from "./inGamut.js";
 import toGamut from "./toGamut.js";
 import clone from "./clone.js";
@@ -77,7 +76,9 @@ export default function serialize (color, {
 			alpha = util.toPrecision(alpha, precision);
 		}
 
-		let strAlpha = color.alpha < 1 && !format.noAlpha? `${format.commas? "," : " /"} ${alpha}` : "";
+		let strAlpha = (format.withAlpha || (format.withAlpha === undefined && alpha < 1))
+			? `${format.commas? "," : " /"} ${alpha}`
+			: "";
 		ret = `${name}(${args.join(format.commas? ", " : " ")}${strAlpha})`;
 	}
 
--- a/src/spaces/hsl.js
+++ b/src/spaces/hsl.js
@@ -74,7 +74,7 @@ export default new ColorSpace({
 		"hsla": {
 			coords: ["<number> | <angle>", "<percentage>", "<percentage>"],
 			commas: true,
-			lastAlpha: true,
+			withAlpha: true,
 		}
 	},
 });
--- a/src/spaces/srgb.js
+++ b/src/spaces/srgb.js
@@ -47,18 +47,19 @@ export default new RGBColorSpace({
 			name: "rgb",
 			commas: true,
 			coords: coordGrammarNumber,
-			noAlpha: true,
+			withAlpha: false,
 		},
 		"color": { /* use defaults */ },
 		"rgba": {
 			coords: coordGrammar,
 			commas: true,
-			lastAlpha: true,
+			withAlpha: true,
 		},
 		"rgba_number": {
 			name: "rgba",
 			commas: true,
-			coords: coordGrammarNumber
+			coords: coordGrammarNumber,
+			withAlpha: true,
 		},
 		"hex": {
 			type: "custom",
--- a/tests/index.html
+++ b/tests/index.html
@@ -17,6 +17,7 @@
 	<ul id="tests">
 		<li><a href="parse.html">Parsing</a></li>
 		<li><a href="construct.html">Object construction</a></li>
+		<li><a href="serialize.html">Serialization</a></li>
 		<li><a href="conversions.html">Conversions</a></li>
 		<li><a href="delta.html">DeltaE</a></li>
 		<li><a href="gamut.html">Gamut mapping</a></li>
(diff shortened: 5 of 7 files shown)


LeaVerou:
Hi there,

Thank you for contributing and sorry it took so long to review this! 

Could you elaborate a bit on this new `withAlpha` parameter that replaced the two previous `noAlpha` and `lastAlpha` params? It seems like a single boolean param cannot represent the three different states of syntax:

- Mandatory alpha, e.g. `rgba(0, 0, 0, 0)`
- Optional alpha, e.g. `rgb(0 0 0 / 0)`
- Disallowed alpha, e.g. `rgb(0, 0, 0)`

OTOH if we hardcode that commas cannot have optional alpha, I suppose we *could* use a single param for this.

kajkal:
Hi,
a single optional boolean param can represent 3 states, at least in my mind.
I described my reasoning in more detail in #296.

kajkal:
By the way, I was reading the [specification ](https://www.w3.org/TR/css-color-4/#rgb-functions) (_Candidate Recommendation Draft_) recently and from the way I understand it `rgb()` and `rgba()` are equivalent to each other, i.e. that `rgba(0,0,0)` or `rgb(0,0,0,0)` or `rgba(0 0 0 / 0)` are according to my interpretation of the specification valid. What is your opinion on the matter? Should `rgba()` force uses to define alpha after `,` or force uses to set alpha in `rgb()` only after `/`?

LeaVerou:
Oh that's a good point, I had actually forgotten about that. 

I do think that most people using `format: "rgba"` expect "old-style" `rgba()` though, with mandatory alpha. Now whether we cater to that or not is a separate discussion. Unfortunately often people expect what they're familiar with, not what is allowed or supported. We even had comments from people complaining that we return percentages by default, because they thought only numbers 0-255 were valid.

kajkal:
I looked more closely at the code some time ago and don't remember much anymore, so my comments may not be quite accurate.

From the output format perspective:
Currently there are 3 formats that work very similarly: `"rgb_number"`, `"rgba"` and `"rgba_number"`.
I would suggest replacing `"rgba"` and `"rgba_number"` with a single `"legacy"` or `"css3"` that would return color in a format compliant with CSS color 3 - so commas and 0..255 for r,g,b, `rgb` when no alpha and `rgba` when with alpha - or always `rgba`.
And as for the `"rgb_number"` format, I would give the option to include alpha.

So to summarize the 3 output formats beyond keyword and hex:
`default`: `rgb(100% 80% 60% / 0.3)` 0%..100%
`rgb_number`: `rgb(255 204 153 / 0.3)` 0..255
`legacy`: `rgba(255,204,153,0.3)` 0..255

From the input format perspective:
`rgb()` and `rgba()` are interchangeable so maybe they can be replaced by one format? Or two if you would like to distinguish rgb(0..255 x3) from rgb(0%..100% x3). Speaking of which I don't know what to think about `"rgb(255 0 50%)"` with which color.js doesn't see any problems but is not compliant with the specification. I solved this exact problem quite recently in a custom parser [this way](https://github.com/kajkal/maplibre-gl-style-spec/blob/ac57c36c574c01302ac5ece742dd08cb79da74fb/src/util/parse_css_color.ts#L80-L83) - maybe you could introduce similar concept in your library by adding `separatorGrammar` alongside `coordGrammar`?

So to summarize the 3 input formats beyond keyword and hex:
`default`: `rgba?(100% 80% 60% / 0.3)` or `rgba?(100%,80%,60%,0.3)` or mix with `<number>` and discard `rgb_number`
`rgb_number`: `rgba?(255 204 153 / 0.3)` or `rgba?(255,204,153,0.3)`
`legacy`: served by both above 

Two additional comments:
1) [this regexp](https://github.com/LeaVerou/color.js/blob/c4fcbbec8bdb3ec7dee2338d25a16cb363407b45/src/spaces/srgb.js#L66) allows for `#fffffff` which according to the specification is not allowed
2) [this regexp](https://github.com/LeaVerou/color.js/blob/c4fcbbec8bdb3ec7dee2338d25a16cb363407b45/src/util.js#L59) does not allow for values like `+0.0` or `12e1` which I think should be [supported](https://drafts.csswg.org/css-values/#numbers).

These are my thoughts, I don't know if they interest you at all but for my own sake I leave them here.

kajkal:
Thanks for your reply. Yes you are right. The theoretical ~~`legacy1`~~/`css3` (CSS Color 3 compliant) format should use integers (not real) 0..255, and alpha as real 0..1 not %.

My initial motivation for creating this PR was to provide a reliable way to get a valid CSS Color 3 compliant `rgb()`/`rgba()` color value. I don't know if the change proposed here is desirable or if a larger revision of the `rgb` formats is planned to better match the current state of the CSS Color 4 specification. 
In other words: should I implement [@LeaVerou suggestion](https://github.com/LeaVerou/color.js/issues/296#issuecomment-1567482530) and change the type from `withAlpha?: boolean` to `alpha?: "always" | "never" | "auto"` or should I close this PR?

PR #299, opened 2023-04-10
Fix incorrect type definition for Range

### LINKED ISSUE #298: `Range` function definition returns a `number` instead of a `Color`
I'm trying to use `Color.range(color2)` to interpolate between two colors. However, TypeScript is complaining that the function returned from `range()` returns a `number` instead of a `Color`. [The documentation](https://colorjs.io/api/#Color#range) states that this function returns a `Color`, and in runtime, the object is indeed a `Color`.

Looking at the [TypeScript definition](https://github.com/LeaVerou/color.js/blob/cf124a4931914a2ac3bf33988fdf8e73170bd5bb/types/src/interpolation.d.ts#L5), it seems that this function is defined to return a number:

```typescript
export type Range = ((percentage: number) => number) & {
    rangeArgs: { colors: [Color, Color]; options: Record<string, any> };
};
```

Do I misunderstand the API, or is there a mistake in the TypeScript definition?

nsilvestri:
The function represented by Range should return a Color, not a number.

Fixes #298.

### DIFF
--- a/types/src/interpolation.d.ts
+++ b/types/src/interpolation.d.ts
@@ -2,7 +2,7 @@ import Color, { ColorTypes } from "./color";
 import ColorSpace from "./space";
 import { Methods } from "./deltaE/index";
 
-export type Range = ((percentage: number) => number) & {
+export type Range = ((percentage: number) => Color) & {
 	rangeArgs: { colors: [Color, Color]; options: Record<string, any> };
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):

LeaVerou:
Thank you!

PR #300, opened 2023-04-12
Fix oklch and oklab colorspace names

lloydk:
Change the oklch and oklab colorspace names so they align with the CSS Color Level 4 spec and the original oklab article.

https://www.w3.org/TR/css-color-4/
https://bottosson.github.io/posts/oklab/

### DIFF
--- a/src/spaces/oklab.js
+++ b/src/spaces/oklab.js
@@ -29,7 +29,7 @@ const LabtoLMS_M = [
 
 export default new ColorSpace({
 	id: "oklab",
-    name: "OKLab",
+    name: "Oklab",
     coords: {
 		l: {
 			refRange: [0, 1],
--- a/src/spaces/oklch.js
+++ b/src/spaces/oklch.js
@@ -4,7 +4,7 @@ import {constrain as constrainAngle} from "../angles.js";
 
 export default new ColorSpace({
 	id: "oklch",
-	name: "OKLCh",
+	name: "Oklch",
 	coords: {
 		l: {
 			refRange: [0, 1],


facelessuser:
All of the literature from bottosson seems to indicate the names as `Oklab` and `OkLCh`. Is there a plane where this is not the case?

lloydk:
> All of the literature from bottosson seems to indicate the names as `Oklab` and `OkLCh`. Is there a plane where this is not the case?

'Oklab' is what i've changed the oklab colorspace to. I see that in his post about [color pickers ](https://bottosson.github.io/posts/colorpicker/)he did use OkLCh. The CSS color level 4 spec uses Oklch. OkLCh seems to be the one space of the four (Okhsl, Okhsv, Oklab, OkLCh) that doesn't follow the pattern of Uppercase 'O' and lower case for the rest of the letters.

facelessuser:
Yes, that has been my observation as well. OkLCh seems to be the standout. Though, it does make sense as LCh is the common casing.

LeaVerou:
This seems like it may actually be a bug in the CSS specs. @svgeesus what do you think?

svgeesus:
Not a bug, this is the relevant issue:

 - https://github.com/w3c/csswg-drafts/issues/7880

There is a case to be made for a particular capitalization with lowercase `h` (to represent an angle) ass opposed to uppercase `H` (to represent a chord length, at that angle, and the mean Chroma, as used in deltaE2000) but remembering to type `OkLCh` and (CIE) `LCh` seemed like too much to ask; also note that with this scheme, the 'L" in Oklab becomes uppercase (like CIE Lab) while Björn consistently uses Oklab which is, now, what CSS Color 4 and 5 use, too.

PR #302, opened 2023-04-21
Add verbose option to parse()

LeaVerou:
The `verbose` option gets `parse()` to return not just the result of the parsing, but also metadata about how the parsing was done.
For now the only metadata is `formatId` (which can be resolved to a format object through `color.space.formats[formatId]`) but more may be added in the future.
This is useful in UIs that handle colors (e.g. I needed it in a color game because I wanted to provide a syntax hint).

A little unsure about whether it's a good idea to have a function that returns a different type based on an option, but having a separate `parseVerbose()` function that `parse()` calls seems overkill.

Requesting review from @jgerigmeyer and @MysteryBlokHed because I don't trust myself with TS definitions yet 😁 
I did change `ColorConstructor` to `ColorObject` there as it seemed more suitable?

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -2,13 +2,20 @@ import * as util from "./util.js";
 import hooks from "./hooks.js";
 import ColorSpace from "./space.js";
 
-// CSS color to Color object
-export default function parse (str) {
+/**
+ * Convert a CSS Color string to a color object
+ * @param {string} str
+ * @param {object} options
+ * @param {boolean} options.verbose - If true, return an object with more information
+ * @returns { Color | { color: Color, formatId: string } }
+ */
+export default function parse (str, {verbose} = {}) {
 	let env = {"str": String(str)?.trim()};
 	hooks.run("parse-start", env);
 
 	if (env.color) {
-		return env.color;
+		let color = env.color;
+		return verbose? {color} : env.color;
 	}
 
 	env.parsed = util.parseFunction(env.str);
@@ -32,7 +39,8 @@ export default function parse (str) {
 						// If less <number>s or <percentage>s are provided than parameters that the colorspace takes, the missing parameters default to 0. (This is particularly convenient for multichannel printers where the additional inks are spot colors or varnishes that most colors on the page won’t use.)
 						const coords = Object.keys(space.coords).map((_, i) => env.parsed.args[i] || 0);
 
-						return {spaceId: space.id, coords, alpha};
+						let color = {spaceId: space.id, coords, alpha};
+						return verbose? {color, formatId: "color"} : color;
 					}
 				}
 			}
@@ -47,6 +55,7 @@ export default function parse (str) {
 					didYouMean = `Did you mean color(${cssId})?`;
 				}
 			}
+
 			throw new TypeError(`Cannot parse color(${id}). ` + (didYouMean || "Missing a plugin?"));
 		}
 		else {
@@ -87,16 +96,17 @@ export default function parse (str) {
 							let toRange = coordMeta.range || coordMeta.refRange;
 
 							if (fromRange && toRange) {
-
 								coords[i] = util.mapRange(fromRange, toRange, coords[i]);
 							}
 						});
 					}
 
-					return {
+					let color = {
 						spaceId: space.id,
 						coords, alpha
 					};
+
+					return verbose? {color, formatId: format.name} : color;
 				}
 			}
 		}
@@ -104,7 +114,6 @@ export default function parse (str) {
 	else {
 		// Custom, colorspace-specific format
 		for (let space of ColorSpace.all) {
-
 			for (let formatId in space.formats) {
 				let format = space.formats[formatId];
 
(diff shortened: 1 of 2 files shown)


types/src/parse.d.ts
```diff
@@ -1,3 +1,15 @@
-import { ColorConstructor } from "./color";
+import { PlainColorObject } from "./color";
```
MysteryBlokHed, line 1: I think that the `ColorConstructor` type is right here. Here are the type definitions for both types:

<https://github.com/LeaVerou/color.js/blob/64430d9e56a025cd6dea750d7e0073caf4c37eec/types/src/color.d.ts#L31-L41>

Note that `PlainColorObject` has the key `space`, while `ColorConstructor` has the key `spaceId`.

MysteryBlokHed (commented):
Other than a question about `PlainColorObject` vs `ColorConstructor`, everything TypeScript LGTM 😄

It’d also be worth updating the TS tests at <https://github.com/LeaVerou/color.js/blob/main/types/test/parse.ts> for the new parameter, i.e. something like

```typescript
parse('foo', {})
parse('foo', { verbose: true })
```

src/parse.js
```diff
@@ -2,13 +2,20 @@ import * as util from "./util.js";
...
 import ColorSpace from "./space.js";
 
-// CSS color to Color object
-export default function parse (str) {
+/**
+ * Convert a CSS Color string to a color object
+ * @param {string} str
+ * @param {object} options
+ * @param {boolean} options.verbose - If true, return an object with more information
+ * @returns { Color | { color: Color, formatId: string } }
```
jgerigmeyer, line 10: It looks like `formatId` here is optional? And I think this is also missing `argTypes`.

```suggestion
 * @returns { Color | { color: Color, formatId?: string, argTypes?: { type: string, range: [number, number] }[] | undefined } }
```

types/src/parse.d.ts
```diff
@@ -1,3 +1,15 @@
...
 
-export default function parse(str: string): ColorConstructor;
+export interface Options {
+	verbose?: boolean | undefined;
+}
+
+export interface VerboseParseResult {
+	color: PlainColorObject;
+	formatId?: string | undefined;
+}
```
jgerigmeyer, line 10:
```suggestion
export interface VerboseParseResult {
	color: PlainColorObject;
	formatId?: string | undefined;
	argTypes?: {
		type: string;
		range: [number, number];
	}[] | undefined;
}
```

types/src/parse.d.ts
lloydk: I found it a bit annoying that the `parse` function doesn't return a `PlainColorObject`. I currently use `getColor` for parsing because it does the `spaceId` to `space` fixup and is typed to return a `PlainColorObject`.

LeaVerou:
Actually, I've decided against this design due to [this](https://github.com/w3ctag/design-principles/issues/131#issuecomment-1518454261) and am about to push a different one where the metadata object is passed as a parameter, allowing for the return type to always be the same.

LeaVerou:
Could you take another look?

types/src/parse.d.ts
LeaVerou: Not sure what distinction this is making, since the actual `Color` constructor accepts both `spaceId` and `space`. 

But `parse()` can definitely be made to return an object with both `space` and `spaceId`, I suppose I just wanted to avoid the duplication. This is orthogonal to this PR though

LeaVerou: I will revert to using ColorConstructor for now though, since that's more accurate, and we can discuss how to fix the broader issue after this PR is merged

bakkot:
@LeaVerou Hope you'll forgive a drive-by comment from someone with very little context; I saw this linked from https://github.com/w3ctag/design-principles/issues/131 and couldn't resist chiming in.

Anyway: in my experience, having an output parameter (i.e. an argument which gets mutated by the function as an additional form of output) is quite unusual among JS programs - it's really the sort of thing you'd see more often in C/C++. Usually functions in JS _either_ return a computed value _or_ mutate their parameters - one or the other, not both.

I think your original idea of having separate `parse` and `parseVerbose` functions is the most natural way of doing this. Functions aren't particularly expensive, and any documentation can be kept simple by deferring to that of `parse` - e.g.,  "like `parse`, but returns a `{ color: Color, meta: Metadata }` pair, where `Metadata` is an object which has the `formatId` used for the parse".

jgerigmeyer (approved):

LeaVerou:
Actually, thinking about it, there is also a (d) `parse()` as designed here, and `parseVerbose()` that just calls `parse()` with suitable arguments.
This is a superset of the proposed `parse()` and `parseVerbose()` API, and doesn't have the perf issues of (b). It's basically (c) without an extra hidden function.
However, if we go with this option, there is nothing preventing us from adding `parseVerbose()` later, if doing it via a `parse()` argument proves confusing to users. Therefore, I'm inclined to just merge this for now.

Issue #220, opened 2022-09-29
Support for the experimental syntax 'classPrivateMethods' isn't currently enabled


dungle-scrubs:
I'm attempting to use this library in a React CodeSandbox, but I get the error:

`Support for the experimental syntax 'classPrivateMethods' isn't currently enabled`.

I've tried adding `@babel/plugin-proposal-private-methods ` to my `.babelrc` but the error remains.

Any suggestions?

LeaVerou:
Hi there,

I think this is a React CodeSandbox issue, could you try creating an issue here? https://github.com/codesandbox/codesandbox-client/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc

Class private methods aren't exactly cutting edge at this point…

PR #306, opened 2023-04-28
Make Color.js proxy-friendly, resolves #305

### LINKED ISSUE #305: Color.js objects cannot be wrapped in a Proxy (and thus, cannot be used with Vue or other Proxy-based libraries)
Due to using private properties, every time a Color instance is wrapped in a Proxy, errors are just around the corner.

For context, see: 

- https://twitter.com/LeaVerou/status/1650562320702099457
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#no_private_property_forwarding
- https://github.com/vuejs/core/issues/8149
- https://github.com/tc39/proposal-class-fields/issues/106

We only use a single private property: [`#space`](https://github.com/LeaVerou/color.js/blob/main/src/color.js#L78), and it's primarily to ensure color space is immutable. We can just use `Object.defineProperty()` instead and be done with it.

Edit: There is actually a bunch more privates in space.js but nothing too hard to refactor out.

LeaVerou:


jgerigmeyer (approved):
@LeaVerou LGTM! I agree that comparing `this.id === space.id` isn't ideal... but I can't think of an obviously better alternative?

src/space.js
```diff
@@ -169,23 +125,16 @@ export default class ColorSpace {
...
-		let ret = [this];
-
-		for (let space = this; space = space.base;) {
-			ret.push(space);
-		}
-
-		return ret;
+	// We cannot rely on simple === because then ColorSpace objects cannot be proxied
+	equals (space) {
+		return this === space || this.id === space.id;
```
stof, line 137: shouldn't this logic be used in the `to()` method then to replace the `this === space` check ?

src/space.js
```diff
@@ -57,7 +57,12 @@ export default class ColorSpace {
 		this.referred = options.referred;
 
 		// Compute ancestors and store them, since they will never change
-		this.#path = this.#getPath().reverse();
+		Object.defineProperty(this, "path", {
```
stof, line 60: should this have some `@private` jsdoc to let IDEs know they should treat it as private ?

stof:
This will also close #220 due to not using private methods anymore

src/space.js
LeaVerou: It’s not private, just immutable. Private properties were used  in conjunction with a public getter to make it immutable.

LeaVerou: It doesn’t really need to be private, as long as it’s immutable.

src/space.js
LeaVerou: It …is? Line 162. Am I missing another place?

LeaVerou: Ah, thanks. Should be fixed now.

LeaVerou:
Ok, merging this with a heavy heart 💔

bakkot:
@LeaVerou There's a simple way to opt in to being transparent to proxies if you explicitly want that: access private fields by deferring to the actual object, instead of the current value of `this`.

```js
class Example {
  _ = this;
  #private = 42;
  method() {
    console.log(this._.#private);
  }
}

let proxiedExample = new Proxy(new Example, {});
console.log(proxiedExample.method()); // 42
```

That won't solve your problems with Vue specifically because it's not just a simple Proxy - it recursively Proxy every single property (including `_` in this example). (This incidentally requires a bunch of effort to work with built-in types and [only works with a small handful they've special-cased](https://github.com/vuejs/core/blob/4e2607a3d316b6d8784c1af5f6e96cc495b74416/packages/reactivity/src/reactive.ts#L43-L56).)

But you can interop with Vue specifically with a little more work, if you really want that:

```js
import { createApp } from 'https://unpkg.com/vue@3/dist/vue.esm-browser.js'

function makeVueNotProxyBox(value) {
  return {
    [Symbol.toStringTag]: 'box', // using a custom toStringTag will prevent Vue from trying to proxy this thing
    _: value,
  };
}
class Foo {
  #bar = 1;
  _ = makeVueNotProxyBox(this);
  
  get bar() {
    return this._._.#bar;
  }
}

createApp({
  data() {
    return {
      foo: new Foo()
    }
  }
}).mount("#app");
```

(This assumes you want to make a thing work with Vue without actually depending on Vue. There's easier ways if you're specifically designing Vue components.)

redzzy30:
> @LeaVerou There's a simple way to opt in to being transparent to proxies if you explicitly want that: access private fields by deferring to the actual object, instead of the current value of `this`. Similarly, compare equality by the actual object, not by the current value of `this`.
> 
> ```js
> class Example {
>   _ = this;
>   #private = 42;
>   method() {
>     console.log(this._.#private);
>   }
> }
> 
> let proxiedExample = new Proxy(new Example, {});
> console.log(proxiedExample.method()); // 42
> ```
> 
> That won't solve your problems with Vue specifically because it's not just a simple Proxy - it recursively Proxy's every single property (including `_` in this example). (This incidentally requires a bunch of effort to work with built-in types and [only works with a small handful they've special-cased](https://github.com/vuejs/core/blob/4e2607a3d316b6d8784c1af5f6e96cc495b74416/packages/reactivity/src/reactive.ts#L43-L56).)
> 
> But you can interop with Vue specifically with a little more work, if you really want that:
> 
> ```js
> import { createApp } from 'https://unpkg.com/vue@3/dist/vue.esm-browser.js'
> 
> function makeVueNotProxyBox(value) {
>   return {
>     [Symbol.toStringTag]: 'box', // using a custom toStringTag will prevent Vue from trying to proxy this thing
>     _: value,
>   };
> }
> class Foo {
>   #bar = 1;
>   _ = makeVueNotProxyBox(this);
>   
>   get bar() {
>     return this._._.#bar;
>   }
> }
> 
> createApp({
>   data() {
>     return {
>       foo: new Foo()
>     }
>   }
> }).mount("#app");
> ```
> 
> (This assumes you want to make a thing work with Vue without actually depending on Vue. There's easier ways if you're specifically designing Vue components.)

It's really sad that we have to resort to this. This should be a feature natively supported by JS, instead of having developers take care of it. If I knew private properties would be implemented like this, I'd rather not have them at all and just stick with the good old `_private_property`

bakkot:
@redzzy30 You have to write code like in my example anyway if you want `===` to work. That's a consequence of the design of Proxies.

PR #307, opened 2023-04-28
Use node v14 for CI.

jgerigmeyer:
Many of the tools used in CI no longer support node v12. Now that node v14 is the oldest maintained version, this runs CI against that version. (Other changes are unrelated, but came along with a `npm run build`.)

### DIFF
--- a/.github/workflows/test-types.yml
+++ b/.github/workflows/test-types.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: "12"
+          node-version: "14"
 
       - run: npm install
       - run: npm run dtslint


LeaVerou:
Whoa, 12. Even 14 is kinda old, can we update to 18?

LeaVerou (changes requested):
To keep things clean, could you remove the CSS files from this PR? We need to git rm them, we shouldn't have build artifacts in the repo…

jgerigmeyer:
> Whoa, 12. Even 14 is kinda old, can we update to 18?

Hm, I'm torn on that. One of the purposes of CI is to ensure that the package as a whole still works for older Node environments, and v14 is still in active maintenance mode. While I assume development will happen in node v18 or even v20, I think it's reasonable to run CI against an older supported version? It's a bit blurry with a package like this which is also intended for browser usage, but I don't see a good reason to skip node v14 unless/until it starts breaking things. 🤷 

> To keep things clean, could you remove the CSS files from this PR? We need to git rm them, we shouldn't have build artifacts in the repo…

Done! ✅

LeaVerou:
> > Whoa, 12. Even 14 is kinda old, can we update to 18?
> 
> Hm, I'm torn on that. One of the purposes of CI is to ensure that the package as a whole still works for older Node environments, and v14 is still in active maintenance mode. While I assume development will happen in node v18 or even v20, I think it's reasonable to run CI against an older supported version? It's a bit blurry with a package like this which is also intended for browser usage, but I don't see a good reason to skip node v14 unless/until it starts breaking things. 🤷
> 
> > To keep things clean, could you remove the CSS files from this PR? We need to git rm them, we shouldn't have build artifacts in the repo…
> 
> Done! ✅

I think it's primarily which version we use to build. Our testsuite currently doesn't even run on Node at all, it's entirely client-side. Anyhow, approving this since 14 is better than 12.

LeaVerou (approved):

PR #314, opened 2023-05-25
[spaces] Add percentages to color functions

jamesnw:
This adds percentage handling to:

- lch
- lab
- oklch
- oklab
- lab-d65

It defines percentage ranges in the `coords` section. 

This PR also adds tests to confirm that the percentage form works in these, as well as the predefined color spaces- no changes were needed for those.


svgeesus (approved):
Thanks, this all looks good

PR #318, opened 2023-06-18
[types] Add missing string option type in toGamut

manuelmeister:
Only types are added to match functionality
https://github.com/LeaVerou/color.js/blob/f22556161af0caf03bff447ff86b918d125ccbe4/src/toGamut.js#L23-L28

### DIFF
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -12,7 +12,7 @@ import getColor from "./getColor.js";
 /**
  * Force coordinates to be in gamut of a certain color space.
  * Mutates the color it is passed.
- * @param {Object} options
+ * @param {Object|string} options object or spaceId string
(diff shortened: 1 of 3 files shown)


manuelmeister:
@LeaVerou could you take a look at this? Otherwise it is creates errors in a typescript environment.

MysteryBlokHed (changes requested):
Could you add a test case to the end of [types/test/toGamut.ts](https://github.com/LeaVerou/color.js/blob/f505c1b5230e1eabf5893f6a8eefb335936e55e0/types/test/toGamut.ts) for passing a string as a parameter here?

Something like this:

```typescript
toGamut(new Color("red"), "srgb"); // $ExpectType PlainColorObject
```

MysteryBlokHed (approved):

PR #323, opened 2023-08-08
[types] Fix signature of `steps` function

### LINKED ISSUE #322: TypeScript error for `steps` and `PlainColorObject`
I discovered this error while using `Color.steps`. 

If you paste the following into the playground on colorjs.io (without Typescript) it functions as I'd expect:

```js
let color = new Color("p3", [0, 1, 0]);
color.steps("red", {
    space: "lch",
    outputSpace: "srgb",
    maxDeltaE: 3, // max deltaE between consecutive steps (optional)
    steps: 10 // min number of steps
}).map(step => step.toString({format: 'hex'}));
```
I get an array of hex codes of all the steps.

When I paste this into my Typescript project, I get the error `This expression is not callable. Type 'never' has no call signatures.` and shows `(property) Color.steps: never`

Additionally, if I try to make steps out of a range, I do the following:

```js
const color = new Color("p3", [0, 1, 0]);
const color2 = new Color("red");
const range = color.range(color2, {space: "lch",
outputSpace: "srgb",})
Color.steps(range, {
    maxDeltaE: 3, // max deltaE between consecutive steps (optional)
    steps: 10 // min number of steps
}).map(step => step.to("srgb").toString({format: 'hex'}));
```
I get an error `Property 'to' does not exist on type 'PlainColor'Object'.`

My workaround has been to do `Color.steps(range, {...}).map(step => new Color(step).....)`

MysteryBlokHed:
Fixes #322

This PR does two things:
- Changes the order of the `steps` overload so that the correct signature is used [here](https://github.com/LeaVerou/color.js/blob/2bd19b0a913da3f3310b9d8c1daa859ceb123c37/types/src/index.d.ts#L72)
- Changes the return type of `steps` back to the actual `Color` class

### DIFF
--- a/types/src/interpolation.d.ts
+++ b/types/src/interpolation.d.ts
@@ -48,7 +48,7 @@ export interface StepsOptions extends RangeOptions {
 	maxSteps?: number | undefined;
 }
 
-export function steps(color1: ColorTypes, color2: ColorTypes, options?: StepsOptions): PlainColorObject[];
-export function steps(range: Range, options?: StepsOptions): PlainColorObject[];
+export function steps(range: Range, options?: StepsOptions): Color[];
+export function steps(color1: ColorTypes, color2: ColorTypes, options?: StepsOptions): Color[];
 
 export function register(color: typeof Color): void;
--- a/types/test/interpolation.ts
+++ b/types/test/interpolation.ts
@@ -47,16 +47,16 @@ mix("red", "blue", {
 	premultiplied: true,
 });
 
-steps("red", "blue"); // $ExpectType PlainColorObject[]
-// $ExpectType PlainColorObject[]
+steps("red", "blue"); // $ExpectType Color[]
+// $ExpectType Color[]
 steps("red", "blue", {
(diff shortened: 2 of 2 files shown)


lloydk:
I don't think this change is correct. The `steps` function really does return `PlainColorObject[]` and the `steps` method on the `Color` class converts the `PlainColorObject[]` to `Color[]`

The `Color` class defines the `steps` methods here: https://github.com/LeaVerou/color.js/blob/2bd19b0a913da3f3310b9d8c1daa859ceb123c37/types/src/index.d.ts#L72
Those methods have the wrong return type (`PlainColorObject[]` instead of `Color[]`).

Not sure what the correct fix should be, maybe inline the method definitions instead of using `ToColorPrototype` so the return type can be changed to `Color[]`?

lloydk:
The `mix` methods on the `Color` class have the same issue. They have a return type of  `PlainColorObject` but the return type should be `Color`.

jgerigmeyer (approved):

PR #324, opened 2023-08-18
Adds support for more angle units

### LINKED ISSUE #180: Only `deg` is supported as an angular unit for hues

### LINKED ISSUE #311: Support grad, rad, and turn angle units for the hue value of hsl()

kleinfreund:
## Changes

Adds support for the angle units "grad", "rad", and "turn".

Resolves #180.

Resolves #311.

## Notes

I was implementing support for this in a project of mine and figured I contribute it here as well. I'm (now) aware of https://github.com/LeaVerou/color.js/pull/181, so let me know if you'd rather incorporate that. Happy to close this again, if so.

Also, I did run the parser tests locally and they passed, but I saw about 75 of the tests failing locally and I'm not sure whether that's just something with my environment, so it would be good to double check this.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -43,6 +43,13 @@ export function toPrecision (n, precision) {
 	}
 }
 
+const angleFactor = {
+	deg: 1,
+	grad: 0.9,
+	rad: 180/Math.PI,
+	turn: 360,
+};
+
 /**
 * Parse a CSS function, regardless of its name and arguments
 * @param String str String to parse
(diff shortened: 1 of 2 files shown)


LeaVerou (approved):
I love the simplicity of this solution, and that you added tests. Thank you!

PR #331, opened 2023-09-19
Remove gamut mapping from hsl, hwb and hsv

jamesnw:
This PR addresses https://github.com/LeaVerou/color.js/issues/325.

The issue doesn't mention hsv, but since it's based on hsl, I removed gamut mapping from there as well. This essentially is reverting 960b4ab878d0e1a8b31726df2a4805e3b5148ad6.


svgeesus (approved):
Great, thanks!

PR #337, opened 2023-10-11
Add CJS file to `/fn` entry

ambar:
- Add CJS file to `/fn` entry, which is especially useful in universal webpack apps, make it consistent with the main entry
- Change files pattern in package.json, some dist files are missing (min/global), see https://unpkg.com/browse/colorjs.io@0.4.5/dist/


jgerigmeyer:
@ambar Thanks! I merged this in with some related fixes in #349.

Issue #154, opened 2022-06-23
Implement CSS Color 4 gamut mapping


svgeesus:
Currently, gamut mapping tries to allow any coordinate in any color space to be reduced (including nonsensical ones like hue) and uses deltaE 2000 as a color distance metric.

- [x] Implement [proposed improved CSS color 4 gamut mapping](https://github.com/w3c/csswg-drafts/issues/7135#issuecomment-1066121795) in [a codepen](https://codepen.io/svgeesus/pen/GRxJRKz?editors=0010) (far end of JND MINDE)
- [x] Investigate [odd results with imaginary colors](https://github.com/w3c/csswg-drafts/issues/7071)
- [ ] Add to color.js codebase
- [ ] Make it the default GMA

See also [this codepen](https://colorjs.io/notebook/?storage=https%3A%2F%2Fgist.github.com%2Fsvgeesus%2F1ae7f34724b0d5e7ca99f217d52446d5) and [this demo](https://danburzo.ro/demos/color/gamut-mapping.html)

svgeesus:
Does not just affect GM single colors, but also affects smoothness of gradients, see [this awful example](https://github.com/w3c/csswg-drafts/issues/7071#issuecomment-1049417689)

svgeesus:
On `lab(50% -127 -127)` I get starting oklch of 
`[ 0.42226189321340196, 1.9540778145796625, 191.20237227483153 ]`
and after GM I get
`[ 0.43100803800138154, 0.0741924017206513, 191.0338159551419 ]`

So the chroma is vastly reduced but the hue is very similar.

svgeesus:
Exploring the CIE Lab L=50 plane over  a,b = -125 to +125 I am struck by the OKLab a value in the -125,-125 corner:

lab(50% 125 125)   =  oklab(60.58% 0.357 0.152)
lab(50% -125 125)  =  oklab(53.41% -0.32 0.194)
lab(50% -125 -125) =  oklab(42.27% **-1.9** -0.37)
lab(50% 125 -125)  =  oklab(65.4% 0.257 -0.35)

svgeesus:
So exploring the CIE a,b plane gives odd results if the GMA is using OKLCH

![image](https://user-images.githubusercontent.com/2506926/177008783-0e9cacc4-b026-4a65-aeb2-bc6a092935e7.png)


but equally, exploring the OKLab ab plane while mapping in CIE LCH is odd too

[...]

PR #344, opened 2023-11-14
Add CSS toGamut algorithm

jamesnw:
This implements the CSS Color 4 Algorithm, as demo'ed in #154, and specified in https://drafts.csswg.org/css-color/#css-gamut-mapping.

Remaining:

- [ ] Tests
- [ ] Documentation


src/defaults.js
```diff
@@ -2,7 +2,7 @@
 const hasDOM = typeof document !== "undefined";
 
 export default {
-	gamut_mapping: "lch.c",
```
jamesnw, line 5: This changes the default `gamut_mapping`, which is a breaking change. Do we instead want to phase this in?

src/toGamut.js
```diff
@@ -104,3 +112,84 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
 }
 
 toGamut.returns = "color";
+
```
jamesnw, line 115: I am wondering if this would have value apart from the Color.js `toGamut` call, as an exact representation of the algorithm? As it is, it is returning in the original space, whereas the CSS algorithm returns in the destination space. We could do that conversion later, and then export this function by itself, if that would be useful.

src/toGamut.js
```diff
@@ -104,3 +112,85 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
...
+	if (space.isUnbounded) return to(origin, space);
+
+	const origin_OKLCH = to(origin, ColorSpace.get('oklch'));
+	let L = origin_OKLCH.coords[0];
+
+	// return media white or black, if lightness is out of range
+	if (L >= 1) {
+		const white = to(parse('white'), space);
+		white.alpha = origin.alpha;
+		return to(white, returnSpace);
```
jamesnw, line 141: I want to confirm that this follows the spec's intention- It says `if the Lightness of origin_Oklch is greater than or equal to 100%, return { 1 1 1 origin.alpha } in destination`. I took that to mean "white" in the destination space, for instance `oklab(1 0 0)`, and not `oklab(1 1 1)` or even `oklab(1 .4 .4)`.

facelessuser: This was my interpretation as well: `color(srgb 1 1 1 / alpha)`. I imagine this requirement is SDR specific though. Granted, the CSS spec does not currently provide any advice related to handling HDR color spaces (which this library does have) with the algorithm. Maybe this project would prefer to treat all color spaces as SDR with the "CSS" gamut mapping algorithm? I'm not sure. I'm also not aware of an API call to determine if a space is SDR or HDR, but I imagine one could be added if this distinction is desired.

src/defaults.js
svgeesus: Removing it as an option would be a much more breaking change. But I don't know how much the arbitrary channel stuff is being used.

src/toGamut.js
svgeesus: From [CSS Color 4](https://drafts.csswg.org/css-color-4/#css-gamut-mapping):

> The CSS gamut mapping algorithm applies to individual, Standard Dynamic Range (SDR) CSS colors which are out of gamut of an RGB display and thus require to be css gamut mapped.

So yes it is explicitly for SDR inputs and returns an SDR RGB color. Which may be an issue for changing the default. On the other hand the old default of reduction in CIE LCH chroma was not especially useful for HDR inputs either.

Confirming that "{ 1 1 1 origin.alpha } in destination" means SDR (media) white.

src/toGamut.js
svgeesus: > We could do that conversion later, and then export this function by itself, if that would be useful.

That would be best, I think.

api/api.json
```diff
@@ -249,19 +249,13 @@
 							"name": "options.method",
 							"optional": true,
 							"type": "string",
-							"description": "How to force into gamut. If \"clip\", coordinates are just clipped to their reference range (don't do that unless you have a reason, it produces very poor results). If in the form [colorSpaceId].[coordName], that coordinate is reduced until the color is in gamut. Please note that this may produce nonsensical results for certain coordinates (e.g. hue) or infinite loops if reducing the coordinate never brings the color in gamut. Defaults to `\"lch.c\"`."
+							"description": "How to force into gamut. If \"css\", chroma is reduced in the Oklch space using the algorithm defined by CSS Color 4. If \"clip\", coordinates are just clipped to their reference range (don't do that unless you have a reason, it produces very poor results). If in the form [colorSpaceId].[coordName], that coordinate is reduced until the color is in gamut. Please note that this may produce nonsensical results for certain coordinates (e.g. hue) or infinite loops if reducing the coordinate never brings the color in gamut. Defaults to `\"css\"`."
```
svgeesus, line 252: This is better, I was always uneasy that clip was presented first here.

api/api.json
```diff
@@ -249,19 +249,13 @@
...
 						},
 						{
 							"name": "options.space",
 							"optional": true,
 							"type": "ColorSpace",
 							"description": "Color space whose gamut we are mapping to. Defaults to the current color space."
-						},
-						{
-							"name": "options.inPlace",
```
svgeesus, line 261: So this is removing the capability to gamut-map in-place. @LeaVerou should we retain this?

jamesnw: This option was removed in https://github.com/LeaVerou/color.js/issues/152 - or rather, the only behavior currently is "in place". This change was removing the option from the docs to match the code.

src/toGamut.js
jamesnw: I made this change in https://github.com/LeaVerou/color.js/pull/344/commits/fc6b08388b23fe99d378f96c2c9c995df86dbddf

src/defaults.js
LeaVerou: I think it's fine to change the default, but we should keep the channel method, at least for now.

src/toGamut.js
```diff
@@ -15,84 +17,91 @@ import getColor from "./getColor.js";
...
-	}
-
 	// 3 spaces:
 	// color.space: current color space
 	// space: space whose gamut we are mapping to
 	// mapSpace: space with the coord we're reducing
+
 	let spaceColor = to(color, space);
+	if (method === "css") {
+		spaceColor = to(toGamutCSS(color, space), color.space);
```
jgerigmeyer, line 41: @jamesnw Should this be:

```suggestion
		spaceColor = to(toGamutCSS(color, {space}), color.space);
```

types/src/toGamut.d.ts
```diff
@@ -14,3 +14,15 @@ declare function toGamut(
 ): PlainColorObject;
 
 export default toGamut;
+
+declare namespace toGamutCSS {
+	let returns: "color";
+}
```
jgerigmeyer, line 20: This isn't necessary for methods unless they're registered on Color objects (e.g. `myColor.toGamutCSS()`).

src/toGamut.js
```diff
@@ -104,3 +113,97 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
 }
 
 toGamut.returns = "color";
+
+/**
+ * toGamutCSS
+ *
```
jgerigmeyer, line 119: Minor, but I don't think elsewhere that the documentation includes the fn name?

```suggestion
```

svgeesus:
Time to un-draft this PR?

I see there are comments from @jgerigmeyer which should be addressed

src/toGamut.js
```diff
@@ -104,3 +113,95 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
...
+	if (space.isUnbounded) {
+		return to(origin, space);
+	}
+
+	const origin_OKLCH = to(origin, ColorSpace.get("oklch"));
+	let L = origin_OKLCH.coords[0];
+
+	// return media white or black, if lightness is out of range
+	if (L >= 1) {
+		const white = to(parse("white"), space);
```
LeaVerou, line 141: I would advise against using `parse()` here. First because it's a slower code path than creating objects (and gamut mapping needs to be *very* fast) and second because the module that resolves keywords may not even be included (Color.js needs to be tree-shakable).

I would suggest just defining an object (outside of the function, since it doesn't change) in one of the color spaces that are already imported by the dependencies of this module (lab?).

svgeesus:
> Although, I'd be likely using my implementation as both the source of truth and the system under test, as I'm not aware of a source for confirmed outputs from the algorithm.

Right, that has been a problem for a while. Safari has the CSS GMA in there (from when it was needed for conversion to HSL or HWB) but it might be an older iteration.

Testing framework is now stable, and easier to write to without all the copy-paste boilerplate that we had previously.

src/toGamut.js
jamesnw: I implemented this in https://github.com/LeaVerou/color.js/pull/344/commits/118e6f752b4de00fed1485d30399a0579fe78377, choosing Oklab. It's used in the DeltaOK calculation, so the Oklab space is explicitly required for the algorithm.

src/toGamut.js
```diff
@@ -104,3 +113,95 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
...
+		const spaceCoords = Object.values(space.coords);
+		destColor.coords = destColor.coords.map((coord, index) => {
+			const spaceCoord = spaceCoords[index];
+			if (("range" in spaceCoord)) {
+				if (coord < spaceCoord.range[0]) {
+					return spaceCoord.range[0];
+				}
+				if (coord > spaceCoord.range[1]) {
+					return spaceCoord.range[1];
+				}
```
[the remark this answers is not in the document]
LeaVerou: If a clamp function is correct (they don't 100% have the same semantics as this code), I think the best course of action is to define a `clamp(min, value, max)` in util.js, and then use it here.

src/toGamut.js
```diff
@@ -104,3 +113,103 @@ export default function toGamut (color, {method = defaults.gamut_mapping, space
 }
 
 toGamut.returns = "color";
+
+// The reference colors to be used if lightness is out of the range 0-1 in the
+// `Oklch` space. These are created in the `Oklab` space, as it is used by the
+// DeltaEOK calculation, so it is guaranteed to be imported.
```
LeaVerou, line 119: I'd use a reference to the color space then. Using string ids requires color space registration, which breaks tree-shaking. Can't remember if we do something special for basic color spaces like this one, but I don't think so,

jamesnw:
@LeaVerou I'll open a new PR. I'm also seeing an issue with the code, where `toGamut` is returning out of gamut colors in some instances- see https://deploy-preview-344--colorjs.netlify.app/apps/convert/?color=color(display-p3%201%200.78%200.86)&precision=6. In sRGB, the gamut mapped version is out of gamut.

It appears that `min` approaches `max` without finding a fit. I'm assuming it's with my implementation of the algorithm, and not the algorithm itself.

PR #347, opened 2023-11-15
Fix type def for MixOptions

### LINKED ISSUE #346: Type definitions for `mix` options should also accept `range` options (`hue`, `progression`).
The type definitions for `mix` only accept `space`, `outputSpace`, and `premultiplied`, but they should also accept the other valid `range` options: `progression` and `hue`.

jgerigmeyer:
Fixes #346


MysteryBlokHed (approved):
LGTM; I’m not sure where I got that definition for `MixOptions` in the first place

PR #349, opened 2023-11-15
Add CJS file to /fn entry and fix legacy builds

jgerigmeyer:
Includes #337 and fixes missing side-effects in legacy bundles.


LeaVerou (approved):
Thx!

PR #352, opened 2023-11-16
CSS4 toGamut fixes

jamesnw:
- Uses an imported oklab space, instead of by string.
- Adds an initial check, where if the delta between the color and the clipped version is less than the `JND`, the clipped color is returned before adjusting further.
- Always returns `clipped` instead of `current`.

```js
let current = clone(origin_OKLCH);
let clipped = clip(current);

let E = deltaEOK(clipped, current);
if (E < JND) {
    return clipped;
}
```

To compare:

```js
let pink = new Color("p3", [1, 0.77535, 0.85752]);
pink = pink.to('srgb');
pink.inGamut();
pink.toGamut();
pink.inGamut();
```

Previous output:

```js
color(display-p3 1 0.78 0.86)
rgb(104% 76.4% 86%)
false
rgb(104% 76.4% 86%)
false
```

With this change, the output is:

```js
color(display-p3 1 0.78 0.86)
rgb(104% 76.4% 86%)
false
rgb(100% 76.4% 86%)
true
```


jamesnw:
@svgeesus Do you have ideas on why I was seeing this not resolve to an in gamut color with the check that the (JND - DeltaEOK) < epsilon?

facelessuser:
I haven't looked too close at the implementation yet, but I would note one thing, using oklab `[1, 0, 0]` for white when lightness is above 100% is probably not a great idea, at least with CSS's current transform matricies. The matrix that CSS uses are 32 bit matrices which makes conversion of `[1, 0, 0]` not nearly as close to white as would be preferred. Essentially, everything after ~7 decimal places (approx. 32 precision) is garbage.

```js
> new Color('oklab', [1, 0, 0]).to('srgb')
{
    "spaceId": "srgb",
    "coords": [
        0.9999999694343976,
        1.000000008665371,
        1.000000114856359
    ],
    "alpha": 1
}
```

If using an Oklab value is desired, I would recommend you use what color.js gives you when convert white to oklab:

```js
> new Color('srgb', [1, 1, 1]).to('oklab')
{
    "spaceId": "oklab",
    "coords": [
        0.9999999934735462,
        8.095285553011422e-11,
        3.727390762708893e-8
    ],
    "alpha": 1
}
```

A 64 bit matrix can be calculated for Oklab which gives a really good value for white. This is what I do in my own library. It gives the same results as CSS up to 32 bit precision, but also has a much tighter transform for achromatic values in Oklab.

```py
>>> Color('oklab', [1, 0, 0]).convert('srgb')[:]
[1.0000000000000004, 0.9999999999999997, 0.9999999999999997, 1.0]
```

I'm not necessarily saying CSS needs to do this, but for those that are curious, this is how I went about it: https://github.com/facelessuser/coloraide/blob/main/tools/calc_oklab_matrices.py.

LeaVerou:
> If using an Oklab value is desired, I would recommend you use what color.js gives you when convert white to oklab:
> 
> ```js
> > new Color('srgb', [1, 1, 1]).to('oklab')
> {
>     "spaceId": "oklab",
>     "coords": [
>         0.9999999934735462,
>         8.095285553011422e-11,
>         3.727390762708893e-8
>     ],
>     "alpha": 1
> }
> ```

What was desired was to specify a white without adding a dependency on a color space not already imported. Nothing specific to oklab. Any ideas?

facelessuser:
You don't have to use sRGB, just the precalculated oklab value. Alternatively, you could use white as defined by XYZ D65. That is what I do as it is the one color space that is assumed to always be imported as all conversions that require chromatic adaptation pass through it.

jamesnw:
Since the intention is "white/black in the destination color space", would it make sense to add static white and black coordinates to each color space, and then referring directly to those? This would work for `toGamutCSS`, but since `toGamutCSS` converts to the source space, there still would be conversion math there.

facelessuser:
> Since the intention is "white/black in the destination color space", would it make sense to add static white and black coordinates to each color space, and then refer directly to those? This would work for toGamutCSS, but since toGamut converts to the source space, there still would be conversion math there.

This would be one way to approach it. You could definitely control and optimize the "best" white/black for each color space.

Personally, I prefer a more generic approach, at least that is what I did in my implementation. Using XYZ D65 and just applying chromatic adaptation is more than enough and gets you pretty darn close.

Color.js unfortunately seems to be using very poor chromatic adaptation matrices and are introducing errors that are unnecessary. I plan to issue a pull to fix it.

You can see here how there is a bit of error introduced due to chromatic adaptation from D65 to D50.

```js
> new Color('white').to('prophoto')
Color {
  space: <ref *1> RGBColorSpace {
    id: 'prophoto',
    name: 'ProPhoto',
    base: RGBColorSpace {
      id: 'prophoto-linear',
      name: 'Linear ProPhoto',
      base: [ColorSpace],
      aliases: undefined,
      fromBase: [Function (anonymous)],
      toBase: [Function (anonymous)],
      coords: [Object],
      white: [Array],
      formats: {},
      referred: 'display',
      path: [Array]
    },
    aliases: undefined,
    fromBase: [Function: fromBase],
    toBase: [Function: toBase],
    coords: { r: [Object], g: [Object], b: [Object] },
    white: [ 0.9642956764295677, 1, 0.8251046025104602 ],
    formats: { color: [Object] },
    referred: 'display',
    path: [ [ColorSpace], [ColorSpace], [RGBColorSpace], [Circular *1] ]
  },
  coords: [ 0.9999999886663737, 1.0000000327777285, 0.9999999636791804 ],
  alpha: 1
}
```

This is because they don't calculate the inverse of the CAT matrix. They are doubles but an inverse well below the double precision.

```js
defineCAT({
	id: "Bradford",
	// Convert an array of XYZ values in the range 0.0 - 1.0
	// to cone fundamentals
	toCone_M: [
		[  0.8951000,  0.2664000, -0.1614000 ],
		[ -0.7502000,  1.7135000,  0.0367000 ],
		[  0.0389000, -0.0685000,  1.0296000 ]
	],
	// and back
	fromCone_M: [
		[  0.9869929, -0.1470543,  0.1599627 ],
		[  0.4323053,  0.5183603,  0.0492912 ],
		[ -0.0085287,  0.0400428,  0.9684867 ]
	]
});
```

They should be using:

```js
defineCAT({
	id: "Bradford",
	// Convert an array of XYZ values in the range 0.0 - 1.0
	// to cone fundamentals
	toCone_M: [
		[  0.8951000,  0.2664000, -0.1614000 ],
		[ -0.7502000,  1.7135000,  0.0367000 ],
		[  0.0389000, -0.0685000,  1.0296000 ]
	],
	// and back
	fromCone_M: [
		[0.9869929054667123, -0.14705425642099013, 0.15996265166373122],
		[0.43230526972339445, 0.5183602715367776, 0.049291228212855594],
		[-0.008528664575177328, 0.04004282165408486, 0.96848669578755]
	]
});
```

Further, they precalculated the D65 <--> D50 CAT here, for speed, but it was precalculated with the less accurate matrices:

https://github.com/LeaVerou/color.js/blob/main/src/adapt.js#L39

They should use:

```js
	if (!env.M) {
		if (env.W1 === WHITES.D65 && env.W2 === WHITES.D50) {
			env.M = [
				[ 1.0479297925449969, 0.022946870601609708, -0.05019226628920527  ],
				[ 0.02962780877005588, 0.99043442675388, -0.017073799063418806    ],
				[ -0.009243040646204504, 0.015055191490298138, 0.7518742814281371 ]
			];
		}
		else if (env.W1 === WHITES.D50 && env.W2 === WHITES.D65) {

			env.M = [
				[ 0.947386632323667, 0.28196156725620036, -0.1708280666484637 ],
				[ -0.7357288996314816, 1.6804471734451398, 0.035992069603406264 ],
				[ 0.029218329379919382, -0.05145129980782719, 0.7733468362356041 ]
			];
		}
	}
```

All of this improves the chromatic adaptation:

```js
> new Color('white').to('prophoto')
Color {
  space: <ref *1> RGBColorSpace {
    id: 'prophoto',
    name: 'ProPhoto',
    base: RGBColorSpace {
      id: 'prophoto-linear',
      name: 'Linear ProPhoto',
      base: [ColorSpace],
      aliases: undefined,
      fromBase: [Function (anonymous)],
      toBase: [Function (anonymous)],
      coords: [Object],
      white: [Array],
      formats: {},
      referred: 'display',
      path: [Array]
    },
    aliases: undefined,
    fromBase: [Function: fromBase],
    toBase: [Function: toBase],
    coords: { r: [Object], g: [Object], b: [Object] },
    white: [ 0.9642956764295677, 1, 0.8251046025104602 ],
    formats: { color: [Object] },
    referred: 'display',
    path: [ [ColorSpace], [ColorSpace], [RGBColorSpace], [Circular *1] ]
  },
  coords: [ 1.0000000000000002, 0.9999999999999999, 1.0000000000000002 ],
  alpha: 1
}
```

svgeesus:
> Do you have ideas on why I was seeing this not resolve to an in gamut color with the check that the (JND - DeltaEOK) < epsilon?

The original gamut map code does not return, but falls through to the clip part. IIRC we did it like that to catch any final slightly-oog values.

src/toGamut.js
```diff
@@ -189,25 +189,18 @@ export function toGamutCSS (origin, { space = origin.space }) {
 		current.coords[1] = chroma;
 		if (min_inGamut && inGamut(current, space)) {
 			min = chroma;
-			continue;
 		}
 		else if (!inGamut(current, space)) {
 			const clipped = clip(current);
 			const E = deltaEOK(clipped, current);
+			// Note- this is missing several steps of the CSS Gamut Mapping
```
svgeesus, line 196: That just sounds worrying without being informative

src/toGamut.js
```diff
@@ -189,25 +189,18 @@ export function toGamutCSS (origin, { space = origin.space }) {
...
+			// Algorithm at this point.
 			if (E < JND) {
-				if ((JND - E < ε)) {
-					// match found
-					current = clipped;
-					break;
-				}
-				else {
-					min_inGamut = false;
```
svgeesus, line 204: I don't understand why those two lines are removed

jamesnw: There's some more context and examples in the description of this issue, but this resolves an issue I was seeing (although I'm not confident in this correct resolution.)

For some colors the `if ((JND - E < ε))` branch was never true, and so the algorithm continued until the `while` conditional failed, at which point the color was back out of gamut.

svgeesus: Ah, I see

svgeesus:
> Color.js unfortunately seems to be using very poor chromatic adaptation matrices and are introducing errors that are unnecessary. I plan to issue a pull to fix it.

That would be great. From memory, the forward and inverse matrices were just copied directly from [Lindbloom](http://www.brucelindbloom.com/index.html?Eqn_ChromAdapt.html).

Another issue is that Lindbloom uses the ASTM E308-01 values for white points; color.js and CSS Color 4 did so at first too, but then settled on consistently using the published four-figure values for D50 and D65 (again to hunt down some small errors from using different values in different places).

facelessuser:
Yep, I'll update all the CAT matrices with accurate inverse matrices over the weekend.

src/toGamut.js
jamesnw: I updated this in https://github.com/LeaVerou/color.js/pull/352/commits/b22d42be08af31944fad2e0ea8422217a35e941c to describe the steps I removed, and the reasoning.

facelessuser: I'm pretty sure I also just use the equivalent of `E < JND`.

facelessuser: Oh, never mind, I misunderstood. So this is my implementation. Ignore the LCh references as we subclass it for OkLCh updating the JND, space used, etc., but the logic is the same.  All of this code is tested and confirmed to be covered. I would be concerned about this over simplification of the algorithm.

https://github.com/facelessuser/coloraide/blob/main/coloraide/gamut/fit_lch_chroma.py

facelessuser: Just in case there is any doubt that the deleted steps get exercised: https://app.codecov.io/gh/facelessuser/coloraide/blob/master/coloraide%2Fgamut%2Ffit_lch_chroma.py#L77.

jamesnw:
We are continuing to explore the out of gamut issue, and this comment is documentation of our exploration so far. Essentially, what we're seeing is if the delta is always less than the JND, but the difference between JND and delta is greater than epsilon, chroma continues to project outwards until it gets within an epsilon of the original value. 

This is a rough graph of chroma progressing through the algorithm, with the red line as an approximate gamut boundary.
image

What I would expect to see is the chroma going above the boundary, and then bouncing back and forth as it gets closer.

What we observed is that the delta can be very small, and consistently less than the JND. One theory is that the magnitude of DeltaEOK is not as expected, and multiplying it by 100 (or reducing the JND to .0002) did produce better results.

We compared our algorithm to the spec, to the WIP exploration [codepen](https://codepen.io/svgeesus/pen/GRxJRKz?editors=0010), as well as Coloraide, and didn't find any significant differences that would cause this difference.

facelessuser:
> or reducing the JND to .0002

Keep in mind that the goal is not to exactly hit the cusp, but to find the closest acceptable color without reducing chroma too much. A small amount of alteration to lightness and hue is tolerable, which is why clipping is performed.

We aren't trying to hit the 0.07 chroma cusp exactly. We only want to reduce the chroma to get in close enough range to clip. 
That's why the target is above the gamut boundary.

Over-reduction can cause other side effects: https://github.com/w3c/csswg-drafts/issues/7135.

src/toGamut.js
```diff
@@ -179,21 +179,24 @@ export function toGamutCSS (origin, { space = origin.space }) {
...
-
 	let min_inGamut = true;
-	let current;
+	let current = clone(origin_OKLCH);
+    let clipped = clip(current);
+
+    let E = deltaEOK(clipped, current);
+    if (E < JND) {
+		return clipped;
+	}
```
jgerigmeyer, line 189: @jamesnw I think this is what was missing here (and also missing from the spec?). Before entering the `while` loop, both the codepen and ColorAide are _first_ checking if the delta between the color and the clipped version is less than the `JND`, and if so returning the clipped version.

@svgeesus Should this be added to the spec?

src/toGamut.js
```diff
@@ -179,21 +179,24 @@ export function toGamutCSS (origin, { space = origin.space }) {
...
 
 	while ((max - min) > ε) {
 		const chroma = (min + max) / 2;
-		current = clone(origin_OKLCH);
 		current.coords[1] = chroma;
 		if (min_inGamut && inGamut(current, space)) {
 			min = chroma;
-			continue;
 		}
 		else if (!inGamut(current, space)) {
```
jgerigmeyer, line 197: I notice that both the codepen and ColorAide are using a plain `else` here and not including the `if (!inGamut(current, space))` -- is that correct (and the spec should be adjusted), or do we want to leave this here?

jgerigmeyer (commented):
@jamesnw I think I found the difference? This is now returning the same value for `new Color("p3", [1, 0.77535, 0.85752]).toGamut({ space: 'srgb' })` as ColorAide and the codepen.

src/toGamut.js
facelessuser: I understand the confusion as if you are reading the spec is says:

> otherwise, if inGamut(current) is false carry out these steps:

It should be noted that this is not a false statement, you just don't need to explicit check it because the state is implied.

Once the lower limit is out of gamut (`min_inGamut` is `false`), you don't need to check if you are in gamut anymore as it is implied you are out of gamut. If `min_inGamut` is `true`, then you have to do the "in gamut" check, and then if it yields `false` you know you are not in gamut.  In short `if (min_inGamut && inGamut(current, space))` is fails, it is implied that you are out of gamut, so all you need is an `else`.

jgerigmeyer: @facelessuser Thanks -- that makes sense. I removed the unnecessary check.

src/toGamut.js
```diff
@@ -118,8 +118,8 @@ toGamut.returns = "color";
 // `Oklch` space. These are created in the `Oklab` space, as it is used by the
 // DeltaEOK calculation, so it is guaranteed to be imported.
 const COLORS = {
-	WHITE: { space: "oklab", coords: [1, 0, 0] },
-	BLACK: { space: "oklab", coords: [0, 0, 0] }
+	WHITE: { space: oklab, coords: [1, 0, 0] },
```
facelessuser, line 121: Just to make sure this isn't forgotten, a resolution of how color.js wishes to ensure a good white/black should be decided as the current Oklab matrix is not exact enough to give a decent white and will introduce a notable error.

Whether each space reports the optimal white/black, a precalculated perfect white in Oklab is used, a generic method using XYZ D65 white and chromatically adapting the value to the target color space's white point is used, a recalculating Oklab with 64 bit precision eliminate the inaccuracies near achromatic values is used, or some other method is used, a solution should be considered to reduce errors here.

facelessuser: It should be noted that I do have a pull request to fix the issues related to chromatic adaptation here: https://github.com/LeaVerou/color.js/pull/354. I imagine this could be rebased on that to test for accuracy.

facelessuser:
Aside from my comments about the inaccuracy of the current white being used, without testing the actual results, I think the logic looks good to me.

jamesnw:
It looks like @jgerigmeyer's change fixes almost but not all cases, and I found a very small number of cases where the algorithm does not return an in gamut color.

<details><summary>Notes on the failures</summary>
I wrote a script to create colors in the `display-p3` space, with each channel from 0-1 in increments of .01, and then convert that to the sRGB gamut. Out of those 101^3 checks, [58 colors](https://codepen.io/jamessw/pen/QWYQrWB?editors=1000) still ended up out of gamut. 

Given a quick look, the failures seem to be fairly representative of the areas that are in gamut for p3 and out of gamut for
sRGB.

I reran the test with smaller increments, with 201^3 checks. This had [483](https://codepen.io/jamessw/full/OJdQGvE) ending out of gamut). This was in 0.006% of cases, which was proportional to the 101^3 checks.

It also appears that the algorithm is failing for small ranges- given `color(display-p3 0.99 0.79 0.07)`, it fails for `g` values of 0.7897-0.7901, but works outside that range.

Given this, it seems to be some mathematical/precision/conversion issue, and not something inherent about a color.
</details>

I spot checked the failures against Coloraide, and it appears we would get the same outputs if we would apply a final clip.

<details><summary>Comparison</summary>
For instance, given `color(display-p3 0.99 0.79 0.07)`:

Coloraide:
```
>>> r=Color('display-p3', [0.99, 0.79, 0.07]).convert('srgb')
// color(srgb 1.0281 0.78007 -0.26893 / 1)
>>> r.fit(method='oklch-chroma')
// color(srgb 1 0.78207 0 / 1)
```

Color.js
```
let r = new Color("p3", [0.99, 0.79, 0.07]).to('srgb');
// rgb(102.81% 78.007% -26.89%)
r.toGamut();
// rgb(102.25% 78.207% -23.7%)
```

Applying a clip to that would result in `rgb(100% 78.207% 0%)`, identical to the
Coloraide result (except in format).
</details>

However, I'm not certain that a clip of `current` (the return value) is the most correct fix here. I stepped through the code, comparing `color(display-p3 0.99 0.7901 0.07)` which fails with `color(display-p3 0.99 0.7902 0.07)` which succeeds. Notably, in the failure case, `max - min` reaches epsilon _before_ the delta reaches the JND. This means that for the failure case, `current` is never reassigned a new value that is in gamut. 

Instead of clipping `current`, I think we should use the value of `clipped`, as that was the value that was approaching a match before the `while` loop was cut off. 

I pushed that change in https://github.com/LeaVerou/color.js/pull/352/commits/1375f3eab8546a2dedcd281e760dd63db3310812. 

With that change, no colors in the 101^3 checks are returning out of gamut.

@svgeesus Should this check be added to the spec, in addition to the [check](https://github.com/LeaVerou/color.js/pull/352#discussion_r1397954684) @jgerigmeyer added before the `while` loop? 

I think the spec change on this would be replacing step 15 with-

```
15. If `inGamut(current)` is false, then set `current` to `clipped`.
16. Return current as the gamut mapped color.
```

Alternatively, we could say that 50 out of a million colors being naively clipped is a tradeoff for an extra inGamut check on all colors, in which case I think the spec change would be changing line 15 to-

```
15. Return clip(current) as the gamut mapped color
```

facelessuser:
@jamesnw The clipped version is what you want to return. You are reducing chroma of the current color is until it is just under the JND distance to the clipped version of the color. When this is satisfied, you can use the clipped version as the color difference won't be noticeable.

facelessuser:
I would actually propose simplifying the code and just always return clipped. No extra gamut check is needed at the end. If the color is already in gamut, the clipped would equal the current color. There is no need to set current to clipped, just return clipped.

facelessuser:
It may not be clear in my original proposal, but we were modifying `color` in place while reducing `mapcolor`, the (gamut mapping color space). `color` was always clipped when updated.

facelessuser:
I've created an issue over at CSSWG to address an error in the GM algorithm: https://github.com/w3c/csswg-drafts/issues/9618

jamesnw:
> @svgeesus, @jamesnw I think this statement is a mistake in the CSS spec. My original proposal always returned the clipped color, never the the LCh color (`mapcolor`) we were reducing.
> 
> > return current as the gamut mapped color
> 
> [w3c/csswg-drafts#7135 (comment)](https://github.com/w3c/csswg-drafts/issues/7135#issuecomment-1066121795)

Thanks for the context, @facelessuser.  I think that's exactly where this differed from Coloraide. In the 50-in-a-milliion cases where max - min reaches epsilon before the delta reaches the JND, we were returning `current` instead of clipped. I pushed a change to reflect your proposed change to the spec.

jamesnw:
Given the proposed change in spec from @facelessuser, this code now differs from the spec in 2 ways-

1. Returns `clipped` instead of current, as @facelessuser proposes
2. Contains an optimization from @jgerigmeyer for when the initial delta is less than the JND

@svgeesus, I'd appreciate your input on each if the spec or the code should be updated.

For #2, the optimization avoids roughly 10 loops through the `while` loop, but produces slightly difference output. Given `color(display-p3 1 0.78 0.86)` to `srgb`-

With the early return: 
`rgb(100% 76.394093% 86.046255%)`

Without the early return:
`rgb(100% 76.403826% 86.045681%)`

Both are correct, in that the Delta is within a JND (and are [indistinguishable](https://www.oddcontrast.com/#srgb__rgb(100~_76.404~_86.046~)__rgb(100~_76.394~_86.046~))). 

Is #2 worth adding to the spec?

facelessuser:
I think \#2 makes sense. There is no reason to waste time iterating if the color is already within the JND range. I think clipping and returning makes sense. It is also what I proposed in the original version, I think I overlooked that it was removed. I think in the final revision it was simplified and removed, but I think there is value in adding, especially for performance. Even in @svgeesus's original JS interpretation of it, the same was done: https://github.com/w3c/csswg-drafts/issues/7135#issuecomment-1175087912. The only error was that `current` was being returned.

src/toGamut.js
svgeesus: Yes I think this should be added

svgeesus:
The spec should be updated, yes

src/toGamut.js
```diff
@@ -193,9 +193,14 @@ export function toGamutCSS (origin, { space = origin.space }) {
...
-			// Note- this is missing several steps of the CSS Gamut Mapping
-			// Algorithm at this point.
 			if (E < JND) {
+				// Note- this implementation does not conform to the CSS Spec.
+				// This removes the check that (JND - E < ε) and its else
+				// statement. For some colors, JND - E was never less than ε, so
+				// the algorithm continued projecting chroma higher, until the
+				// while statement ended. At that point, the chroma was returned
+				// to `max`, which is the original chroma, so the color was
+				// still out of gamut.
```
svgeesus: This comment can now be simplified because

 - https://github.com/w3c/csswg-drafts/commit/06748226bb73a7fee401ba0d00c4e4655ec17555

src/toGamut.js
svgeesus: That PR is now merged, so the CA is now good. I also put the same values in the CSS Color 4 sample code.

src/toGamut.js
svgeesus: Now added: https://github.com/w3c/csswg-drafts/commit/06748226bb73a7fee401ba0d00c4e4655ec17555

src/toGamut.js
facelessuser: I've opened another PR with new Oklab matrices for working in 64 bit. https://github.com/LeaVerou/color.js/pull/357

facelessuser: Due to my own carelessness, I had updated the inverse D50 <-> D65 matrix incorrectly. The last fix is here: https://github.com/LeaVerou/color.js/pull/360.

Once #357 and #360 are approved, all the fixes for proper GMA should be in place and then this issue can be resolved.

src/toGamut.js
jamesnw: This comment is removed now.

jamesnw:
@svgeesus I'm a bit confused by the change in https://github.com/w3c/csswg-drafts/commit/06748226bb73a7fee401ba0d00c4e4655ec17555- I think you may have added the early return in the wrong place, and then removed some needed steps that had become unreachable.

The change we were looking to make in # 2 above was to insert after line 10:

- set |current| to |origin_Oklch|
- set |clipped| to clip(|current|)
- set |E| to delta(|clipped|, |current|)
- if |E| < |JND| return |clipped| as the gamut mapped color

This is identical to the logic within the `while`, but crucially is checked _before_ the chroma is adjusted.

src/toGamut.js
facelessuser: Since #357 and #360 are now on main, it no longer matters that `oklab(1 0 0)` is used for white as it should be plenty accurate now.

jamesnw: Great, I've merged main in, so we can test both the toGamut logic and white conversions [here](https://deploy-preview-352--colorjs.netlify.app/).

jamesnw:
I updated the gamut tests in the new suite (I appreciate the update, btw!). I added some tests of the new algorithm, which I  verified these results against Coloraide's results. You can see them [here](https://deploy-preview-352--colorjs.netlify.app/test/?test=gamut).

 I also reorganized it, as the new suite doesn't appear to support tests that are 3 layers deep. There were also some failing tests which I fixed by updating the Expected to match the Actual, which I think was appropriate in this case. This also allowed me to increase the epsilon magnitude for a tighter check.

LeaVerou:
> I updated the gamut tests in the new suite (I appreciate the update, btw!).

Would love to hear more about your experience with the testsuite since I'm piloting it on color.js but plan to release it more widely soon.

jamesnw:
FYI, I opened a [CSSWG Issue](https://github.com/w3c/csswg-drafts/issues/9651) with some suggested from questions that arose in this PR.

svgeesus (approved):
This all looks good now!

svgeesus:
It looks as if a rebase will be needed before this can be merged

jamesnw:
> It looks as if a rebase will be needed before this can be merged

Done! Thanks!

src/toGamut.js
```diff
@@ -163,40 +165,36 @@ export function toGamutCSS (origin, { space = origin.space }) {
...
+    let E = deltaEOK(clipped, current);
+    if (E < JND) {
+		return clipped;
+	}
 
 	while ((max - min) > ε) {
 		const chroma = (min + max) / 2;
-		current = clone(origin_OKLCH);
 		current.coords[1] = chroma;
 		if (min_inGamut && inGamut(current, space)) {
```
facelessuser, line 190: One thing that I missed, this should really be using an `epsilon` of zero. The earlier shortcut out should be based on an `epsilon` of zero as well (line 160).

jamesnw: I updated this in https://github.com/color-js/color.js/pull/352/commits/907c8f3544cd6436287530fd8ce515b2fc2216b2

The [Gamut tests](https://deploy-preview-352--colorjs.netlify.app/test/?test=gamut) are passing on this branch- are there additional cases I should add for checking what you were looking at in #378 ?

LeaVerou:
Is this ready to merge?

jamesnw:
> Is this ready to merge?

Yes, I don’t think there are any outstanding questions.

LeaVerou (approved):

LeaVerou:
Merged! Thanks everyone!

PR #351, opened 2023-11-16
Add Prettier and recommended ESLint rules.

jgerigmeyer:
I know these are subjective and I'm very open to changes, but I'm proposing a clearer and more opinionated (and automated) set of style/formatting rules. I see two main advantages:

1. Contributors don't need to expend mental energy trying to determine stylistic expectations.
2. Codebase uses a consistent style, reducing inconsistencies between different files.

(I'd also _personally_ prefer a switch from tabs to 2-spaces, but I'm not looking to start any fights 😅)

@LeaVerou @svgeesus Thoughts?

PR #367, opened 2023-11-28
Add more ESLint recommended rules, and enforce them in CI.

jgerigmeyer:
A replacement of #351. Based off of #372.

A few notes/questions:

1. What is the value of "warnings" vs "errors", and what should cause CI to fail? Right now the default recommended rules are set to "error" (except where overridden), but most of the custom rules are set to "warn". CI will fail if there are any errors _or_ warnings. I'd probably recommend moving all of them to "error" and not using "warn" at all -- it sort of defeats the purpose if warnings pile up unaddressed, and can easily mask newly introduced unintentional warnings.
2. There are a handful of places where rules are enabled globally (e.g. [no-cond-assign](https://eslint.org/docs/latest/rules/no-cond-assign), [no-sparse-arrays](https://eslint.org/docs/latest/rules/no-sparse-arrays), [no-redeclare](https://eslint.org/docs/latest/rules/no-redeclare), [no-unused-vars](https://eslint.org/docs/latest/rules/no-unused-vars)) because they're helpful in revealing common typos, but then are overridden in specific instances where the usage is intentional. There are other rules (e.g. [no-loss-of-precision](https://eslint.org/docs/latest/rules/no-loss-of-precision)) that we might want to disable globally?

Very open to feedback and/or rules changes!


.editorconfig
```diff
@@ -2,7 +2,6 @@ root = true
 
 [*]
 indent_style = tab
-indent_size = 4
```
LeaVerou, line 5: Should we convert to `tab_width` instead?

.eslintrc.json
```diff
@@ -11,27 +12,71 @@
...
+	"globals": { "Proxy": "readonly" },
+	"extends": [
+		"eslint:recommended",
+		"plugin:@typescript-eslint/recommended",
+		"plugin:@stylistic/disable-legacy"
+	],
+	"plugins": ["@typescript-eslint", "@stylistic"],
 	"rules": {
-		"semi": 1,
+		"@stylistic/arrow-spacing": 1,
```
LeaVerou, line 23: Could we use [a config format that allows comments](https://eslint.org/docs/latest/use/configure/configuration-files#configuration-file-formats) so we can document what each of these is? (at least those not entirely obvious from the name, which is most of them)

That will also help me review the rest of this :)

.eslintrc.json
```diff
@@ -11,27 +12,71 @@
...
+			"files": [
+				"apps/**/*",
+				"assets/**/*",
+				"get/**/*",
+				"notebook/**/*",
+				"scripts/**/*",
+				"tests/**/*"
+			],
+			"rules": {
+				"@typescript-eslint/no-unused-vars": 0,
```
LeaVerou, line 65: Not sure I want to enforce this, sometimes having unused vars can be useful for documentation purposes or to iterate.

.editorconfig
jgerigmeyer: That would be fine, although is there a reason to enforce this globally at all? It seems valuable to allow users to define this locally for their own IDEs, and I wasn't finding it obvious how to tell VS Code to ignore this setting. From the [.editorconfig docs](https://editorconfig.org/#file-format-details):

> Also, when `indent_style` is set to tab, it may be desirable to leave `indent_size` unspecified so readers may view the file using their preferred indentation width.

.eslintrc.json
```diff
@@ -11,27 +12,71 @@
 		"browser": true,
 		"node": true
 	},
+	"globals": { "Proxy": "readonly" },
+	"extends": [
+		"eslint:recommended",
```
LeaVerou, line 17: I’m a little uncomfortable with leaving coding style up to mystery bags of rules that we have little control over, and don't fully realize what we're enforcing. Could we list the rules explicitly?

.eslintrc.json
```diff
@@ -11,27 +12,71 @@
 		"browser": true,
 		"node": true
 	},
+	"globals": { "Proxy": "readonly" },
```
LeaVerou, line 15: Pretty sure we use more globals than that?

notebook/repl.js
```diff
@@ -49,17 +49,17 @@ let editObserver = new Mavo.Observer(container, "mv-mode", () => {
 
 (async () => {
 
-await Mavo.ready;
+	await Mavo.ready;
```
LeaVerou, line 52: I don't think adding indentation here helps readability, do you?.

src/color.js
```diff
@@ -78,7 +78,9 @@ export default class Color {
 		for (let id in this.space.coords) {
 			Object.defineProperty(this, id, {
 				get: () => this.get(id),
-				set: value => this.set(id, value)
+				set: value => {
+					this.set(id, value);
+				}
```
LeaVerou, line 83: I don't think this change improves readability, do you?

src/equals.js
```diff
@@ -5,6 +5,6 @@ export default function equals (color1, color2) {
 	color2 = getColor(color2);
 
 	return color1.space === color2.space
-		   && color1.alpha === color2.alpha
-		   && color1.coords.every((c, i) => c === color2.coords[i]);
+		  && color1.alpha === color2.alpha
+		  && color1.coords.every((c, i) => c === color2.coords[i]);
```
LeaVerou, line 9: Oof, neither of these is actually good formatting 😅

src/hooks.js
```diff
@@ -5,8 +5,8 @@ export class Hooks {
 	add (name, callback, first) {
 		if (typeof arguments[0] != "string") {
 			// Multiple hooks
-			for (var name in arguments[0]) {
-				this.add(name, arguments[0][name], arguments[1]);
+			for (var _name in arguments[0]) {
```
LeaVerou, line 8: This looks like it needs a proper refactoring

src/serialize.js
```diff
@@ -25,8 +24,8 @@ export default function serialize (color, {
 
 	let formatId = format;
 	format = color.space.getFormat(format)
-		   ?? color.space.getFormat("default")
-		   ?? ColorSpace.DEFAULT_FORMAT;
+		    ?? color.space.getFormat("default")
+		    ?? ColorSpace.DEFAULT_FORMAT;
```
LeaVerou, line 28: Same as the && one, neither of this is good formatting. I'd instead right align the `??` with the `=`.

src/space-accessors.js
```diff
@@ -20,11 +20,6 @@ hooks.add("colorspace-init-end", space => {
 });
 
 function addSpaceAccessors (id, space) {
-	// Coordinates can be looked up by both id and name
-	let coordIds = Object.keys(space.coords);
-	let coordNames = Object.values(space.coords).map(c => c.name);
-
-
```
LeaVerou, line 27: Wow, these were entirely unused?!

LeaVerou (changes requested):
Hey, thank you so much for your hard work and perseverance on this! 

1. Regarding errors vs warnings: I tend to specify all linting as warnings, to distinguish from actual errors.

2. As a general philosophy, I am against `// eslint-disable-*` directives. I think the code should be the source of truth, written in the optimal way, and tools like ESLint should just work with it without leaving traces in the codebase. It should not require more migration to swap ESLint for another linter than migrating the config file itself. If we need to override a rule (rather than refactoring the code) I take it as a strong indication that maybe the rules we have are too strict and we've ended up fighting against them rather than them helping us, so we should not be enforcing the rule in the first place, or we should be enforcing a more lax version of the rule that would allow that kind of formatting.

3. It would be good to commit some of the uncontroversial code changes this PR introduces (e.g. adding semicolons, space between func name and opening paren, things like that) in a separate PR so that this would not be as huge to review. In fact, I'd suggest this workflow: 
	1. Take all code changes in here. Remove those that require eslint overrides, and those that clearly do not improve readability.
	2. Send a PR with *just* the code changes, without the linting infrastructure. 
	3. Once we iterate and merge that, *then* we can focus on any changes the linting infrastructure introduces without being lost in the hundreds of missing semicolons added and the like.
	How does that sound?

.editorconfig
LeaVerou: Oh wow, I had assumed it was entirely advisory, otherwise what's even the point of using tabs?!

src/space-accessors.js
jgerigmeyer: They were made obsolete by https://github.com/color-js/color.js/commit/cd0b54311fd6f5c80545aaa45022c99192f1b246, but not removed.

.eslintrc.json
jgerigmeyer: I'm torn on this. On one hand I understand wanting the rules to be explicit and opt-in; on the other hand ESLint is very stable, and maintains a long list of recommended rules that don't change often and would be duplicated here. And `typescript-eslint` also _disables_ some of the default ESLint rules that it replaces, so it would be a very long list. I'm open to copying them here if you prefer that (and we could try narrowing them down). The lists this is currently using are:

- https://github.com/eslint/eslint/blob/v8.54.0/packages/js/src/configs/eslint-recommended.js
- https://github.com/typescript-eslint/typescript-eslint/blob/v6.13.1/packages/eslint-plugin/src/configs/eslint-recommended.ts
- https://github.com/typescript-eslint/typescript-eslint/blob/v6.13.1/packages/eslint-plugin/src/configs/recommended.ts
- https://github.com/typescript-eslint/typescript-eslint/blob/v6.13.1/packages/eslint-plugin/src/configs/stylistic.ts

jgerigmeyer:
>    1. Take all code changes in here. Remove those that require eslint overrides, and those that clearly do not improve readability.
>    2. Send a PR with _just_ the code changes, without the linting infrastructure.
>    3. Once we iterate and merge that, _then_ we can focus on any changes the linting infrastructure introduces without being lost in the hundreds of missing semicolons added and the like.
>       How does that sound?

@LeaVerou I pulled out the code changes into #372, and rebased this PR off of that one.

.eslintrc.json
jgerigmeyer: Continued in #373

PR #369, opened 2023-11-28
Add gamutSpace to ColorSpace

lloydk:
Some colorspaces such as hsluv, okhsl and okhsv need to do gamut checking in a color space that is not their own space or their base space. For example hsluv needs to do gamut checking in the srgb color space.

A gamutSpace can now be specified when constructing a ColorSpace and a gamutSpace property has beed added to ColorSpace.

This PR is backwards compatible with existing color spaces and uses the "least bad" option of specifying "self" for the gamutSpace of polar color spaces that need to do gamut checking in their own color space (hpluv) as discussed [here](https://github.com/color-js/color.js/pull/339#issuecomment-1772971055).

In the original `inGamut` method the `coords` were converted to the base space by directly calling the `toBase` method. This version of inGamut calls the `to` method on `gamutSpace` which will result in `NaN` values being converted to 0 before the gamut check. Not sure if this matters.

I've also added an optimization to override inGamut for unbounded spaces. If you don't think it's appropriate I can remove it.


svgeesus:
> Some colorspaces such as hsluv, okhsl and okhsv need to do gamut checking in a color space that is not their own space or their base space

HSL, HSV and HWB already do this (they check in sRGB)

src/space.js
```diff
@@ -53,6 +53,27 @@ export default class ColorSpace {
...
+			this.gamutSpace = this.base;
+		}
+
+		// Optimize inGamut for unbounded spaces
+		if (this.gamutSpace.isUnbounded){
+			this.inGamut = (coords, options) => {
+				return true;
+			};
+		}
+
```
LeaVerou, line 76: Minor nit: Setting `this.gamutSpace` to `this` and then undoing it later makes the logic hard to follow. I'd restructure a bit to avoid this, possibly via having two branches for the two main cases like:

```js
if (options.gamutSpace) {
	// Gamut space explicitly specified
	...
}
else {
	// Gamut space not explicitly specified, fall back to sensible defaults
	...
}
```

Note that if the gamut space is unbounded, `this.gamutSpace` is effectively still `this`, we just override `inGamut` as a performance optimization.

LeaVerou (changes requested):
Thanks for this! 
So sorry for the delay, somehow these PRs completely slipped through the cracks, I don't recall ever seeing them before!

Overall looks good, see comments for a few minor issues.

src/space.js
lloydk: I've changed the code to look more like what you suggested.

types/src/space.d.ts
```diff
@@ -37,6 +37,7 @@ export interface Options {
 	cssId?: string | undefined;
 	referred?: string | undefined;
 	formats?: Record<string, Format> | undefined;
+	gamutSpace?: string | ColorSpace | null | undefined;
```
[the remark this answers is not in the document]
lloydk: `string` can be `"self"` or a spaceId. I added `"self"` to the list of types just to make it a bit more clear. Ideally we'd have a type alias for spaceId. Something like:

```type SpaceId = string;```

And then we could replace `string` with `SpaceId`.

File: src/space.js [hunk 3]
```diff
@@ -53,6 +53,23 @@ export default class ColorSpace {
 			this.formats.color.id = this.id;
 		}
 
+		// Gamut space
+
+		if (options.gamutSpace) {
+			this.gamutSpace = options.gamutSpace === "self" ? this : ColorSpace.get(options.gamutSpace);
```
LeaVerou, line 59:
```suggestion
			// Gamut space explicitly specified
			this.gamutSpace = options.gamutSpace === "self" ? this : ColorSpace.get(options.gamutSpace);
```

src/space.js
```diff
@@ -53,6 +53,23 @@ export default class ColorSpace {
...
+		}
+		else {
+			// Do not check gamut through polar coordinates
+			this.gamutSpace = this.isPolar ? this.base : this;
```
LeaVerou, line 63:
```suggestion
			// No gamut space specified, calculate a sensible default
			if (this.isPolar) {
				// Do not check gamut through polar coordinates
				this.gamutSpace = this.base;
			}
			else {
				 this.gamutSpace =  this;
			}
```

LeaVerou (approved):
LGTM, see suggestions for clarity

lloydk:
> LGTM, see suggestions for clarity

I've made the suggested changes.

PR #370, opened 2023-11-28
Allow coord grammar to be specified for the color syntax

lloydk:
Some non CSS level 4 color spaces need coordinate grammars that go beyond what is currently allowed for the color() syntax.

For example the luv color space requires <percentage>[-1, 1] to be specified for its u and v coordinates. Other color spaces such as hsluv, hpluv, okhsl, okhsv have coordinates that are angles.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -5,6 +5,56 @@ import ColorSpace from "./space.js";
 const noneTypes = new Set(["<number>", "<percentage>", "<angle>"]);
 
 /**
+ * Validates the coordinates of a color against a format's coord grammar and
+ * maps the coordinates to the range or refRange of the coordinates.
+ * @param {ColorSpace} space - Colorspace the coords are in
+ * @param {object} format - the format object to validate against
+ * @param {string} name - the name of the color function. e.g. "oklab" or "color"
+ * @returns {object[]} - an array of type metadata for each coordinate
+ */
+function coerceCoords (space, format, name, coords) {
+	let types = Object.entries(space.coords).map(([id, coordMeta], i) => {
+		let coordGrammar = format.coordGrammar[i];
+		let arg = coords[i];
+		let providedType = arg?.type;
+
+		// Find grammar alternative that matches the provided type
+		// Non-strict equals is intentional because we are comparing w/ string objects
+		let type;
+		if (arg.none) {
+			type = coordGrammar.find(c => noneTypes.has(c));
+		}
+		else {
+			type = coordGrammar.find(c => c == providedType);
+		}
+
+		// Check that each coord conforms to its grammar
+		if (!type) {
+			// Type does not exist in the grammar, throw
+			let coordName = coordMeta.name || id;
+			throw new TypeError(`${providedType ?? arg.raw} not allowed for ${coordName} in ${name}()`);
+		}
+
+		let fromRange = type.range;
+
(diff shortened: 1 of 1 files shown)


svgeesus:
>

Good, point, in CSS Color 4 the parameters of the `color()` function, after the initial color space identifier, are

[ [&lt;number>](https://drafts.csswg.org/css-values-4/#number-value) [|](https://drafts.csswg.org/css-values-4/#comb-one) [&lt;percentage>](https://drafts.csswg.org/css-values-4/#percentage-value) [|](https://drafts.csswg.org/css-values-4/#comb-one) none ][{3}](https://drafts.csswg.org/css-values-4/#mult-num)

so specifying polar spaces inside `color()` is an unresolved problem.

LeaVerou:
> Good, point, in CSS Color 4 the parameters of the `color()` function, after the initial color space identifier, are
> 
> [ [<number>](https://drafts.csswg.org/css-values-4/#number-value) [|](https://drafts.csswg.org/css-values-4/#comb-one) [<percentage>](https://drafts.csswg.org/css-values-4/#percentage-value) [|](https://drafts.csswg.org/css-values-4/#comb-one) none ][{3}](https://drafts.csswg.org/css-values-4/#mult-num)
> 
> so specifying polar spaces inside `color()` is an unresolved problem.

Not something to resolve in this PR, but I think this should be fixed in the spec. We should allow `<angle> | <number> | <percentage>` and IACVT if needed (or define how each type resolves to the other two)

LeaVerou:
Hi there, apologies again for the delay, apparently I had seen this PR before, but it somehow completely fell off my radar. 

I'm …not quite sure what this is doing. Abstracting that block of code to a separate function seems like a useful refactor, but beyond that I'm not quite sure what new use cases it's enabling. A coord grammar can already be specified for `color()`, by explicitly adding it as a format, and a range can already be specified in a coord grammar (e.g. `<number>[1, -1]`). Pretty sure I'm just not getting it, so if you could explain it a bit more, that would be useful! Sorry for the trouble @lloydk, especially after you've already waited so long!!

lloydk:
> Hi there, apologies again for the delay, apparently I had seen this PR before, but it somehow completely fell off my radar.
> 
> I'm …not quite sure what this is doing. Abstracting that block of code to a separate function seems like a useful refactor, but beyond that I'm not quite sure what new use cases it's enabling. A coord grammar can already be specified for `color()`, by explicitly adding it as a format, and a range can already be specified in a coord grammar (e.g. `<number>[1, -1]`). Pretty sure I'm just not getting it, so if you could explain it a bit more, that would be useful! Sorry for the trouble @lloydk, especially after you've already waited so long!!

Maybe I'm missing something but if you look at the current code the logic for handling the coord grammar is only executed for color space specific functions (oklab, oklch, lab, etc.).

lloydk:
I think you might have missed the fact that the extracted code is now called in two places instead of one place in the original code.

You can currently specify a coord grammar for the color function but it isn't used and that's what this PR fixes.

LeaVerou:
One thing — I did not check if the code that was moved to the function was the same, did you make any changes there?

lloydk:
> One thing — I did not check if the code that was moved to the function was the same, did you make any changes there?

I didn't make any changes to the code in the extracted function

LeaVerou:
K, merged, thanks!

PR #374, opened 2023-12-13
Fix node 16 module resolution for CJS

wongsean:
Fix module resolution for CJS when TypeScript is set with `node16` or `nodenext`.

Currently, it reports an error when imported, `The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'. Consider writing a dynamic 'import("colorjs.io")' call instead.`, despite it having a package.json `exports`.

It seems that the require and import conditions should have separate type files for TypeScript to be able to resolve CJS modules correctly ([reference](https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports)), 

```json
"exports": {
  ".": {
    "import": {
      "default": "./dist/color.js",
      "types": "./types/index.d.ts"
    },
    "require": {
      "default": "./dist/color.cjs",
      "types": "./types/index.d.cts"
    }
  }
}
```
instead of the current one
```json
"exports": {
  ".": {
    "types": "./types/index.d.ts",
    "import": "./dist/color.js",
    "require": "./dist/color.cjs"
  }
}
```

Note also that the `.d.cts` file extension, after testing, specifying `require.types` to the `.d.ts` file doesn't work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,18 +7,29 @@
 		"src/",
 		"types/dist/",
 		"types/src/",
-		"types/index.d.ts"
+		"types/index.d.ts",
+		"types/index.d.cts"
 	],
 	"exports": {
 		".": {
-			"types": "./types/index.d.ts",
-			"import": "./dist/color.js",
-			"require": "./dist/color.cjs"
+			"import": {
+				"default": "./dist/color.js",
+				"types": "./types/index.d.ts"
+			},
+			"require": {
+				"default": "./dist/color.cjs",
+				"types": "./types/index.d.cts"
+			}
 		},
 		"./fn": {
-			"types": "./types/src/index-fn.d.ts",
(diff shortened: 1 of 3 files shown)


wongsean:
Hi @jgerigmeyer, here is a reproduce [repo](https://github.com/wongsean/reproduce-colorjs-cjs-resolution) and error could be found in the [failed ci](https://github.com/wongsean/reproduce-colorjs-cjs-resolution/actions/runs/7384445188/job/20087400586).

jgerigmeyer:
@wongsean Thanks! The sample repo was very helpful -- I've pushed this and a few other minor changes in #383.

wongsean:
Glad it helps😄

PR #378, opened 2023-12-16
Ensure clipped is always returned and initialized

### LINKED ISSUE #376: Near-blacks in `oklab` crash the convert utility
Compare [`oklab(0.002 0 0)`](https://colorjs.io/apps/convert/?color=oklab(0.002%200%200)&precision=6) which has no output at all, with [`oklab(0.00000 0 0)`](https://colorjs.io/apps/convert/?color=oklab(0.00000%200%200)&precision=6) which does

image

facelessuser:
Ensures clipped value is always returned and initialized, removes some unnecessary code, and tests with zero tolerance to ensure value is fully in gamut.

Fixes #376

### DIFF
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -155,7 +155,7 @@ export function toGamutCSS (origin, { space = origin.space }) {
 		return to(black, space);
 	}
 
-	if (inGamut(origin_OKLCH, space)) {
+	if (inGamut(origin_OKLCH, space, { epsilon: 0 })) {
 		return to(origin_OKLCH, space);
 	}
 
@@ -180,23 +180,23 @@ export function toGamutCSS (origin, { space = origin.space }) {
 	let max = origin_OKLCH.coords[1];
 
 	let min_inGamut = true;
+	let clipped = clip(clone(origin_OKLCH));
 	let current;
 
 	while ((max - min) > ε) {
 		const chroma = (min + max) / 2;
 		current = clone(origin_OKLCH);
 		current.coords[1] = chroma;
-		if (min_inGamut && inGamut(current, space)) {
+		if (min_inGamut && inGamut(current, space, { epsilon: 0 })) {
 			min = chroma;
 			continue;
 		}
-		else if (!inGamut(current, space)) {
-			const clipped = clip(current);
+		else {
+			clipped = clip(current);
 			const E = deltaEOK(clipped, current);
 			if (E < JND) {
 				if ((JND - E < ε)) {
 					// match found
-					current = clipped;
 					break;
 				}
 				else {
@@ -206,9 +206,8 @@ export function toGamutCSS (origin, { space = origin.space }) {
 			}
 			else {
 				max = chroma;
-				continue;
 			}
 		}
 	}
-	return to(current, space);
+	return to(clipped, space);
 }


facelessuser:
Gamut tests are broken on the release and here...I'm looking into that to make sure nothing else is broken.

facelessuser:
Tests are busted on the official site and here. The changes themselves ensure we don't return and undefined color and that the final color is always clipped. We remove implied logic. Lastly, we ensure our gamut checks are done with zero epsilon for better gamut mapping. They should all be safe changes, but they may have to be evaluated manually.

facelessuser:
Okay, since the HTML tests are broken, I did a little sanity check here to demonstrate that no existing things are broken and the targetted issue is fixed. I also explain in a little more detail why I made the other changes.

```js
mport Color from "./dist/color.js"

const colors = [
    'color(display-p3 1 0 0)',
    'color(display-p3 0 1 0)',
    'color(display-p3 0 0 1)',
    'color(display-p3 1 1 0)',
    'color(display-p3 0 1 1)',
    'color(display-p3 1 0 1)'
]

colors.map(c => {
    console.log(new Color(c).toGamut('srgb').toString({precision: 6, inGamut: true}));
})
// Failing case
console.log(new Color('oklab(0.00002 0 0)').toGamut('acescc').toString({precision: 6, inGamut: true}));
```

This is what color.js did before the fix:

```console
➜  color.js git:(master) ✗ node example.js
color(display-p3 0.91779 0.21072 0.15424)
color(display-p3 0.45147 0.97123 0.33189)
color(display-p3 0.00008 0.00074 0.99984)
color(display-p3 0.99673 0.99892 0.33034)
color(display-p3 0.45667 0.98176 0.97814)
color(display-p3 0.91957 0.26299 0.95178)
file:///Users/facelessuser/Code/github/color.js/dist/color.js:1010
		throw new TypeError("Empty color reference");
		      ^

TypeError: Empty color reference
    at getColor (file:///Users/facelessuser/Code/github/color.js/dist/color.js:1010:9)
    at to (file:///Users/facelessuser/Code/github/color.js/dist/color.js:1769:10)
    at toGamutCSS (file:///Users/facelessuser/Code/github/color.js/dist/color.js:1758:9)
    at toGamut (file:///Users/facelessuser/Code/github/color.js/dist/color.js:1585:19)
    at func (file:///Users/facelessuser/Code/github/color.js/dist/color.js:4339:14)
    at Color.<computed> [as toGamut] (file:///Users/facelessuser/Code/github/color.js/dist/color.js:4366:12)
    at file:///Users/facelessuser/Code/github/color.js/example.js:15:45
    at ModuleJob.run (node:internal/modules/esm/module_job:192:25)
    at async DefaultModuleLoader.import (node:internal/modules/esm/loader:228:24)
    at async loadESM (node:internal/process/esm_loader:40:7)

Node.js v20.5.1
```

This is after the fix.

```console
color(display-p3 0.91779 0.21072 0.15424)
color(display-p3 0.45147 0.97123 0.33189)
color(display-p3 0.00014 0.00074 0.95959)
color(display-p3 0.99673 0.99892 0.33034)
color(display-p3 0.45667 0.98176 0.97814)
color(display-p3 0.91957 0.26299 0.95178)
oklab(0.39061% 0 0)
```

1. Results are the same, but now the aces case doesn't fail due to an uninitialized variable.
2. There is no longer a risk of an infinite loop by removing the second `else if` condition. This could only occur in an extreme case with unexpected color space geometry.
3. A better tolerance check for gamut mapping is used.

LeaVerou:
> Okay, since the HTML tests are broken, I

They are? Are you visiting the old testsuite or the new one?

facelessuser:
@LeaVerou Let's ignore this review for a second and I'll instead reference what I think are the official tests, the legacy and new (so I think). There are two and the gamut tests are either failing or not running: https://colorjs.io/tests/ and https://colorjs.io/test/.

The first seems to be broken and all failing:

Screenshot 2023-12-16 at 3 58 49 PM

The second seems broken in a completely different way:

Screenshot 2023-12-16 at 3 59 07 PM

facelessuser:
I had not realized this issue had not merged: https://github.com/color-js/color.js/pull/352. That review should fix most of these issues, so I will close this as a duplicate.

svgeesus:
> Let's ignore this review for a second and I'll instead reference what I think are the official tests, the legacy and new (so I think). There are two and the gamut tests are either failing or not running: https://colorjs.io/tests/ and https://colorjs.io/test/.

> The first seems to be broken and all failing:

The link in the footer went to the old (`/tests`) so I changed it today to link to the new (`/test`)

PR #379, opened 2023-12-17
Add CAM16 (JMh)

facelessuser:
This is the first step in porting over support for the HCT color space. This adds the CAM16 model along with the output in JMh for verification.

### DIFF
--- a/src/spaces/cam16.js
+++ b/src/spaces/cam16.js
@@ -1,0 +1,359 @@
+import ColorSpace from "../space.js";
+import {multiplyMatrices, interpolate, copySign, spow, zdiv, bisectLeft} from "../util.js";
+import {constrain} from "../angles.js";
+import xyz_d65 from "./xyz-d65.js";
+import {WHITES} from "../adapt.js";
+
+const white = WHITES.D65;
+const adaptedCoef = 0.42;
+const adaptedCoefInv = 1 / adaptedCoef;
+const tau = 2 * Math.PI
+
+const cat16 = [
+	[  0.401288,  0.650173, -0.051461 ],
+	[ -0.250268,  1.204414,  0.045854 ],
+	[ -0.002079,  0.048952,  0.953127 ]
+];
+
+const cat16Inv = [
+	[1.8620678550872327, -1.0112546305316843, 0.14918677544445175],
+	[0.38752654323613717, 0.6214474419314753, -0.008973985167612518],
+	[-0.015841498849333856, -0.03412293802851557, 1.0499644368778496]
+]
+
+const m1 = [
+	[460.0, 451.0, 288.0],
+	[460.0, -891.0, -261.0],
+	[460.0, -220.0, -6300.0]
+]
+
+const surroundMap = {
+	dark: [0.8, 0.525, 0.8],
+	dim: [0.9, 0.59, 0.9],
+	average: [1, 0.69, 1]
+}
+
+const hueQuadMap = {
+	// Red, Yellow, Green, Blue, Red
+	h: [20.14, 90.00, 164.25, 237.53, 380.14],
+	e: [0.8, 0.7, 1.0, 1.2, 0.8],
+	H: [0.0, 100.0, 200.0, 300.0, 400.0]
+}
+
+const rad2deg = 180 / Math.PI;
+const deg2rad = Math.PI / 180;
+
+export function adapt (coords, fl) {
+	const temp = coords.map(c => {
+		const x = spow(fl * Math.abs(c) * 0.01, adaptedCoef);
+		return 400 * copySign(x, c) / (x + 27.13);
+	});
+	return temp;
+}
+
+export function unadapt (adapted, fl) {
+	const constant = 100 / fl * (27.13 ** adaptedCoefInv);
+	return adapted.map(c => {
+		const cabs = Math.abs(c);
+		return copySign(constant * spow(cabs / (400 - cabs), adaptedCoefInv), c);
+	});
+}
+
+export function hueQuadrature (h) {
+	let hp = constrain(h)
+	if (hp <= hueQuadMap.h[0]) {
+		hp += 360;
+	}
+
+	const i = bisectLeft(hueQuadMap.h, hp) - 1;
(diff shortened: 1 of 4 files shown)


facelessuser:
One thing to note, undefined values in CSS and this library are assumed to be zero when a real number is required. CAM16 has a non-linear achromatic response that often has achromatic values a non-zero chroma/colorfulness/saturation values. What this means is that chroma and hue both matter when converting achromatic values. Knowing what those values are allows for better round trip conversions.

There currently isn't a mechanism in Color.js to handle such things. Ideally, in the future, two things should be considered:

1. A way to determine when a color is achromatic and assign undefined hues appropriately. We could do this fairly easy in the  forward conversion to CAM16 JMh. We do not do this currently because of issue number 2.
2. A way to resolve undefined values for achromatics when zero is not sufficient. The hue matters very much in the conversion from CAM16 JMh back to XYZ. How should color.js convert an undefined CAM16 JMh hue?

I do not think these questions need to be resolved now to implement the color space, but it should be noted that when undefined values in CAM16 are used, and in the future HCT, that it may not yield the achromatic values that are desired.

facelessuser:
I will also note the "colorfulness" range is based on P3. This can be changed to whatever is desired.

facelessuser:
Adding a little more info on validation.

A number of references used:

- https://observablehq.com/@jrus/cam16: reference implementation
- https://arxiv.org/pdf/1802.06067.pdf: Nico Schlömer
- https://onlinelibrary.wiley.com/doi/pdf/10.1002/col.22324: hue quadrature

Results were compared against Colour Science and my implementation in ColorAide as well: https://github.com/colour-science/colour

```py
>>> import numpy as np
>>> import colour
>>> from coloraide.everything import ColorAll as Color
>>> xyz = np.array(Color('red').convert('xyz-d65').coords()) * 100
>>> colour.XYZ_to_CAM16(xyz, **colour.appearance.CAM_KWARGS_CIECAM02_sRGB)
CAM_Specification_CAM16(J=46.025701408152251, C=112.3966873794304, h=27.39325656758691, s=98.395239471843041, Q=83.926266178253229, M=81.254248158497788, H=9.2041940637581128, HC=None)
>>> Color('red').convert('cam16-jmh')
color(--cam16-jmh 46.026 81.254 27.393 / 1)
```

Color.js

```js
> const { default: Color } = require("colorjs.io");
undefined
> new Color('red').to('cam16-jmh').coords
[ 46.025701408152244, 81.25424815849786, 27.39325656758689 ]
```

svgeesus:
I have read the Nico Schlömer paper before, and it troubles me that the ISO document did not even consider things like division by zero when producing their specification. Effectively, this paper is the real specification.

svgeesus:
The full text of the Li _et al_ paper originally defining CAM16 [is available on ResearchGate](https://www.researchgate.net/publication/318152296_Comprehensive_color_solutions_CAM16_CAT16_and_CAM16-UCS), avoiding the Wiley paywall

facelessuser:
> The full text of the Li _et al_ paper originally defining CAM16 [is available on ResearchGate](https://www.researchgate.net/publication/318152296_Comprehensive_color_solutions_CAM16_CAT16_and_CAM16-UCS), avoiding the Wiley paywall

Oh, nice. I searched around for a non-paywall version for quite a bit. I can update the reference link.

svgeesus (approved):
Overall this looks great, thanks for doing it.

svgeesus:
There should probably be some dim and dark tests, but that should not hold up merging this.

src/spaces/cam16.js
```diff
@@ -0,0 +1,364 @@
...
+
+	return {J: J, C: C, h: h, s: s, Q: Q, M: M, H: H};
+}
+
+
+// Provided as a way to directly evaluate the CAM16 model
+// https://observablehq.com/@jrus/cam16: reference implementation
+// https://arxiv.org/pdf/1802.06067.pdf: Nico Schlömer
+// https://onlinelibrary.wiley.com/doi/pdf/10.1002/col.22324: hue quadrature
+// https://doi.org/10.1002/col.22131: pay walled
```
svgeesus, line 328:
```suggestion
// https://www.researchgate.net/publication/318152296_Comprehensive_color_solutions_CAM16_CAT16_and_CAM16-UCS
```

svgeesus:
> 2\. A way to resolve undefined values for achromatics when zero is not sufficient. The hue matters very much in the conversion from CAM16 JMh back to XYZ. How should color.js convert an undefined CAM16 JMh hue?
> 
> 
> I do not think these questions need to be resolved now to implement the color space, but it should be noted that when undefined values in CAM16 are used, and in the future HCT, that it may not yield the achromatic values that are desired.

Perhaps we can just say that hue is never powerless in CAM16-JMh and treat `none` as an error if it occurs?

facelessuser:
> Perhaps we can just say that hue is never powerless in CAM16-JMh and treat `none` as an error if it occurs?

Certainly a possibility. The two easiest approaches are:

1. Do as you say, and treat `none` as an error.
2. Treat `none` as zero and the user gets what they get.

---

I went for a more complicated approach in ColorAide. I provide some hook to allow undefined channels to resolve differently than zero and provide a resolution there.

I'm not suggesting that Color.js does this, but I calculated a "good enough" achromatic response and use a spline to tell me for a given lightness, what the ideal chroma and hue would be for that lightness. I kept it simple in that lightness was the decider (though a negative chroma will cause the ideal hue to be rotated). It does well enough, but obviously, this approximation for undefined hues and chroma comes at the cost of some accuracy to provide convenience. You also have to either generate the data on the fly at some point in memory or increase the size of your library to ship the pre-calculated data.

But it does provide a sane response.  Obviously, a more invasive change with costs, so you can see why I did not suggest it by default. I imagine Color.js will likely want to go with one of the easier options. It was probably more work than I was ready for here as well 🙃.

Screenshot 2023-12-18 at 11 11 36 AM

LeaVerou:
Feel free to ignore this is not relevant, just a drive-by comment. I know that CAM16 involves a bunch of variables about the viewing conditions, and presumably we had to make some assumptions here. Would it make sense to add a factory method that produces `ColorSpace` objects for *different* viewing conditions?

PR #384, opened 2024-01-09
Fix toPrecision (was off by one for fractional inputs)

### LINKED ISSUE #385: toPrecision is broken
toPrecision outputs the wrong values for some inputs. The function considers the leading "0" in a decimal number to be a significant digit, which it is not.

For example: The `toPrecision` function is wrong for decimal inputs like `0.222`:

Ex:
```js
toPrecision(0.222, 2) // 0.2
```
Instead, it should be
```js
toPrecision(0.222, 2) // 0.22
```

efergus:
The `toPrecision` function was wrong for fractional inputs like `0.222` with small values for precision, for example:

Ex:
```js
toPrecision(0.222, 2) // 0.2
```
Instead, it should be
```js
toPrecision(0.222, 2) // 0.22
```

closes #385

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -29,7 +29,7 @@ export function serializeNumber (n, {precision, unit }) {
 		return "none";
 	}
 
-	return toPrecision(n, precision) + (unit ??  "");
+	return toPrecision(n, precision) + (unit ?? "");
 }
 
 /**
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
LGTM

Issue #45, opened 2020-06-24
ICC profile support


svgeesus:
Not that this should be added for first release, or indeed soon, but just so I can close some browser tabs :)

- [JavaScript source code to parse ICC profile, version 4 ](https://formats.kaitai.io/icc_4/javascript.html)
- [Javascript notes for the above](http://doc.kaitai.io/lang_javascript.html)

- [node-transicc](https://www.npmjs.com/package/transicc) which is a node wrapper around
- [Little-CMS](https://github.com/mm2/Little-CMS)

svgeesus:
To be clear, this is a future "nice to have". For additive colorspaces, our code does the same thing. The main benefits would be:

1. Cross checking the JS code against the ICC result, for existing supported spaces
2. Add support for non-additive colorspaces (CMYK, CMTKOGV) which require multidimensional interpolation of nearby measured color values.

LeaVerou:
Indeed. 

For anyone interested in embarking on this:
- I’m envisioning a `ColorSpace.load()` async factory method, that would read a path/URL to an ICC profile and return a `ColorSpace` object that behaves like normal `ColorSpace` objects. We will likely need to introduce a few constructor options to make this possible.
- ICC profile parsing is quite involved, but there are also many libraries on it (mostly in other languages) that can be used as a base.

necessarylion:
I found this library https://github.com/glennwilton/jsColorEngine and result is quite similar compare to Little-CMS and written in 100% Javascript.

Achder:
No pressure or anything, I'm just interested in potential progress.
Is there anything to report yet? ICC profiles would be amazing!

PR #386, opened 2024-01-18
[spaces/cmyk]: added a new color space

blunteshwar:
Added support for a new color-space 'CYMK'

### DIFF
--- a/src/spaces/cmyk.js
+++ b/src/spaces/cmyk.js
@@ -32,7 +32,7 @@ function toCmyk (rgb) {
 	let y = (1 - b - k) / (1 - k);
 	let a = [c, m, y, k];
 	let ans = a.map(el => el * 100);
-	return a; //[c, m, y, k].map(el => Math.round(el * 100));
+	return ans; //[c, m, y, k].map(el => Math.round(el * 100));
 }
 
 export default new ColorSpace({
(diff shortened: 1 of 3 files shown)


Crissov:
On second thought, I’m not sure anymore if this was meant as a joke.  This is an awkward representation of sRGB which isn’t really useful for anything, except as a last resort fallback. https://www.w3.org/TR/css-color-5/#cmyk-rgb Labeling this as “CMYK” would be misleading and harmful.

blunteshwar:
So, what do you suggest should I do?

LeaVerou:
Hello there, thank you so much for contributing!

I'd love to add CMYK color spaces, but CMYK is not a simple transformation of sRGB. Implementing it this way includes colors that would not be representable by any CMYK space *actually* in use. See this blog post I wrote almost 15 years ago (!): https://lea.verou.me/blog/2009/03/100-cyan-in-cmyk-is-not-rgb0255255/

I respect your efforts and hate closing PRs without merging, but this would basically require a complete rewrite to be mergeable.

Adding CMYK in Color.js properly would be quite involved.
1. First step would be to [support ICC profiles](https://github.com/color-js/color.js/issues/45#issuecomment-1902237326)
2. Then we can add some common CMYK spaces (e.g. US Coated SWOP, FOGRA, etc.) to be available as predefined spaces and decide which one `"cmyk"` should resolve to.

@svgeesus may have more advice or thoughts here.

Sorry I didn't have better news, and I hope this does not discourage you from future contributions! 🙏🏼

PR #392, opened 2024-02-03
Fix incorrect parsing of Jzazbz percentage values

lloydk:
The `az` and `bz` coordinates for the Jzazbz color space have negative reference range values so percentage values are being mapped incorrectly.

Before this PR
before

After this PR
after

The parsing tests haven't been converted to the new format so I added a test case to the old parsing tests.


LeaVerou (approved):
Thanks!

Any chance you could open an issue for the parsing tests? Can't believe I missed that when converting!

PR #400, opened 2024-02-04
Ensure alpha is a number in Color constructor

lloydk:
Fixes two failing constructor tests.

### DIFF
--- a/src/color.js
+++ b/src/color.js
@@ -65,7 +65,7 @@ export default class Color {
 		this.coords = coords ? coords.slice() : [0, 0, 0];
 
 		// Clamp alpha to [0, 1]
-		this.alpha = 0 <= alpha && alpha <= 1? alpha : 1;
+		this.alpha = 0 <= alpha && alpha <= 1 ? alpha : 1;
 
 		// Convert "NaN" to NaN
 		for (let i = 0; i < this.coords.length; i++) {
--- a/src/color.js
+++ b/src/color.js
@@ -64,13 +64,8 @@ export default class Color {
 
 		this.coords = coords ? coords.slice() : [0, 0, 0];
 
-		// Ensure alpha is a number
-		if (!Number.isFinite(alpha)) {
-			alpha = 1;
-		}
-
 		// Clamp alpha to [0, 1]
-		this.alpha = alpha > 1 ? 1 : (alpha < 0 ? 0 : alpha);
+		this.alpha = 0 <= alpha && alpha <= 1? alpha : 1;
 
 		// Convert "NaN" to NaN
 		for (let i = 0; i < this.coords.length; i++) {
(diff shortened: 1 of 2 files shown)


src/color.js
```diff
@@ -64,6 +64,11 @@ export default class Color {
 
 		this.coords = coords ? coords.slice() : [0, 0, 0];
 
+		// Ensure alpha is a number
+		if (!Number.isFinite(alpha)) {
+			alpha = 1;
+		}
+
 		// Clamp alpha to [0, 1]
 		this.alpha = alpha > 1 ? 1 : (alpha < 0 ? 0 : alpha);
```
LeaVerou, line 73: Would something like this work?
```suggestion
		// Clamp alpha to [0, 1]
		this.alpha = alpha < 1 && Number.isFinite(alpha) ? (alpha < 0 ? 0 : alpha) : 1;
```

Or even:
```suggestion
		// Clamp alpha to [0, 1]
		this.alpha = Math.min(Math.max(0, alpha), 1) : 1;
```

Or (this should work even for non-numerical values):

Or even:
```suggestion
		// Clamp alpha to [0, 1]
		this.alpha = 0 <= alpha && alpha <= 1? alpha : 1;
```

lloydk: I preferred the last suggestion

jgerigmeyer (approved):

jgerigmeyer (commented):
Before, `alpha` could be passed in as `NaN` and it would be retained as such -- this now coerces it to `1`. Is that intentional?

src/color.js
LeaVerou: I just realized this had an issue: it does not actually clamp, if you pass an alpha of -1, it will also become 1 (which is both unintuitive and a breaking change).

How about:

```
// Clamp alpha to [0, 1]
this.alpha = Number.isNaN(alpha)? 1 : Math.max(0, Math.min(alpha, 1));
```

lloydk:
> Before, `alpha` could be passed in as `NaN` and it would be retained as such -- this now coerces it to `1`. Is that intentional?

You're right, I think we need to allow `NaN` to be specified. 

The `getColor` function checks `alpha === undefined` so maybe that's what we should do in the `Color` constructor. 
https://github.com/color-js/color.js/blob/44f3741696f9b8c4f0d59b8448bd2055191ed35e/src/getColor.js#L27-L29


Should `getColor` clamp alpha like the `Color` constructor?

LeaVerou:
Oh good point. We *do* plan to eventually use `null` for `none`, rather than `NaN`, since that's a separate keyword, but for now we should coerce to NaN.

lloydk:
I've added some additional constructor tests for NaN/none/clamping and alpha is set 1 one only if it's > 1 or undefined.

jgerigmeyer (approved):

LeaVerou (approved):

PR #404, opened 2024-02-06
Add color spaces for HSLuv and HPLuv

lloydk:
Add color spaces for HSLuv and HPLuv.

The conversion routines where adapted from the [typescript source code](https://github.com/hsluv/hsluv-javascript/blob/14b49e6cf9a9137916096b8487a5372626b57ba4/src/hsluv.ts) in the HSLuv repository. 

## Tests
The HSLuv repository has [test data](https://raw.githubusercontent.com/hsluv/hsluv/master/snapshots/snapshot-rev4.json) in json format.

I added a script to generate tests suites for hpluv and hsluv. `npm run test:generate-hsluv` will create `test/hpluv.js` and `test/hsluv.js`. The tests take forever to load/run on my ancient Windows PC when run via `npm run test`. Running the individual test files is less bad. I've added the files to .gitignore so they aren't committed accidently.

image

### In Gamut Tests
I added a test suite for `inGamut()` and the various methods of specifying the `gamutSpace` for a color space. HSLuv has a `gamutSpace` of `srgb` and HPLuv has a `gamutSpace` of `"self"`.


scripts/generate-hsluv-tests.js
```diff
@@ -0,0 +1,115 @@
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+import https from "node:https";
+
+const testDataUrl = "https://raw.githubusercontent.com/hsluv/hsluv/ebe5f6fbe7082e04468a325b4c62f783e35987ad/snapshots/snapshot-rev4.json";
```
LeaVerou, line 6: Cool idea to import the tests from that repo!
However, if we're gonna pin it to a specific commit, we may as well just copy the file over and `import` it in the test. There is however value in loading the latest version of the file. Not sure what route you'd rather go with.

lloydk: The [test data hasn't changed in 8 years](https://github.com/hsluv/hsluv/tree/master/snapshots) so I guess there's no value in pinning to a specific commit. I don't think the file will ever change and if changes are required a `snapshot-rev5.json` will be created.

My preference right now would be to do the least amount of work. So either keep it as is or change the `testDataUrl` to not reference a specific commit.

lloydk: Walked the dog and changed my mind.

- I'll create a `test/hsluv` directory that contains a copy of the json test data and test files for hpluv and hsluv
- the test files will load the json test data and create the appropriate tests
- the `test:generate-hsluv` script will be deleted
- `npm run test:hsluv` will run the hsluv tests

lloydk:
I've revamped the tests:
- no generation step
- all tests and test data are stored in `test/hsluv`
- `npm run test:hsluv` to run the tests

jgerigmeyer (approved):

LeaVerou (approved):
I would move `test/hsluv/hpluv.js` and `test/hsluv/hsluv.js` to be directly into `test` (and they can still import the files they need from `hsluv/`. Other than that, LGTM.

lloydk:
> I would move `test/hsluv/hpluv.js` and `test/hsluv/hsluv.js` to be directly into `test` (and they can still import the files they need from `hsluv/`. Other than that, LGTM.

Done

LeaVerou:
Merged, thanks! 🎉

Issue #388, opened 2024-01-29
Precision lets scientific notation leak for small values


SebastianStehle:
Hi,

I have very little experience with color spaces. I found your solution for a simple theme designer that i build for my app using DaisyUI.

They store the colors in oklch and when the user pickers a rgb color, using a color picker I have to make the conversion. It is actually super easy:

```                        
const color = new Color(input.value);
const converted = color.to('oklch');

root.style.setProperty(style, `${converted.coords[0]} ${converted.coords[1]} ${converted.coords[2]}`);
```

It works fine, except for black and white:

Except black and white (`#00000` and `#ffffff)

* Black returns the same: `1.0000000000000002 4.996003610813204e-16 NaN`
* White returns `1.0000000000000002 4.996003610813204e-16 NaN`

facelessuser:
> Black returns the same: 1.0000000000000002 4.996003610813204e-16 NaN
> White returns `1.0000000000000002 4.996003610813204e-16 NaN`

Are you saying that a conversion from sRGB `#000000` is giving you the same value that `#ffffff` gives you? I'm a bit confused by your statement.

This is what I get for black and white:

```js
> new Color('#000000').to('oklch').coords
[ 0, 0, NaN ]
> new Color('#ffffff').to('oklch').coords
[ 1.0000000000000002, 4.996003610813204e-16, NaN ]
```

This matches what I get in their live playground: https://colorjs.io/notebook/

SebastianStehle:
Yes, you are right. I also get [0, 0, NaN] now. Not sure what I did.

SebastianStehle:
What is the status?

facelessuser:
Your issue is not reproducible, and then it also seems you are now getting what you should, so there is nothing to look into. The issue has been closed.

SebastianStehle:
I am confused. My test with black was obviously wrong, but both colors do not return the expected result. 

Black:
* Result: `[1,5e-16,NaN]`
* Expected `[100, 0, 0]` (https://oklch.com/#100,0,0,100)

White
* Result `[0, 0, NaN]`
* Expected `[0, 0, 0]`

My problem was not that both return the same color, but that the conversion from black and white return invalid colors.

facelessuser:
Unfortunately, it was not clear from your filed issue what your issue was, so let me explain.

1. The site you linked is giving you rounded values after conversion, if they were to give you full precision values, you would see that chroma for white is very near zero but not quite zero. This is due to the perils of floating point math.

[...]

PR #413, opened 2024-02-09
Make color space optional in `getAll()` + `colorSpace.equals()` improvements

LeaVerou:
This PR does two things:
1. Supports string ids in `colorSpace.equals()` (and adds JSDoc for the method)
2. Makes the color space optional in `getAll()`.
Not only does this make `getAll()` more consistent with the rest of the API, it also sets the stage for option 3 here: https://github.com/color-js/color.js/issues/388#issuecomment-1936678737 . As a side effect, it also implements a performance improvement when the color passed *happens* to be in the same color space as the one specified as it now goes through the same code path.

### DIFF
--- a/src/getAll.js
+++ b/src/getAll.js
@@ -1,12 +1,17 @@
 import ColorSpace from "./space.js";
 
 /**
- * Get the coordinates of a color in another color space
- *
- * @param {string | ColorSpace} space
- * @returns {number[]}
+ * Get the coordinates of a color in any color space
+ * @param {Color} color
+ * @param {string | ColorSpace} [space = color.space] The color space to convert to. Defaults to the color's current space
+ * @returns {number[]} The color coordinates in the given color space
  */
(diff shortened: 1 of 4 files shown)


MysteryBlokHed (commented):
The JavaScript changes LGTM, but the types should be changed to reflect that the `space` parameter is now optional:

<https://github.com/color-js/color.js/blob/2f264bfc49c172272f0881d4b9811b481e7832ca/types/src/getAll.d.ts#L4-L7>

```diff
 export default function getAll (
 	color: Color | ColorObject,
-	space: string | ColorSpace
+	space?: string | ColorSpace
 ): [number, number, number];

```

jgerigmeyer (approved):

LeaVerou:
@MysteryBlokHed Yup, I was hoping that you or @jgerigmeyer would be kind enough to do that 😁

PR #416, opened 2024-02-10
Fix precision

### LINKED ISSUE #388: Precision lets scientific notation leak for small values

facelessuser:
- Handle small precision properly
- Zero precision will return value rounded to integer

Fixes #388

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -54,13 +54,16 @@ export function skipNone (n) {
  * @param {number} precision - Number of significant digits
  */
 export function toPrecision (n, precision) {
-	n = +n;
 	if (n === 0) {
 		return 0;
 	}
-	precision = +precision;
-	const multiplier = Math.pow(10, precision - Math.floor(Math.log10(Math.abs(n))) - 1);
-	return Math.round(n * multiplier) / multiplier;
+	let integer = ~~n;
+	let digits = 0;
+	if (integer && precision) {
+		digits = ~~Math.log10(Math.abs(integer)) + 1;
+	}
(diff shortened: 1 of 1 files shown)


facelessuser:
Precision zero will give integer rounding and small values are rounded properly now.

```js
> Color.util.toPrecision(23.3, 0)
23
> Color.util.toPrecision(0.222, 2)
0.22
> Color.util.toPrecision(1e-8, 3)
0
> Color.util.toPrecision(1e-8, 8)
1e-8
```

LeaVerou:
Merged, thanks! But I don't think it fixes #388 entirely, since it was not only about serialization (sure, people can use `toPrecision` to map the coords, but it seems wrong to require jumping through hoops for common things).

facelessuser:
Understood, my intention was only to fix the bug. I viewed the other part as an enhancement, one that an official direction was not yet decided upon.

PR #418, opened 2024-02-10
Allow angle values for HCT Hue coordinate

lloydk:
Allow angle values for HCT Hue coordinate.

### DIFF
--- a/src/spaces/hct.js
+++ b/src/spaces/hct.js
@@ -150,7 +150,8 @@ export default new ColorSpace({
 	},
 	formats: {
 		color: {
-			id: "--hct"
+			id: "--hct",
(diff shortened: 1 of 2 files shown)


LeaVerou (approved):

LeaVerou:
Gah, approved too soon. Could you please add that parse test to the actual testsuite? The old testsuite is only there for historical reasons.

lloydk:
> Gah, approved too soon. Could you please add that parse test to the actual testsuite? The old testsuite is only there for historical reasons.

If the parse tests were ported I would have added them to that suite. 

If you want to hold off on merging until there's a parse test suite in the new format I'm fine with that.

lloydk:
I guess you already merged it, wasn't paying attention.

LeaVerou:
Ah ok, I guess it makes sense then.

PR #419, opened 2024-02-10
[types] Add missing overload for `set`

MysteryBlokHed:
The `set` function can also take an object of coordinates to modify, as described in the README: <https://github.com/color-js/color.js#manipulating-colors>

This PR changes the types to reflect this.

The reason that the definition on the `Color` class had to be changed is because of how TypeScript handles overloaded functions in other types (i.e. `ToColorPrototype`). Without this, only the last signature of the `set` function would be present on `Color`.


MysteryBlokHed:
(Force pushed because of accidental formatting changes)

jgerigmeyer (approved):

PR #420, opened 2024-02-10
Add HCT gamut mapping and HCT color distance

facelessuser:
- Add an HCT color distancing to keep from converting out of HCT which can be slower. Since HCT is based on CAM16, convert the C and h components to CAM16 UCS a and b. Perform euclidean distance on the resulting Lab lightness (tone) and CAM16 a and b.
- Add support for HCT gamut mapping. Add two keywords to configure HCT gamut mapping. "hct" giving a normal gamut mapping assuming a JND of 2 and "hct-tonal" which will perform a tighter gamut mapping with a JND of close to zero and will force clamping of black and white based on tone value. This will give results that quite close to Material's tonal palettes.
- toGamut has been extended to allow configuring the delta E method used, the JND, and enabling white and black SDR clamp.
- Epsilon is now generically calculated to be relative to the JND.
- The epsilon of min/max bisect threshold is still based on the JND epsilon, but now it is arbitrarily clamped to 1e-6 to prevent calculating a very small epsilon that could cause infinite loops.
- Gamut tests have been extended to allow for converting to target space after gamut mapping (if needed). This is used to demonstrate tonal maps which require us (for best results) to stay in HCT.

### DIFF
--- a/src/deltaE/deltaEHCT.js
+++ b/src/deltaE/deltaEHCT.js
@@ -5,10 +5,12 @@ const rad2deg = 180 / Math.PI;
 const deg2rad = Math.PI / 180;
 const ucsCoeff = [1.00, 0.007, 0.0228];
 
-
+/**
+* Convert HCT chroma and hue (CAM16 JMh colorfulness and hue) using UCS logic for a and b.
+* @param {number[]} coords - HCT coordinates.
+* @return {number[]}
+*/
 function convertUcsAb (coords) {
-	// Convert HCT chroma and hue (CAM16 JMh colorfulness and hue) using UCS logic for a and b.
-
 	// We want the distance between the actual color.
 	// If chroma is negative, it will throw off our calculations.
 	// Normally, converting back to the base and forward will correct it.
@@ -30,9 +32,13 @@ function convertUcsAb (coords) {
 }
 
 
+/**
+* Color distance using HCT.
+* @param {Color} color - Color to compare.
+* @param {Color} sample - Color to compare.
+* @return {number[]}
+*/
 export default function (color, sample) {
-	// Delta E HCT color distance formula.
-
 	let [ t1, a1, b1 ] = convertUcsAb(hct.from(color));
 	let [ t2, a2, b2 ] = convertUcsAb(hct.from(sample));
 
--- a/src/deltaE/deltaEHCT.js
+++ b/src/deltaE/deltaEHCT.js
@@ -1,0 +1,42 @@
+import hct from "../spaces/hct.js";
+import {viewingConditions} from "../spaces/hct.js";
+
+const rad2deg = 180 / Math.PI;
+const deg2rad = Math.PI / 180;
+const ucsCoeff = [1.00, 0.007, 0.0228];
+
+
+function convertUcsAb (coords) {
+	// Convert HCT chroma and hue (CAM16 JMh colorfulness and hue) using UCS logic for a and b.
+
+	// We want the distance between the actual color.
+	// If chroma is negative, it will throw off our calculations.
+	// Normally, converting back to the base and forward will correct it.
+	// If we have a negative chroma after this, then we have a color that
+	// cannot resolve to positive chroma.
+	if (coords[1] < 0) {
+		coords = hct.fromBase(hct.toBase(coords));
+	}
+
+	// Only in extreme cases (usually outside the visible spectrum)
+	// can the input value for log become negative.
+	// Avoid domain error by forcing a zero result via "max" if necessary.
+	const M = Math.log(Math.max(1 + ucsCoeff[2] * coords[1] * viewingConditions.flRoot, 1.0)) / ucsCoeff[2];
+	const hrad = coords[0] * deg2rad;
+	const a = M * Math.cos(hrad);
+	const b = M * Math.sin(hrad);
+
+	return [coords[2], a, b];
+}
+
+
+export default function (color, sample) {
+	// Delta E HCT color distance formula.
+
+	let [ t1, a1, b1 ] = convertUcsAb(hct.from(color));
+	let [ t2, a2, b2 ] = convertUcsAb(hct.from(sample));
+
+	// Use simple euclidean distance with a and b using UCS conversion
+	// and LCh lightness (HCT tone).
+	return Math.sqrt((t1 - t2) ** 2 + (a1 - a2) ** 2 + (b1 - b2) ** 2);
+}
--- a/src/deltaE/index.js
+++ b/src/deltaE/index.js
@@ -6,7 +6,15 @@ import deltaEITP from "./deltaEITP.js";
 import deltaEOK from "./deltaEOK.js";
 import deltaEHCT from "./deltaEHCT.js";
 
-export { deltaE76, deltaECMC, deltaE2000, deltaEJz, deltaEITP, deltaEOK, deltaEHCT };
+export {
+	deltaE76,
+	deltaECMC,
+	deltaE2000,
+	deltaEJz,
+	deltaEITP,
+	deltaEOK,
+	deltaEHCT
+};
 
 export default {
 	deltaE76,
--- a/src/deltaE/index.js
+++ b/src/deltaE/index.js
@@ -4,8 +4,9 @@ import deltaE2000 from "./deltaE2000.js";
 import deltaEJz from "./deltaEJz.js";
 import deltaEITP from "./deltaEITP.js";
 import deltaEOK from "./deltaEOK.js";
+import deltaEHCT from "./deltaEHCT.js";
 
-export { deltaE76, deltaECMC, deltaE2000, deltaEJz, deltaEITP, deltaEOK };
+export { deltaE76, deltaECMC, deltaE2000, deltaEJz, deltaEITP, deltaEOK, deltaEHCT };
 
 export default {
 	deltaE76,
@@ -14,4 +15,5 @@ export default {
 	deltaEJz,
 	deltaEITP,
 	deltaEOK,
+	deltaEHCT
 };
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -7,9 +7,35 @@ import inGamut from "./inGamut.js";
 import to from "./to.js";
 import get from "./get.js";
 import oklab from "./spaces/oklab.js";
+import xyzd65 from "./spaces/xyz-d65.js";
 import set from "./set.js";
 import clone from "./clone.js";
 import getColor from "./getColor.js";
+import deltaEMethods from "./deltaE/index.js";
+import {WHITES} from "./adapt.js";
+
+
+function calcEpsilon (jnd) {
+	// Calculate the epsilon to 2 degrees smaller than the specified JND.
+
+	const order = (!jnd) ? 0 : Math.floor(Math.log10(Math.abs(jnd)));
+	return parseFloat(`1e${order - 2}`);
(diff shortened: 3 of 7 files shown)


facelessuser:
This completes my plan for enabling HCT tonal maps.

facelessuser:
Forgot to mention that we now source `mappedColor` from `color` not `spaceColor` now. If we want good tonal palettes, we need to keep the original HCT color (when provided). Converting to the gamut space and back to the mapped color introduces error and can actually break tonal palettes if the color is too far out of gamut, beyond what HCT (and the underlying CAM16 algorithm) can handle.

src/toGamut.js
```diff
@@ -24,7 +50,16 @@ import getColor from "./getColor.js";
...
 
-export default function toGamut (color, { method = defaults.gamut_mapping, space = color.space } = {}) {
+export default function toGamut (
+	color,
+	{
+		method = defaults.gamut_mapping,
+		space = color.space,
+		deltaEMethod = "",
+		jnd = 2,
+		blackWhiteClamp = {}
```
jgerigmeyer, line 60: Could you update the JSDoc comments above to reflect these new options?

jgerigmeyer (commented):
`deltaEHCT` needs a new type definition, and the type def for `toGamut` should be adjusted to include the new options.

src/toGamut.js
facelessuser: I can update comments. I'm not that familiar with type stuff in JS, so I may need very straight forward guidance for anything non-trivial, but I'll do what I can.

src/deltaE/index.js
```diff
@@ -4,8 +4,9 @@ import deltaE2000 from "./deltaE2000.js";
 import deltaEJz from "./deltaEJz.js";
 import deltaEITP from "./deltaEITP.js";
 import deltaEOK from "./deltaEOK.js";
+import deltaEHCT from "./deltaEHCT.js";
 
-export { deltaE76, deltaECMC, deltaE2000, deltaEJz, deltaEITP, deltaEOK };
+export { deltaE76, deltaECMC, deltaE2000, deltaEJz, deltaEITP, deltaEOK, deltaEHCT };
```
LeaVerou, line 9:
```suggestion
export { 
	deltaE76,
	deltaECMC,
	deltaE2000,
	deltaEJz,
	deltaEITP,
	deltaEOK,
	deltaEHCT
};
```

src/toGamut.js
```diff
@@ -46,26 +81,59 @@ export default function toGamut (color, { method = defaults.gamut_mapping, space
...
 				let mapSpace = coordMeta.space;
 				let coordId = coordMeta.id;
 
-				let mappedColor = to(spaceColor, mapSpace);
+				let mappedColor = to(color, mapSpace);
 				let bounds = coordMeta.range || coordMeta.refRange;
 				let min = bounds[0];
-				let ε = .01; // for deltaE
+				let ε = calcEpsilon(jnd);
+				let minConverge = Math.max(ε, 1e-6); // limit delta of low and high to prevent infinite looping
```
LeaVerou, line 127: Why do we need a separate variable here? Can't we apply the clamping to ε directly?

src/deltaE/deltaEHCT.js
```diff
@@ -0,0 +1,42 @@
+import hct from "../spaces/hct.js";
+import {viewingConditions} from "../spaces/hct.js";
+
+const rad2deg = 180 / Math.PI;
+const deg2rad = Math.PI / 180;
+const ucsCoeff = [1.00, 0.007, 0.0228];
+
+
+function convertUcsAb (coords) {
+	// Convert HCT chroma and hue (CAM16 JMh colorfulness and hue) using UCS logic for a and b.
```
LeaVerou, line 10: Nit: shouldn't the comment of what the function does be above the function?

src/toGamut.js
facelessuser: Do you prefer I limit the JND? In the end, I need to place some limit on size so we never enter an infinite loop. To be honest, I don't know why the same epsilon is used in these two cases as they are measuring different things.

src/toGamut.js
jgerigmeyer: Sure, I'm happy to help with type defs -- I think you both of these will be relatively straightforward changes, but let me know if you have questions.

`deltaEHCT` can duplicate the other `deltaE*` types in `types/src/deltaE/`.
`toGamut` needs the new options added in `types/src/toGamut.d.ts`, e.g.:

```ts
declare function toGamut (
	color: ColorTypes,
	options?: {
		method?: string | undefined;
		space?: string | ColorSpace | undefined;
		deltaEMethod?: string | undefined;
		jnd?: number | undefined;
		blackWhiteClamp?: {
			channel: Ref;
			min: number;
			max: number;
		} | undefined;
	} | string
): PlainColorObject;
```

src/toGamut.js
```diff
@@ -46,26 +81,59 @@ export default function toGamut (color, { method = defaults.gamut_mapping, space
...
+					}
+				}
+			}
+
 			let clipped = toGamut(clone(spaceColor), { method: "clip", space });
-			if (deltaE2000(color, clipped) > 2) {
+			if (de(color, clipped) > jnd) {
+
+				// Clamp to SDR white and black if required
+				if (Object.keys(blackWhiteClamp).length > 0) {
```
jgerigmeyer, line 104: It looks like all three `channel`, `max`, and `min` are required -- should this `if` check for them explicitly?

src/toGamut.js
facelessuser: To be clear, I can use the same value for both of today is what is desired. Is my arbitrary limit satisfactory?

src/toGamut.js
facelessuser: You mean check for a length of 3? That seemed reasonable.

jgerigmeyer: Yeah, I was thinking either that or explicitly checking for the existence of `blackWhiteClamp.channel`, `blackWhiteClamp.max`, and `blackWhiteClamp.min`.

facelessuser: I'm not sure we want to waste the time for explicit checks. For now, I'll check for 3 unless others are adamant.

src/deltaE/deltaEHCT.js
facelessuser: Probably 😅.

src/toGamut.js
facelessuser: I went ahead and combined them.

types/src/deltaE/deltaEHCT.d.ts
```diff
@@ -0,0 +1,5 @@
+import Color, { ColorObject } from "../color.js";
+export default function (
```
jgerigmeyer, line 2: This should also be exported from `types/src/deltaE/index.d.ts` along with the other deltaE methods.

types/src/toGamut.d.ts
```diff
@@ -10,6 +10,13 @@ declare function toGamut (
 	options?: {
 		method?: string | undefined;
 		space?: string | ColorSpace | undefined;
+		deltaEMethod?: string | undefined;
+		jnd?: number | undefined;
+		blackWhiteClamp?: {
+			channel: string;
```
jgerigmeyer, line 16: It seems like this could be a `Ref` (imported from `types/src/space.d.ts`) instead of limited to a `string`?

jgerigmeyer (approved):
Should we also add tests for `blackWhiteClamp`? I don't see any.

types/src/toGamut.d.ts
facelessuser: The current intention was a string. I'm not quite sure what you mean by Ref though.

types/src/deltaE/deltaEHCT.d.ts
facelessuser: Okay

facelessuser:
> Should we also add tests for `blackWhiteClamp`? I don't see any.

It is indirectly tested in https://github.com/color-js/color.js/pull/420/files#diff-1b0f3b2f1f7d58d10d552dbdb4c7fff78b66b41a159f75db8fd5622da772205bR220. `hct-tonal` utilizes the black/white clamping. It ensures we get a proper black and white for tone 0 and 100.

types/src/toGamut.d.ts
jgerigmeyer: It's not a big deal, but from the implementation it looks like `blackWhiteClamp.channel` is only used to pass to `ColorSpace.resolveCoord`. [That function](https://github.com/color-js/color.js/blob/main/types/src/space.d.ts#L65) can accept a [Ref](https://github.com/color-js/color.js/blob/main/types/src/space.d.ts#L43-L46), which can be a string or array or object -- so I think any of those would also be acceptable here?

jgerigmeyer:
> > Should we also add tests for `blackWhiteClamp`? I don't see any.
> 
> It is indirectly tested in https://github.com/color-js/color.js/pull/420/files#diff-1b0f3b2f1f7d58d10d552dbdb4c7fff78b66b41a159f75db8fd5622da772205bR220. `hct-tonal` utilizes the black/white clamping. It ensures we get a proper black and white for tone 0 and 100.

Ah, I missed that -- thanks!

types/src/toGamut.d.ts
facelessuser: I guess? When I ran the type linting, it complained about `Ref`. I'm happy to allow it if it doesn't break anything and I can get the type checker to not yell at me 🙃.

jgerigmeyer: My guess is it was complaining because it doesn't exist in that file, and needs to be imported.

facelessuser: I think I fixed it.

svgeesus (approved):
This is all looking good to me

svgeesus:
I wish it was possible to have the [API documentation](https://github.com/color-js/color.js/pull/422#pullrequestreview-1874398861) be a bit less mysterious about the available options for the options objects on `[color.deltaE](https://deploy-preview-420--colorjs.netlify.app/api/#Color#deltaE)` and `[color.toGamut](https://deploy-preview-420--colorjs.netlify.app/api/#Color#toGamut)`.

I don't know if it is possible to automatically collect together the various options that have been implemented.

svgeesus:
> I wish it was possible

To clarify, that occurred to me while reading through this PR; but not as a blocker, just would like it to be possible.

facelessuser:
I've made one last change to ensure that if the color is already in the mapping color space, that undefined values are normalized to zero. The conversion to the target gamut space will resolve undefined values, but then you are gamut mapping with undefined values. The recent change fixes this oddity.

I think I am done unless other issues are discovered.

svgeesus:
@facelessuser I see some commits a few minutes ago, is this ready to merge? Won't merge until you say you are ready. But it would be nice to get this in the 0.5 release!

facelessuser:
@svgeesus I am good, my last comment states what I did, why I did it, and that I am done unless other issues are brought to my attention. Thanks for checking.

svgeesus:
> I've made one last change to ensure that if the color is already in the mapping color space, that undefined values are normalized to zero. The conversion to the target gamut space will resolve undefined values, but then you are gamut mapping with undefined values. The recent change fixes this oddity.

Perfect, I made a similar change to CSS Color 4 recently just to deal with powerless values. Converting "if needed" was ambiguous if the color was already in the destination color space.

PR #421, opened 2024-02-10
Add a workaround for Next.js production build bug

### LINKED ISSUE #260: .display and .toString different behavior on nextjs in dev and build mode
Trying to use `color.display({"format": "hex"})` as well as `.toString({"format": "hex"})` in **nextjs** v13. 
In `dev` it works fine, outputs `#111` for example, but in `prod` mode it outputs `color(srgb 0.0667 0.0667 0.0667)` when running bundle after `next build`.
So I use this method to get the same result in `dev` and in `prod` mode `Color.defaults.display_space.formats.hex.serialize(color.coords)`.
Is there any "correct" solution to output hexadecimal representation of colour (`#123456`)?

lloydk:
Workaround for a Next.js bug when generating a production build of a Next.js project.

Fixes #260

### DIFF
--- a/src/serialize.js
+++ b/src/serialize.js
@@ -27,6 +27,10 @@ export default function serialize (color, {
 	       ?? color.space.getFormat("default")
 	       ?? ColorSpace.DEFAULT_FORMAT;
 
+	// The assignment to coords and inGamut needs to stay in the order they are now
+	// The order of the assignment was changed as a workaround for a bug in Next.js
+	// See this issue for details: https://github.com/color-js/color.js/issues/260
+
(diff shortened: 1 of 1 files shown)


jgerigmeyer (approved):

MysteryBlokHed (approved):

LeaVerou (changes requested):
Can you add a comment that points to the issue so that we don't inadvertently change it in the future?

lloydk:
> Can you add a comment that points to the issue so that we don't inadvertently change it in the future?

Done

LeaVerou (approved):
Excellent, thank you!

PR #423, opened 2024-02-11
[website] Fix "Defined in" API links

MysteryBlokHed:
Currently, the API links on the webpage go to non-existent files on the [mavoweb/mavo](https://github.com/mavoweb/mavo) repository instead of this one. For example, this link goes to <https://github.com/mavoweb/mavo/blob/master/src/color.js>: 

A screenshot of the API page on colorjs.io. There is a hyperlink on the right, labelled "Defined in color.js"

This PR fixes the API's `index.njk` file to have the correct link at the base.


LeaVerou (approved):
Wow 🤣 how did this go undetected for so long?! Thanks for catching it!

PR #430, opened 2024-02-13
Add format to hsv color space

lloydk:
The hsv color space couldn't be parsed because it didn't have a format entry.

I wasn't sure if the `id` should be `hsv` or `--hsv`. Ideally it should be `--hsv` but I went with `hsv` because even though you couldn't parse hsv colors they were being serialized as `color(hsv ...)` so the serialization is still backwards compatible.

### DIFF
--- a/src/spaces/hsv.js
+++ b/src/spaces/hsv.js
@@ -54,5 +54,11 @@ export default new ColorSpace({
 			(l === 0 || l === 1) ? 0 : ((v - l) / Math.min(l, 1 - l)) * 100,
 			l * 100
 		];
+	},
+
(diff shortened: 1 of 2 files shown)


LeaVerou (changes requested):
Thanks! I think we should serialize as `--hsv` per our recent resolution that nonstandard spaces use dashed idents.

That said, if we have a colorspace with id "--foo" and we are parsing a color with id "foo", it would be nice to parse it correctly and print out a warning (or at least throw a meaningful error). But that's a separate feature.

LeaVerou (approved):

PR #431, opened 2024-02-13
Allow hsluv to be gamut mapped using the CSS algorithm

lloydk:
The `hsluv` color space wasn't being gamut mapped by the CSS algorithm because it was considered to be an unbounded space and the CSS algorithm exits early if a space is unbounded.

Changing the `s` and `l` coordinates to use `range` instead of `refRange` makes `hsluv` a bounded space and allows it to be gamut mapped properly.

Note that testing if a `hsluv` color was in gamut worked properly because the `gamutSpace` of `hsluv` is `srgb` which is a bounded space.

Convert app with `s` and `l` using `refRange`:

image

`s` and `l` changed to use `range`:

image

### DIFF
--- a/src/spaces/hsluv.js
+++ b/src/spaces/hsluv.js
@@ -97,11 +97,11 @@ export default new ColorSpace({
 			name: "Hue"
 		},
 		s: {
-			refRange: [0, 100],
+			range: [0, 100],
 			name: "Saturation"
 		},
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):

LeaVerou:
Thank you!

PR #435, opened 2024-02-14
Add missing formats to color spaces

lloydk:
Some color spaces couldn't be parsed because they didn't have any formats.

Added formats for:

- a98rgb-linear
- p3-linear
- prophoto-linear
- xyz-abs-d65

I used a dashed-ident for all of the spaces above similar to #430.

### DIFF
--- a/src/spaces/a98rgb-linear.js
+++ b/src/spaces/a98rgb-linear.js
@@ -23,5 +23,10 @@ export default new RGBColorSpace({
 	name: "Linear Adobe® 98 RGB compatible",
 	white: "D65",
 	toXYZ_M,
-	fromXYZ_M
+	fromXYZ_M,
+	formats: {
+		color: {
+			id: "--a98rgb-linear"
+		}
+	}
 });
--- a/src/spaces/p3-linear.js
+++ b/src/spaces/p3-linear.js
@@ -17,5 +17,10 @@ export default new RGBColorSpace({
 	name: "Linear P3",
 	white: "D65",
 	toXYZ_M,
-	fromXYZ_M
+	fromXYZ_M,
+	formats: {
+		color: {
+			id: "--p3-linear"
+		}
+	}
 });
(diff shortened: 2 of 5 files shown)


LeaVerou (changes requested):
Hey, thanks for working on this! 

In general, as an API Design principle, I think boilerplate should be avoided. It makes for annoying DX, increases the chance of errors, and makes it harder to change things. Software is much better than humans at generating repetitive structures; therefore any boilerplate needed should be generated automatically by the software.

How dies this apply here? So, I think we should handle this a bit differently (which is how I *thought* it was handled, and maybe it was and it broke at some point): *Every* color space, automatically, gets a `color()` format supporting a dashed ident and/or ident for the color name plus `<number> | <percentage>` for the coords (or `<percentage> | `<number>`, I'm fine either way). Authors can still specify `color` formats explicitly, and they will override this but we should not risk having color spaces with no formats because an author forgot to include this.

What do you think?

lloydk:
> How dies this apply here? So, I think we should handle this a bit differently (which is how I _thought_ it was handled, and maybe it was and it broke at some point): _Every_ color space, automatically, gets a `color()` format supporting a dashed ident and/or ident for the color name plus `<number> | <percentage>` for the coords (or `<percentage> | ``, I'm fine either way). Authors can still specify `color` formats explicitly, and they will override this but we should not risk having color spaces with no formats because an author forgot to include this.
> 

That's similar to an idea I've had for a while which is the `coordGrammer` for most spaces could be generated from the coordinate metadata. So in addition to number or percentage angle coordinates would also be supported in the auto generated `color` format.

LeaVerou:
> That's similar to an idea I've had for a while which is the `coordGrammer` for most spaces could be generated from the coordinate metadata. So in addition to number or percentage angle coordinates would also be supported in the auto generated `color` format.

That would not be spec compliant as it's unlikely for `color()` to ever support angles. I think for spaces that need angles, it makes more sense to have dedicated functions (potentially prefixed? Not sure).

LeaVerou:
@lloydk I *think* so (together with a custom function), but we should discuss in a new issue.

lloydk:
> @lloydk I _think_ so (together with a custom function), but we should discuss in a new issue.

Created a new issue #442

PR #436, opened 2024-02-14
[types] Add JSDoc based on website docs

### LINKED ISSUE #424: [chore] Use typedoc for API docs
Our `/api` docs are currently a total mess because they are not being updated. Instead, the only docs we really have are those under `/docs`. We also need proper low level API docs.

I’ve recently started using typedoc in other projects and it seems to work better than any of the others I’ve tried from time to time.

MysteryBlokHed:
Related to #424

This is a work-in-progress PR to add documentation from the website to the types. This will not only allow for better developer experience by providing more informative in-editor documentation, but will also allow the docs to be auto-generated with a tool like TypeDoc in the future.

I've started by adding docs for the `Hooks` and `ColorSpace` classes, but docs for other classes/functions/properties should be added too.

If there are any classes or functions in particular that anyone thinks should have documentation brought over from the website, I'm open to suggestions. Some things seem relatively self-explanatory with their types, so I don't think that _every_ method and property needs to include super-detailed docs.

**Note:** In the `toGamut.d.ts` file, I did actually modify the types themselves a bit in order to make them more accurate, and to make it easier to access the types for the function's options.

### DIFF
--- a/types/src/space.d.ts
+++ b/types/src/space.d.ts
@@ -27,15 +27,34 @@ export interface CoordMeta {
 }
 
 export interface Options {
+	/** Id of this space, used in things such as conversions */
 	id: string;
+	/** The readable name of the space, used in user-facing UI */
 	name: string;
+	/** The base color space */
 	base?: string | ColorSpace | null | undefined;
+	/**
+	 * Function that converts coords in the base color space to coords in this color space.
+	 * Must perform chromatic adaptation if needed
+	*/
 	fromBase?: ((coords: Coords) => number[]) | undefined;
+	/**
+	 * Function that converts coords in this color space to coords in the base color space.
+	 * Must perform chromatic adaptation if needed
+	 */
 	toBase?: ((coords: Coords) => number[]) | undefined;
+	/**
+	 * Object mapping coord ids to coord metadata
+	 * @see {@link CoordMeta}
+	*/
 	coords?: Record<string, CoordMeta> | undefined;
 	white?: string | White | undefined;
 	cssId?: string | undefined;
 	referred?: string | undefined;
+	/**
+	 * Details about string formats to parse from / serialize to
+	 * @see {@link Format}
+	*/
 	formats?: Record<string, Format> | undefined;
 	gamutSpace?: "self" | string | ColorSpace | null | undefined;
 }
@@ -45,6 +64,7 @@ export type Ref =
 	| [string | ColorSpace, string]
 	| { space: string | ColorSpace; coordId: string };
 
+/** Class for color spaces. Each color space corresponds to a `ColorSpace` instance */
 export default class ColorSpace {
 	constructor (options: Options);
 
--- a/types/src/space.d.ts
+++ b/types/src/space.d.ts
@@ -2,7 +2,9 @@ import { White } from "./adapt.js";
 import Color, { ColorConstructor, ColorObject, Coords } from "./color.js";
 
 export interface Format {
+	/** @default "function" */
 	type?: string | undefined;
+	/** @default "color" */
 	name?: string | undefined;
 	id?: string | undefined;
 	coords?: string[] | undefined;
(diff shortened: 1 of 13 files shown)


MysteryBlokHed:
While not in the website, this could probably use some documentation for each property:

<https://github.com/color-js/color.js/blob/4dcc46bd071118c3bda37a6710e5477dd045586f/types/src/space.d.ts#L4-L27>

MysteryBlokHed:
This entire file could also probably use documentation:

<https://github.com/color-js/color.js/blob/5727b1c954aa8e4d9a452ab08c087478a585f5f6/types/src/interpolation.d.ts#L1-L50>

types/src/toGamut.d.ts
```diff
@@ -1,24 +1,45 @@
...
+	 *
+	 * If `"clip"`, coordinates are just clipped to their reference range.\
+	 * If `"css"`, coordinates are reduced according to the CSS 4 Gamut Mapping Algorithm.\
+	 * If in the form `[colorSpaceId].[coordName]`, that coordinate is reduced
+	 * until the color is in gamut. Please note that this may produce nonsensical
+	 * results for certain coordinates (e.g. hue) or infinite loops
+	 * if reducing the coordinate never brings the color in gamut
+	 */
+	// eslint-disable-next-line @typescript-eslint/ban-types
+	method?: "css" | "clip" | (string & {}) | undefined;
```
MysteryBlokHed, line 17: FYI, the `string & {}` type here is done for intellisense while writing code. Doing this means that editors will automatically suggest `"css"` and `"clip"` while typing, but any string type will still be accepted. If the type were just `string` instead, then the entire type would be simplified into `string | undefined` and the intellisense wouldn't be as helpful.

jgerigmeyer: Whoa, TIL 🎉

types/src/space.d.ts
```diff
@@ -27,15 +38,34 @@ export interface CoordMeta {
...
+	 * Must perform chromatic adaptation if needed
+	 */
 	toBase?: ((coords: Coords) => number[]) | undefined;
+	/**
+	 * Object mapping coord ids to coord metadata
+	 * @see {@link CoordMeta}
+	*/
 	coords?: Record<string, CoordMeta> | undefined;
 	white?: string | White | undefined;
 	cssId?: string | undefined;
```
jgerigmeyer, line 63: Probably worth documenting this, as the id used by CSS, e.g. `display-p3` or `--cam16-jmh`.

types/src/toGamut.d.ts
jgerigmeyer: Should we note that this uses `css` by default?

types/src/toGamut.d.ts
```diff
@@ -1,24 +1,45 @@
...
+	 * How to force into gamut.
+	 *
+	 * If `"clip"`, coordinates are just clipped to their reference range.\
+	 * If `"css"`, coordinates are reduced according to the CSS 4 Gamut Mapping Algorithm.\
+	 * If in the form `[colorSpaceId].[coordName]`, that coordinate is reduced
+	 * until the color is in gamut. Please note that this may produce nonsensical
+	 * results for certain coordinates (e.g. hue) or infinite loops
+	 * if reducing the coordinate never brings the color in gamut
+	 */
+	// eslint-disable-next-line @typescript-eslint/ban-types
```
jgerigmeyer, line 16: My assumption is that @LeaVerou would prefer that we adjust the lint rule rather than overriding it in the codebase. If that's the case, we could remove this and use the following for the rule in `.eslintrc.cjs` to allow `{}` types:

```js
		"@typescript-eslint/ban-types": [
			1,
			{
				"extendDefaults": true,
				"types": {
					"{}": false,
				},
			},
		],
```

jgerigmeyer (approved):
Looks good, thanks @MysteryBlokHed!

types/src/toGamut.d.ts
MysteryBlokHed: Probably not a bad idea. I'll add that when I get a chance

types/src/toGamut.d.ts
MysteryBlokHed: IMO it might be worth keeping it as a comment to just disable it for this line, since there are scenarios where `{}` might accidentally end up used where it shouldn't be (like somebody accidentally using it where they meant to have a type that accepts any object). If there were an eslint rule to allow specifically `string & {}` just for this use case then that would be ideal, but I don't know whether that's actually possible.

We can wait to see if @LeaVerou has thoughts on which approach is better

LeaVerou: @jgerigmeyer is right. (I should write up about this somewhere to explain my reasoning so it doesn't just live in random PRs)

MysteryBlokHed:
Oops, it looks like there's a formatting problem that made CI fail. I'll open a new PR to fix that.

PR #439, opened 2024-02-14
Parse both dashed-ident and ident versions, and add missing dashed-idents to color spaces.

### LINKED ISSUE #433: (Some) nonstandard color spaces error when round-tripping
~**This is a blocker for v0.5.0 as it's a pretty serious regression**~ Turns out this exists in v0.4.3 too 🤦🏽‍♀️

Try this in the console:
```
new Color(new Color("red").to("p3-linear").toString())
```

image

jgerigmeyer:
Partial solution for #433, as an alternative to #435.

This PR makes a few changes, and I'm not tied strongly to any of them -- this is a starting point for discussion.

- Parses both `<dashed-ident>` and `<ident>` versions of `color()` spaces, warning if e.g. `color(rec2100-hlg)` is used instead of `color(--rec2100-hlg)`, and similarly if `color(--display-p3)` is used instead of `color(display-p3)`.
- Simplifies color space options, where now _every_ color gets `formats.color.id` defined if it's missing -- defaults to `cssId` if provided, otherwise `id`.
- Adds missing `cssId` option to color spaces, using dashed-ident where appropriate, and removes unnecessary `formats.color` definitions.
- Adds parsing tests for all missing color spaces.

This PR does _not_ attempt to auto-generate `coordGrammer` for color spaces.

This shouldn't be a breaking change in terms of parsing colors (and it does fix #433), but it is a breaking change in the sense that now e.g. `color(acescc)` will be serialized as `color(--acescc)`.


src/spaces/prophoto-linear.js
```diff
@@ -19,6 +19,7 @@ const fromXYZ_M = [
 
 export default new RGBColorSpace({
 	id: "prophoto-linear",
+	cssId: "--prophoto-rgb-linear",
```
jgerigmeyer, line 22: Should this be `--prophoto-linear`? I went with `--prophoto-rgb-linear` for consistency with `color(prophoto-rgb)`.

src/parse.js
```diff
@@ -102,17 +103,24 @@ export default function parse (str, {meta} = {}) {
 							Object.assign(meta, {formatId: "color", types});
 						}
 
+						if (colorSpec.id.startsWith("--") && !id.startsWith("--")) {
```
lloydk, line 106: I think the warning messages could be made more clear

`Use color(--hsluv ...) instead of color(hsluv ...)`

Not super happy with this suggestion so feel free to ignore it.

lloydk (commented):
Should the HDR color spaces in [CSS Color HDR Module Level 1](https://drafts.csswg.org/css-color-hdr) have a \<dashed-ident\>?

Will we remove the \<dashed-ident\> `cssId` as a breaking change if/when the HDR spaces are implemented in browsers?

I think we should go with unprefixed ids for the HDR spaces as they'll most likely be made a standard part of CSS.

Overall I prefer this PR to #435.

src/parse.js
lloydk: Actually, I'm wondering if we really need the warning messages. It's not clear to me who the warning messages are for. Developers? End Users?

As an end user of the convert app I'm happy that the app is liberal in what it accepts as I frequently forget to use the \<dashed-ident\> so I don't care if there's a warning message in the console.

Running the tests in the browser prints out a bunch of warning messages and I'm assuming that running the tests from the command line will do the same.

LeaVerou: Well, the prefixing *does* serve a purpose. That said, I’m happy to not have warnings for now, and become stricter later.

src/spaces/a98rgb.js
```diff
@@ -3,13 +3,9 @@ import A98Linear from "./a98rgb-linear.js";
 
 export default new RGBColorSpace({
 	id: "a98rgb",
+	cssId: "a98-rgb",
```
LeaVerou, line 6: Why do we need to specify this? It should default to `this.id`. This was meant for things like `p3` -> `display-p3`. I can also see having it for dashed-ident color spaces. But for the default ones, why?

src/parse.js
jgerigmeyer: As a developer I personally like the warnings because they flag an issue without preventing me from using it, but I can remove these if you like -- @LeaVerou is that your preference?

src/spaces/a98rgb.js
jgerigmeyer: @LeaVerou Correct, this is only for cases where the `cssId` is different than the `id` -- but that's the case here too... `a98rgb` is different than `a98-rgb`. Am I misunderstanding your question?

jgerigmeyer: This doesn't change anything functionally, but uses the shorter/clearer `cssId: "a98-rgb"` instead of `formats: { color: { id: "a98-rgb" } }`.

jgerigmeyer:
> Should the HDR color spaces in [CSS Color HDR Module Level 1](https://drafts.csswg.org/css-color-hdr) have a <dashed-ident>?
>
> I think we should go with unprefixed ids for the HDR spaces as they'll most likely be made a standard part of CSS.

I went with `<dashed-ident>` for color spaces that aren't currently part of the spec even if they will be added, but I'm fine with whatever @LeaVerou prefers here.

> Will we remove the <dashed-ident> `cssId` as a breaking change if/when the HDR spaces are implemented in browsers?

Yes, my assumption is that as colors are added to the spec and implemented in browsers, we will remove the dashed-ident from their `cssId`. It's not entirely a breaking change in that we will still parse those colors with or without the prefix, but it would mean that they would be serialized without the dashed-ident going forward.

> Overall I prefer this PR to #435.

🎉

src/parse.js
lloydk: What about disabling the warnings if we're in a test environment?

The htest cli runner could set `process.env.NODE_ENV` to `"test"`. Not sure if there is a convention is for browser test runners.

src/spaces/a98rgb.js
lloydk: Should we consider this a bug in the `id` and change the id to `"a98-rgb"`, remove the `cssId` and add an alias for `a98rgb`. If not then we should add an alias of `"a98-rgb"` to the color space.

src/parse.js
LeaVerou: What @jgerigmeyer meant was that the warnings are helpful for developers using Color.js, not just for us running tests. I think it would be good to have a production version and a development version, like Vue does, though that's a much bigger change. I vote to keep the warnings for now.

src/parse.js
```diff
@@ -102,17 +103,24 @@ export default function parse (str, {meta} = {}) {
...
+							console.warn(`Use prefixed color(${colorSpec.id}) instead of color(${id}).`);
```
LeaVerou, line 107: We need to clarify *why* this distinction exists, otherwise the warning makes it sound like Color.js is annoying for no reason. E.g. something like:

```suggestion
							console.warn(`${space.name} is a non-standard space, so you should use color(${colorSpec.id}) instead of color(${id}).`);
```

src/parse.js
```diff
@@ -102,17 +103,24 @@ export default function parse (str, {meta} = {}) {
...
+
 						return {spaceId: space.id, coords, alpha};
 					}
 				}
 			}
 
 			// Not found
 			let didYouMean = "";
-			let registryId = id in ColorSpace.registry ? id : undashedId;
+			let registryId = id in ColorSpace.registry ? id : alternateId;
```
LeaVerou, line 120: Nice, I would have *really* struggled with how to name this variable 😅

src/spaces/a98rgb.js
LeaVerou: Oh I see, there's an extra hyphen in the CSS version. I totally missed that and thought you were repeating the same id (which should not be needed)

LeaVerou: > Should we consider this a bug in the id and change the id to "a98-rgb", remove the cssId and add an alias for a98rgb. If not then we should add an alias of "a98-rgb" to the color space.

Since the color space ids are also used as a accessors, it would provide poor ergonomics to have ids with hyphens, so I think this is fine.

src/spaces/p3-linear.js
```diff
@@ -14,6 +14,7 @@ const fromXYZ_M = [
 
 export default new RGBColorSpace({
 	id: "p3-linear",
+	cssId: "--display-p3-linear",
```
LeaVerou, line 17: Oof. I understand why it's as awkward as it is, but it makes me sad.

I wonder if it may make sense to allow providing multiple dashed ids, so we could also provide `--p3-linear` but that's a out of scope for this PR…

src/spaces/p3.js
```diff
@@ -4,14 +4,10 @@ import sRGB from "./srgb.js";
...
 	base: P3Linear,
 	// Gamma encoding/decoding is the same as sRGB
 	fromBase: sRGB.fromBase,
-	toBase: sRGB.toBase,
-	formats: {
-		color: {
-			id: "display-p3",
-		}
-	},
+	toBase: sRGB.toBase
```
LeaVerou, line 12: Does our linting disallow trailing commas? If so, it shouldn't. I think trailing commas in things like this would really help minimize the diffs.

LeaVerou (approved):
This is fantastic work @jgerigmeyer, thank you so much for working on this!!

My only substantive comment at this point is that I'd prefer to keep the trailing commas to keep the diff small, but even that I wouldn't elevate it to a blocker.

src/parse.js
lloydk: My concern was more about seeing the warnings every time I run the conversion tests. But it's not a showstopper.

lloydk (approved):
LGTM

src/parse.js
jgerigmeyer: @LeaVerou Adjusted in https://github.com/color-js/color.js/pull/439/commits/b1cf95a4a68dd7c54f0a6e62609318be780e9d92 -- what do you think?

src/spaces/p3.js
jgerigmeyer: Good call. Right now our linter is agnostic about trailing commas, so I tried to match the pattern I found in the files. But I agree it results in noisy diffs, so I adjusted in https://github.com/color-js/color.js/commit/b1cf95a4a68dd7c54f0a6e62609318be780e9d92.

I also opened #440 to separately consider enforcing trailing commas.

src/parse.js
jgerigmeyer: That makes some sense to me @lloydk. I pushed one approach that skips these warnings in tests, @LeaVerou what do you think? #441

svgeesus:
I like the idea of parsing both the dashed and undashed forms.

> I went with `<dashed-ident>` for color spaces that aren't currently part of the spec even if they will be added, but I'm fine with whatever @LeaVerou prefers here.

Note that [CSS Color HDR](https://drafts.csswg.org/css-color-hdr/) is now  [an official CSSWG spec](https://github.com/w3c/csswg-drafts/issues/553#issuecomment-1864865513), building on CSS Color 4 which remains SDR-only. So we should not have dashed idents for things already part of that spec.

See also:

 - https://github.com/color-js/color.js/issues/443

src/spaces/ictcp.js
```diff
@@ -66,6 +66,7 @@ const LMStoXYZ_M = [
 // https://professional.dolby.com/siteassets/pdfs/dolby-vision-measuring-perceptual-color-volume-v7.1.pdf
 export default new ColorSpace({
 	id: "ictcp",
+	cssId: "--ictcp",
```
svgeesus, line 69: This should be undashed: https://drafts.csswg.org/css-color-hdr/#ICtCp

src/spaces/jzazbz.js
```diff
@@ -112,6 +112,7 @@ export default new ColorSpace({
 	formats: {
 		// https://drafts.csswg.org/css-color-hdr/#Jzazbz
 		"color": {
+			id: "--jzazbz",
```
svgeesus, line 115: This should be undashed  https://drafts.csswg.org/css-color-hdr/#Jzazbz

src/spaces/jzczhz.js
```diff
@@ -4,6 +4,7 @@ import {constrain as constrainAngle} from "../angles.js";
 
 export default new ColorSpace({
 	id: "jzczhz",
+	cssId: "--jzczhz",
```
svgeesus, line 7: This should be undashed https://drafts.csswg.org/css-color-hdr/#JzCzHz

src/spaces/p3-linear.js
svgeesus: we should not have called it just "p3" because that is ambiguous. Cinema uses DCI-P3, video uses PR-D65 and CSS (among other things) uses Display-P3.

src/spaces/rec2100-hlg.js
```diff
@@ -11,7 +11,7 @@ const scale = 3.7743;	// Place 18% grey at HLG 0.38, so media white at 0.75
 
 export default new RGBColorSpace({
 	id: "rec2100hlg",
-	cssId: "rec2100-hlg",
+	cssId: "--rec2100-hlg",
```
svgeesus, line 14: This should be undashed https://drafts.csswg.org/css-color-hdr/#valdef-color-rec2100-hlg

src/spaces/rec2100-pq.js
```diff
@@ -12,6 +12,7 @@ const c3 = 2392 / (2 ** 7);
 
 export default new RGBColorSpace({
 	id: "rec2100pq",
+	cssId: "--rec2100-pq",
```
svgeesus, line 15: This should be undashed https://drafts.csswg.org/css-color-hdr/#valdef-color-rec2100-pq

svgeesus (changes requested):

jgerigmeyer:
> Note that [CSS Color HDR](https://drafts.csswg.org/css-color-hdr/) is now [an official CSSWG spec](https://github.com/w3c/csswg-drafts/issues/553#issuecomment-1864865513), building on CSS Color 4 which remains SDR-only. So we should not have dashed idents for things already part of that spec.

@svgeesus Adjusted in https://github.com/color-js/color.js/pull/439/commits/da4dcb35043c824d759d07dae52102c37c14bbed

svgeesus (approved):
LGTM

PR #440, opened 2024-02-16
Enforce trailing commas unless `]` or `}` is on the same line.

jgerigmeyer:


LeaVerou (approved):
Wow that was fast! Thanks!

PR #444, opened 2024-02-18
[ci] Update lint rules for PR's

MysteryBlokHed:
A PR I just merged (#436) passed all the CI on the pull request page, but failed once it was merged into main. This is because the lint CI didn't actually run on the PR.

This change makes the CI rerun when a PR is opened, reopened, and when commits are pushed.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,7 +2,6 @@ name: Lint & Test Types
 on:
     push:
     pull_request:
-        types: [reopened]
 
 concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


jgerigmeyer (approved):
Interesting -- I thought `push:` would cover this case as well. 😕

MysteryBlokHed:
> Interesting -- I thought `push:` would cover this case as well. 😕 

I don't blame you—Honestly, I feel like some of the conditions to run CI are pretty poorly documented on GitHub's pages. I found a list of valid `type` values for the pull request trigger, but only a couple had an actual explanation as to what they were.

PR #447, opened 2024-02-19
New gamut mapping approaches

facelessuser:
Scale Achroma: which is like scale, but scales the color towards an achromatic version of itself based on the channel furthest out of gamut.

Scale LH Achroma: Like Scale LH, but uses Scale Achroma. Also clamps white and black if L >1 or L < 0.

All are performed in linear Display P3 and OkLCh.

### DIFF
--- a/apps/gamut-mapping/methods.js
+++ b/apps/gamut-mapping/methods.js
@@ -51,7 +51,7 @@ const methods = {
 				let chroma = mapColor.coords[1];
 				method.scale(gamutColor, achroma);
 				let chroma2 = gamutColor.oklch.c;
-				mapColor.set("c", ((chroma2 - chroma) < 2) ? method.lerp(chroma, chroma2, 0.9) : chroma2 + 0.1);
+				mapColor.set("c", ((chroma2 - chroma) < 2) ? method.lerp(chroma, chroma2, 0.85) : chroma2 + 0.1);
 				let [red, green, blue] = mapColor.to("p3-linear").coords;
 				gamutColor.set({r: red, g: green, b: blue});
 			}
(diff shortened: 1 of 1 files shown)


facelessuser:
I spent some time refining this approach to make it worth submitting. Generally, I tend to prefer Scale LH Achromatic 2. Generally, it is in the ∆E range of Scale LH, give or take some, but hue delta and lightness delta are lower. Obviously, there are different metrics that one may judge GMAs on, but I'll just give an overview of my observations. It at least fixes many of the things that bothered me about Scale LH.

Screenshot 2024-02-19 at 10 09 18 AM

Tone mapping seems to perform better using the achromatic scaling:

Screenshot 2024-02-19 at 10 21 37 AM

Gradients often seem less surprising using achromatic scaling:

Screenshot 2024-02-19 at 10 25 42 AM

Gamut mapping in relation to other colors seems more sane:

Scale LH

flowers-scale-lh

Scale LH Achromatic

flowers-scale-achroma

Scale LH Achromatic 2

flowers-scale-achroma2

facelessuser:
I'm actually pulling Scale LH Achromatic 2 for now. I think it needs a bit more time to bake as there were still a few corner cases. 

I think Scale LH Achromatic is actually much more reliable in its current state and I'm generally pleased with how it performs.  It gives good tones, good gradients, etc. and is much faster than the CSS GMA.

It should be noted that the dark blue region is the one place where Scale LH keeps ∆h closer than Scale LH Achromatic, but it does so by sacrificing ∆L. This, for me, is why blues seem to gamut map so abruptly with Scale LH in some cases (images as an example). These changes in lightness seem more jarring (at least to me) when viewed in context with gradients and images.  When this is occurring though ∆E is still comparable and often even better for Scale LH Achromatic, which seems to reinforce my feelings about the issues with scaling to the midpoint. Again, this is just my personal opinion.

Anyway, I think I am done tinkering with this for now. I'm pretty pleased with the alternate scaling. Hopefully, others will find this interesting.

facelessuser:
Closing in favor of #449 which improves the over-correction that could occur (particularly in the low light blue that I mentioned earlier).

PR #449, opened 2024-02-22
Add new GMA Scale LH Achromatic

facelessuser:
Performs a similar approach as Scale LH, but instead scales color towards the achromatic version of itself. This is performed in a linear light RGB space. When scaling the color towards the achromatic color, it is done in an LRGB form that separates the lightness from the other components. This helps preserve lightness during the process and performs even better, especially in low light blue, which was more problematic in my first attempt.

The procedure is competitive with Scale LH using two iterations, performs even better with 4. 3 iterations was chosen as a middle ground.

Replaces #447

### DIFF
--- a/apps/gamut-mapping/methods.js
+++ b/apps/gamut-mapping/methods.js
@@ -57,6 +57,7 @@ const methods = {
 		},
 
 		lrgbToSrgb: (lrgb) => {
+			// From Prismatic https://studylib.net/doc/14656976/the-prismatic-color-space-for-rgb-computations
 			let rgb = lrgb.slice(1);
 			let l = lrgb[0];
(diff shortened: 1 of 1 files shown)


facelessuser:
For reference, this is the source from where the LRGB approach came from: https://studylib.net/doc/14656976/the-prismatic-color-space-for-rgb-computations

facelessuser:
I'm putting this in draft as there are a few more things I'm exploring.

facelessuser:
After more experimentation, I'm abandoning in favor of #452. It just seems to do better. The use of LRGB was also a misstep as evaluations revealed lightness of out of gamut colors did not behave as I thought they did. Basically, while it did help reduce hue in some cases, it was purely by accident.

PR #451, opened 2024-02-23
Allow any color type when calling individual deltaE functions

lloydk:
The types for the deltaE functions didn't allow all ColorTypes as parameters. For example, the types didn't allow "red" to be passed as a parameter.

Also, calling a deletaE method with a color such as "red" would cause an exception to be thrown.
```javascript
let x = deltaE2000("red", "white") // Exception thrown
```

### DIFF
--- a/src/deltaE.js
+++ b/src/deltaE.js
@@ -1,4 +1,3 @@
-import getColor from "./getColor.js";
 import defaults from "./defaults.js";
 import {isString} from "./util.js";
 import deltaEMethods from "./deltaE/index.js";
@@ -9,9 +8,6 @@ export default function deltaE (c1, c2, o = {}) {
 	}
 
 	let {method = defaults.deltaE, ...rest} = o;
-
-	c1 = getColor(c1);
-	c2 = getColor(c2);
 
 	for (let m in deltaEMethods) {
 		if ("deltae" + method.toLowerCase() === m.toLowerCase()) {
--- a/src/deltaE/deltaE2000.js
+++ b/src/deltaE/deltaE2000.js
@@ -1,5 +1,6 @@
 import lab from "../spaces/lab.js";
 import lch from "../spaces/lch.js";
+import getColor from "../getColor.js";
 
 // deltaE2000 is a statistically significant improvement
 // and is recommended by the CIE and Idealliance
@@ -24,6 +25,9 @@ function pow7 (x) {
 }
 
 export default function (color, sample, {kL = 1, kC = 1, kH = 1} = {}) {
+	color = getColor(color);
+	sample = getColor(sample);
+
 	// Given this color as the reference
 	// and the function parameter as the sample,
 	// calculate deltaE 2000.
--- a/src/deltaE/deltaE2000.js
+++ b/src/deltaE/deltaE2000.js
@@ -25,8 +25,7 @@ function pow7 (x) {
 }
 
 export default function (color, sample, {kL = 1, kC = 1, kH = 1} = {}) {
-	color = getColor(color);
-	sample = getColor(sample);
+	[color, sample] = getColor([color, sample]);
 
 	// Given this color as the reference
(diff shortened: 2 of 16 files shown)


MysteryBlokHed (commented):
TypeScript changes LGTM

LeaVerou:
Agree on solving the problem, not sure about this particular solution, I really don't like the repetition. Can we make this a utility function from `deltaE.js` that these methods import? (with that plus any other commonalities between the functions)

LeaVerou:
Agree on solving the problem, not sure about this particular solution, I really don't like the repetition. If there are other commonalities between the deltaE functions, we could make this part of a utility function from `deltaE.js` that these methods import.

If there aren't other commonalities, maybe a better solution is to make `getColor()` accept a variable number of arguments so it can become a single line:

```js
let [sample, color] = getColor(sample, color);`
```

LeaVerou (changes requested):
Agree on solving the problem, not sure about this particular solution, I really don't like the repetition. If there are other commonalities between the deltaE functions, we could make this part of a utility function from `deltaE.js` that these methods import.

If there aren't other commonalities, maybe a better solution is to make `getColor()` accept a variable number of arguments so it can become a single line:

```js
let [sample, color] = getColor(sample, color);`
```

lloydk:
> Agree on solving the problem, not sure about this particular solution, I really don't like the repetition. If there are other commonalities between the deltaE functions, we could make this part of a utility function from `deltaE.js` that these methods import.
> 

I don't think there are any commonalities between the deltaE functions.

The contrast functions have the same repetition. 

https://github.com/color-js/color.js/blob/1281ad1f6a3da6cea4929d43048d36981003d650/src/contrast/APCA.js#L41-L43


> If there aren't other commonalities, maybe a better solution is to make `getColor()` accept a variable number of arguments so it can become a single line:
> 


What about adding a `getColors` function to `getColor.js`? 
```javascript
export function getColors (...colors) {
	return colors.map(getColor);
}
```

I don't have a strong opinion on any of the solutions:
- keep the repetition
- change `getColor` to take a variable number of arguments
- add a `getColors` function

If we want to make changes to `getColor` or add a `getColors` function I can create a separate PR for that change.

LeaVerou:
How about this option: instead of variable arguments to `getColor()` we support an array argument. If an array is passed in, an array is returned.

Pros:
- Skips the weird discontinuity at 1 -> 2 args
- Explicit opt in to array handling, possible to handle arrays with even 1 element
- Allows us to add an options object later on

It does add two extra chars to usage, but I think that's an ok tradeoff, and it makes it more explicit.

The advantage of extending `getColor` instead of adding a new function is also that this means `Color.get()` gets this for free.

lloydk:
> How about this option: instead of variable arguments to `getColor()` we support an array argument. If an array is passed in, an array is returned.
> 
> Pros:
> 
> * Skips the weird discontinuity at 1 -> 2 args
> * Explicit opt in to array handling, possible to handle arrays with even 1 element
> * Allows us to add an options object later on
> 
> It does add two extra chars to usage, but I think that's an ok tradeoff, and it makes it more explicit.
> 
> The advantage of extending `getColor` instead of adding a new function is also that this means `Color.get()` gets this for free.

Created PR #453 for `getColor`. 

When you say `Color.get()` do you mean this function or am I missing something?

https://github.com/color-js/color.js/blob/52d5f69f3fbc25e602ba47a5872f2543c20e194d/src/color.js#L111-L121

LeaVerou:
> When you say Color.get() do you mean this function or am I missing something?

Ah, I thought it was using `getColor`. We may want to update that too then to keep them in sync.

lloydk:
Messed up my branch. Hopefully I can fix it without making more of a mess.

jgerigmeyer (approved):

LeaVerou (approved):
LGTM!

PR #452, opened 2024-02-24
Add GMA method that uses raytracing

facelessuser:


facelessuser:
I think with my various attempts, I learned a lot about what wasn't working and why it wasn't working. I realize now, that many of these approaches cannot eliminate hue shifts completely unless the operation is performed in OkLCh. The hue twists in these RGB space, so the idea is to just accept that prefection is not achievable (when not working directly in OkLCh), and minimize lightness changes and hue changes as much as possible.

In short, like Scale LH, if you adjust the color in the RGB space, which this approach does, there will be hue shift at certain places. I realize now, the idea is to get as close as you can and smooth it out.

I assume this approach could be used directly on OkLCh, though I imagine it would be more complex that raytracing a perfectly aligned cube.

This approach uses raytracing to find the intersection of the chroma line on the RGB cube. Obviously, due to the way OkLCh twists the hues in the RGB space, this point will not be perfect, so we back off some degree on each attempt to get a better line. At the end, we clip the color.

I've provided two approaches, one that takes 2X passes and one that takes 3X passes. The 3X one obviously reduces hue shift the most, but 2X seems pretty acceptable as well. in use They both seem to produce smooth gradients (in the cases I tested). Lightness doesn't deviate very much, but hue can deviate in some problematic areas, the same as it does in Scale LH. A lot of time our hue deviates less, but we do deviate a little more in some cases, but with the lightness holding better, they seem far less noticeable.

I'm sure this approach can be tweaked and optimized further.

I may revisit my other approach and take what I learned here, but this approach seems pretty decent.

This is ready for review.

PR #453, opened 2024-02-25
Allow getColor to work with an array of color types

lloydk:
Added the ability to pass an array of color types to getColor.

This allows you to do:
```javascript
let [c1, c2] = getColor(["red", "blue"])
```
instead of the more repetitive
```javascript
let c1 = getColor("red")
let c2 = getColor("blue")
```

As discussed in #451

### DIFF
--- a/src/getColor.js
+++ b/src/getColor.js
@@ -4,8 +4,8 @@ import parse from "./parse.js";
 
 /**
  * Resolves a color reference (object or string) to a plain color object
- * @param {Color | {space, coords, alpha} | string} color
- * @returns {{space, coords, alpha}}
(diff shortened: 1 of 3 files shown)


src/getColor.js
```diff
@@ -8,6 +8,10 @@ import parse from "./parse.js";
  * @returns {{space, coords, alpha}}
  */
 export default function getColor (color) {
+	if (Array.isArray(color)) {
```
jgerigmeyer, line 11: Update the JSDoc comments above to reflect this new arg type.

jgerigmeyer (approved):

LeaVerou (approved):

PR #454, opened 2024-02-26
Minor fine tune to reduce hue shift in red and blue region

facelessuser:
More experimentation shows that the threshold for backoff doesn't seem to be needed. Also, we can approach a little closer on the last iteration. The last part seems to reduce hue shift a bit more.


LeaVerou (approved):
FWIW I think this would have been fine to commit directly — it's an app, not Color.js, and it's your own GM method :)

facelessuser:
That's fair, just didn't want to overstep.

LeaVerou:
Yup, I get it — better safe than sorry. Just mentioned it for future calibration :)

PR #456, opened 2024-02-27
Audit function parameter and return types (1/2)

lloydk:
I did an audit of all of the functions to ensure they have the correct types and that all color types can be passed as parameters.

There were a fair number of changes so I'm breaking up the changes into two PRs, this is part 1 of 2.

PR #457, opened 2024-02-27
Audit function parameter and return types (2/2)

lloydk:
Ensure that functions in the following modules have the correct types and ensure that all color types can be passed as parameters:

- inGamut
- luminance
- set
- setAll
- space
- toGamut

### space.js
I wasn't sure about the types for the `to` and `from` methods on `Colorspace`. If we allow any color type and call `getColor` in those methods then we introduce a cyclic dependency between the `getColor` module and the `space` module. If we're ok having the cyclic dependency and want to allow any color type I can make the appropriate changes.

### DIFF
--- a/src/luminance.js
+++ b/src/luminance.js
@@ -6,10 +6,12 @@ import set from "./set.js";
 import xyz_d65 from "./spaces/xyz-d65.js";
 
 export function getLuminance (color) {
+	// Assume getColor() is called on color in get()
 	return get(color, [xyz_d65, "y"]);
 }
 
 export function setLuminance (color, value) {
+	// Assume getColor() is called on color in set()
 	set(color, [xyz_d65, "y"], value);
 }
 
--- a/types/src/set.d.ts
+++ b/types/src/set.d.ts
@@ -1,12 +1,12 @@
-import Color, { ColorTypes } from "./color.js";
+import { ColorTypes, PlainColorObject } from "./color.js";
 import { Ref } from "./space.js";
 
 export default function set (
 	color: ColorTypes,
 	prop: Ref,
 	value: number | ((coord: number) => number)
-): Color;
+): PlainColorObject;
 export default function set (
 	color: ColorTypes,
 	props: Record<string, number | ((coord: number) => number)>
-): Color;
+): PlainColorObject;
--- a/types/test/toGamut.ts
+++ b/types/test/toGamut.ts
@@ -11,3 +11,9 @@ toGamut(new Color("red")); // $ExpectType PlainColorObject
 toGamut(new Color("red"), { method: "clip", space: "srgb" }); // $ExpectType PlainColorObject
 toGamut(new Color("red"), { method: "clip", space: sRGB }); // $ExpectType PlainColorObject
 toGamut(new Color("red"), "srgb"); // $ExpectType PlainColorObject
+
+
+new Color("red").toGamut(); // $ExpectType Color
+new Color("red").toGamut({ method: "clip", space: "srgb" }); // $ExpectType Color
+new Color("red").toGamut({ method: "clip", space: sRGB }); // $ExpectType Color
+new Color("red").toGamut("srgb"); // $ExpectType Color
--- a/src/inGamut.js
+++ b/src/inGamut.js
@@ -7,8 +7,13 @@ const ε = .000075;
  * Check if a color is in gamut of either its own or another color space
  * @return {Boolean} Is the color in gamut?
  */
-export default function inGamut (color, space = color.space, {epsilon = ε} = {}) {
+export default function inGamut (color, space, {epsilon = ε} = {}) {
 	color = getColor(color);
+
+	if (!space) {
+		space = color.space;
+	}
+
 	space = ColorSpace.get(space);
 	let coords = color.coords;
 
--- a/src/setAll.js
+++ b/src/setAll.js
@@ -1,6 +1,9 @@
 import ColorSpace from "./space.js";
+import getColor from "./getColor.js";
 
 export default function setAll (color, space, coords) {
+	color = getColor(color);
+
 	space = ColorSpace.get(space);
 	color.coords = space.to(color.space, coords);
 	return color;
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -65,14 +65,19 @@ export default function toGamut (
 	color,
 	{
 		method = defaults.gamut_mapping,
-		space = color.space,
+		space = undefined,
 		deltaEMethod = "",
 		jnd = 2,
 		blackWhiteClamp = {},
 	} = {},
 ) {
+	color = getColor(color);
+
 	if (util.isString(arguments[1])) {
 		space = arguments[1];
+	}
+	else if (!space) {
+		space = color.space;
 	}
 
 	space = ColorSpace.get(space);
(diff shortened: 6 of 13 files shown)


lloydk:
I wasn't 100% convinced that `getColor` should be called in `setAll` but that's what I've gone with for now.

types/test/toGamut.ts
```diff
@@ -11,3 +11,9 @@ toGamut(new Color("red")); // $ExpectType PlainColorObject
 toGamut(new Color("red"), { method: "clip", space: "srgb" }); // $ExpectType PlainColorObject
 toGamut(new Color("red"), { method: "clip", space: sRGB }); // $ExpectType PlainColorObject
 toGamut(new Color("red"), "srgb"); // $ExpectType PlainColorObject
```
jgerigmeyer, line 13: All of these 4 examples above actually return a `Color`.

jgerigmeyer (changes requested):
> I wasn't sure about the types for the `to` and `from` methods on `Colorspace`.

If the circular dependency doesn't hurt anything in practice, I'd lean toward calling `getColor` there to be consistent.

> I wasn't 100% convinced that `getColor` should be called in `setAll` but that's what I've gone with for now.

I'm not opposed to calling `getColor` in set/setAll etc, but currently there are a number of types that are incorrect, some of them a bit tangential to this PR.

First, the type of `getColor` itself is wrong -- it claims to return a `PlainColorObject` (or `PlainColorObject[]`), where in reality if it is passed a `Color` object, it returns a `Color` object. That means that every fn which calls `getColor` and then returns the response (e.g. `set`, `setAll`, and `toGamut`) is also wrong -- they all claim to return a `PlainColorObject` but often return a `Color`.

Honestly, this (that is, keeping the types up-to-date with the codebase) is going to be an ongoing headache unless we're willing to start re-writing the code itself in TypeScript. That's something I'd be willing to consider taking on (potentially with help from @MysteryBlokHed, if he's willing), but I vaguely recall that @LeaVerou was opposed to that direction -- Lea, can you confirm?

types/src/set.d.ts
```diff
@@ -1,12 +1,12 @@
...
 	color: ColorTypes,
 	prop: Ref,
 	value: number | ((coord: number) => number)
-): Color;
+): PlainColorObject;
 export default function set (
 	color: ColorTypes,
 	props: Record<string, number | ((coord: number) => number)>
-): Color;
+): PlainColorObject;
```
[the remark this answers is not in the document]
MysteryBlokHed: The line here in the source calls `getColor`, which does return `PlainColorObject` according to the types:

<https://github.com/color-js/color.js/blob/599341ff01b0a76c8df479d937ec5397c3d454db/src/set.js#L9-L10>

Maybe the function that's actually part of the `Color` class returns a `Color` object, but this definition returns `PlainColorObject`? There are other functions that do that too.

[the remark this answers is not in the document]
jgerigmeyer: Right, the type definition for `getColor` is incorrect -- in reality it returns a `Color` object any time it is passed a `Color` object as the first arg.

lloydk:
> I'm not opposed to calling `getColor` in set/setAll etc, but currently there are a number of types that are incorrect, some of them a bit tangential to this PR.
> 
> First, the type of `getColor` itself is wrong -- it claims to return a `PlainColorObject` (or `PlainColorObject[]`), where in reality if it is passed a `Color` object, it returns a `Color` object. That means that every fn which calls `getColor` and then returns the response (e.g. `set`, `setAll`, and `toGamut`) is also wrong -- they all claim to return a `PlainColorObject` but often return a `Color`.
> 

Keep in mind that `PlainColorObject` is an interface and the `Color` type implements the `PlainColorObject` interface.

The types for the `getColor` function and other functions that return `PlainColorObject` simply guarantee that the object that is returned conforms to the `PlainColorObject` interface (space, coords and alpha properties). The object may be a `Color`, object literal, or some other class that has `space`, `coords` and `alpha` properties.

```typescript
let x: PlainColorObject = getColor("red"); // returns an object literal
let y: PlainColorObject = getColor(new Color("red")); // returns a Color object
let z: Color = getColor(new Color("red")); // type error
```

If we want to make additional guarantees about returning the same object if it conforms to the `PlainColorObject` interface or that we may modify an object in place we can do that in the JSDoc comments.

None of the functions exported in 'colorjs.io/fn' should have types that reference the `Color` type because the `Color` class isn't available in `colorjs.io/fn`.

LeaVerou:
> Honestly, this (that is, keeping the types up-to-date with the codebase) is going to be an ongoing headache unless we're willing to start re-writing the code itself in TypeScript. That's something I'd be willing to consider taking on (potentially with help from @MysteryBlokHed, if he's willing), but I vaguely recall that @LeaVerou was opposed to that direction -- Lea, can you confirm?

I do have some reservations, one of which being that there are many folks who are proficient in color science but only dabble with JS, and a TS codebase is hostile to them. Whereas having the types separately means we can still accept contributions from folks who are not comfortable with TS, and adjust the typings ourselves after.

I’m willing to examine it , but could you elaborate a bit on how TS would help with this issue? I obviously get how it would help us keep types up to date, but not how it would help with the return type issue.

types/src/set.d.ts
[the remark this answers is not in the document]
lloydk: > Maybe the function that's actually part of the `Color` class returns a `Color` object, but this definition returns `PlainColorObject`? There are other functions that do that too.

That's correct.

src/luminance.js
```diff
@@ -6,10 +6,12 @@ import set from "./set.js";
 import xyz_d65 from "./spaces/xyz-d65.js";
 
 export function getLuminance (color) {
+	// Assume getColor() is called on color in get()
```
jgerigmeyer, line 9: I don't think this is true yet in this PR? I adjusted in https://github.com/lloydk/color.js/pull/1

jgerigmeyer (commented):
> Keep in mind that `PlainColorObject` is an interface and the `Color` type implements the `PlainColorObject` interface.
>
> None of the functions exported in 'colorjs.io/fn' should have types that reference the `Color` type because the `Color` class isn't available in `colorjs.io/fn`.

@lloydk Thanks -- those are both really helpful points, and I was missing the forest for the trees. 🤦 

I've pushed some proposed changes/additions for you to review in https://github.com/lloydk/color.js/pull/1, mostly to see what it looks like if we enforce more consistent `Color` object return types from the Color namespace methods.

> could you elaborate a bit on how TS would help with this issue? I obviously get how it would help us keep types up to date, but not how it would help with the return type issue.

@LeaVerou I'm distracting from the purpose of this PR, so I'll open a new issue to consider this question. 😄 

> My understanding is that TS does quite poorly with very dynamic APIs where functions are computed from other functions etc.

Yes, that can be true in some cases. I may try it out on a few methods and see what it looks like.

src/luminance.js
lloydk: It was done in #456. Should I rebase this PR?

jgerigmeyer: Oh, yeah I missed that, good call. Sure, could you rebase?

lloydk:
> I'm fine merging this as-is, then I have a follow-up PR I can open with a few more proposed changes.

Did you still want me to review https://github.com/lloydk/color.js/pull/1 or should I wait for the follow-up PR?

jgerigmeyer:
> > I'm fine merging this as-is, then I have a follow-up PR I can open with a few more proposed changes.
> 
> Did you still want me to review [lloydk#1](https://github.com/lloydk/color.js/pull/1) or should I wait for the follow-up PR?

Probably easier for me to open the PR here and review it then -- thanks!

PR #458, opened 2024-02-27
[Gamut Mapping App] Always display delta

jamesnw:
See conversation at https://github.com/color-js/color.js/pull/438#discussion_r1489822585. 

The result of a gamut mapping algorithm may change in-gamut colors, so it is useful to see the deltas in those cases as well.

### DIFF
--- a/apps/gamut-mapping/map-color.js
+++ b/apps/gamut-mapping/map-color.js
@@ -180,7 +180,7 @@ export default {
 					</dt>
 					<dd>
 						<css-color swatch="large" :color="mapped[method].color"></css-color>
-						<dl class="deltas" v-if="!color.inGamut('p3')">
+						<dl class="deltas">
 							<div v-for="(delta, c) of mapped[method].deltas" :class="'delta-' + c.toLowerCase()">
 								<dt>Δ{{ c }}</dt>
 								<dd :class="{


facelessuser:
I like this.

facelessuser (approved):

jgerigmeyer (approved):

LeaVerou:
> The result of a gamut mapping algorithm may change in-gamut colors

As I mentioned in the other PR, gamut mapping should never change in-gamut colors. When this happens, it's a bug.

jgerigmeyer:
> > The result of a gamut mapping algorithm may change in-gamut colors
> 
> As I mentioned in the other PR, gamut mapping should never change in-gamut colors. When this happens, it's a bug.

@LeaVerou Even if that's correct (and [this comment](https://github.com/color-js/color.js/pull/438#discussion_r1493853210) makes it sound like not everyone agrees with that), isn't it helpful to surface it?

facelessuser:
> @LeaVerou Even if that's correct (and [this comment](https://github.com/color-js/color.js/pull/438#discussion_r1493853210) makes it sound like not everyone agrees with that), isn't it helpful to surface it?

That was my thinking. It isn't that gamut mapping should alter already in gamuts, but it lets you see whether they are.

facelessuser:
> we should show deltas iff the color is different than the original (whether or not they are in gamut).

I think that would be a fine alternative. It solves the original concern but also filters out the useless noise.

jamesnw:
I like that as an alternative, and @jgerigmeyer has offered to implement. Thanks!

PR #464, opened 2024-02-29
Allow hwb to accept numbers

### LINKED ISSUE #368: [spaces] HWB parsing requires percentages, does not accept `none`
Compare [`hwb(90 20% 40%)`](https://colorjs.io/apps/convert/?color=hwb(90%2020%25%2040%25)&precision=6) with `hwb(180 20 40)` which is rejected until you make it `hwb(180 20% 40%)`.

See [the syntax of the `hwb()` function](https://drafts.csswg.org/css-color-4/#funcdef-hwb)

`none` is also not recognized, e.g `hwb(180 20% none)`

lloydk:
Fixes #368

The issue with parsing `none` appears to have been fixed in a prior commit.

### DIFF
--- a/src/spaces/hwb.js
+++ b/src/spaces/hwb.js
@@ -52,7 +52,7 @@ export default new ColorSpace({
 
 	formats: {
 		"hwb": {
-			coords: ["<number> | <angle>", "<percentage>", "<percentage>"],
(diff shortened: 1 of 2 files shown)


svgeesus (approved):
Thanks!

PR #327, opened 2023-09-14
[types] Re-export some types at the top level

MysteryBlokHed:
Closes #326

Re-exports some potentially desirable types at the top level instead of requiring them to be manually imported from their individual files.

I basically just went through the files in `types/src` and added `export`s in the main `index.d.ts` file for types and interfaces that might be wanted.

LMK if there are any other types that I missed 😊

PR #466, opened 2024-02-29
Export types at top level of /fn entrypoint.

jgerigmeyer:
This is a missing corollary of https://github.com/color-js/color.js/pull/327


MysteryBlokHed (approved):
LGTM

Issue #434, opened 2024-02-13
[chore] Separate website/elements/apps from library?


LeaVerou:
The current structure has served us well but we're at the point where we need to separate the two.

In terms of tests, presumably the JS tests will stay with the library, while the HTML tests (old and new) will live with the website.

Both will still deploy to the web, I’m thinking `dev.colorjs.io` for the library repo.

LeaVerou:
Where would the API docs live? Presumably they need access to the source, but perhaps that can be done by installing `color.js` as an npm package on the website repo and pointing typedoc to that?

MysteryBlokHed:
> Where would the API docs live? Presumably they need access to the source, but perhaps that can be done by installing `color.js` as an npm package on the website repo and pointing typedoc to that?

That should work if we want to keep all website stuff together. Alternatively there could also be a Git submodule that links back to this repo, but that might be overcomplicating things.

LeaVerou:
Yeah, my opinion on this is that submodules are plague and must be avoided at all costs 😛

LeaVerou:
I’ve been thinking about this and I think it makes sense to also separate apps and elements into separate repos / separate npm packages / separate subdomains.

Especially as I plan a bit of a revamp of these, what do you think about separating them?
We could create an npm namespace and publish them as `@colorjs/apps` and `@colorjs/elements`. 
Later down the line we may even want to separate elements into separate packages, but let’s cross that bridge when we get to it.

LeaVerou:
As an update, I have split elements and apps into their own repos and websites (https://apps.colorjs.io and https://elements.colorjs.io respectively). The website will be harder to split out though…

LeaVerou:
Ok, thinking about this some more, I have a plan:
- We split into two repos: `core` (which we can deploy, e.g. to `dev.colorjs.io`) and `colorjs.io` (names TBB, suggestions welcome)
- `core` will contain `test`, `tests`, `src`, `types`, `scripts`, `CONTRIBUTING.md`, `.editorconfig`, ESLint config, tsconfig, LICENSE, README.md
- `colorjs.io` will contain: Notebook, 11ty files, assets, `get/`
- Unsure: `benchmarks`, Typedoc, 

Thoughts?

MysteryBlokHed:
[...]

PR #467, opened 2024-02-29
Document how to serve in development, and add `--serve` to `watch:html`.

jamesnw:
The current npm scripts for development don't include a way to serve the built files. Because 11ty contains a dev server, we can use that. 

This adds the `--serve` flag to `npm run watch:html` (and by extension `npm run watch`) so that developers don't need to separately serve the content.

If this is disruptive to existing workflows, we could add an alternate `npm run dev` that would watch css, js and html, and also serve. Alternatively, we could document how to serve with `npx http-server`, but since the functionality is already available through 11ty, this seems like a higher burden on contributors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
 		"build:space-accessors": "node ./scripts/generate-space-accessor-types.js",
 		"build": "run-s build:html build:css build:js build:js:legacy build:space-accessors",
 		"watch:css": "npx postcss \"**/*.postcss\" --base . --dir . --ext .css --config postcss.config.cjs --watch",
-		"watch:html": "npx @11ty/eleventy --config=.eleventy.cjs --watch",
+		"watch:html": "npx @11ty/eleventy --config=.eleventy.cjs --watch --serve",
 		"watch:js": "rollup -c --watch",
 		"watch": "run-p watch:*",
 		"prepack": "npm run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
 		"build:space-accessors": "node ./scripts/generate-space-accessor-types.js",
 		"build": "run-s build:html build:css build:js build:js:legacy build:space-accessors",
 		"watch:css": "npx postcss \"**/*.postcss\" --base . --dir . --ext .css --config postcss.config.cjs --watch",
-		"watch:html": "npx @11ty/eleventy --config=.eleventy.cjs --watch",
+		"watch:html": "npx @11ty/eleventy --config=.eleventy.cjs --watch --serve",
```
jgerigmeyer, line 62: `--serve` implies `--watch`

https://www.11ty.dev/docs/usage/#re-run-eleventy-when-you-save

jgerigmeyer, line 62: `--serve` implies `--watch`

https://www.11ty.dev/docs/usage/#re-run-eleventy-when-you-save

```suggestion
		"watch:html": "npx @11ty/eleventy --config=.eleventy.cjs --serve",
```

jgerigmeyer (approved):

LeaVerou:
Would this cause issues for #434 ?

jgerigmeyer:
> Would this cause issues for #434 ?

I don't think so, no -- this change just means the server will start along with the other `watch:*` tasks.

#434 is a good idea and will require changes to various aspects of the dev process, so this PR may be temporary.

LeaVerou:
Ok, fine by me!

PR #469, opened 2024-03-02
Add Okhsl and Okhsv

### LINKED ISSUE #312: OKHSV and OKHSL
It would be nice to have support for OKHSV and OKHSL, I find OKLCH challenging to work for color picking.

I have written basic support for OKHSV and OKHSL color spaces within color.js, you can see it here:

https://github.com/mattdesl/workshop-generative-color/blob/f5a26247fac7e6dc617f05fd23886a3bba6b5d4b/lib/color.js#L10-L81

and using the color conversion tools here:

https://github.com/mattdesl/workshop-generative-color/blob/f5a26247fac7e6dc617f05fd23886a3bba6b5d4b/lib/oklab.js

These conversions are only slightly modified and stripped down from Björn's color picker code, see here:

https://github.com/bottosson/bottosson.github.io/blob/master/misc/colorpicker/colorconversion.js (MIT license, as far as I can see)

facelessuser:
Resolves #312


facelessuser:
This will add both Okhsl and Okhsv.  It is implemented in a way that you could easily swap out coefficients and matrices and create Okhsl and Okhsv in other gamuts. Not that we want to do that ever, but that is why somethings are done the way they are.

I am currently implementing it where lightness and saturation is in the range of 0 - 1 as the original author had done, but if it is decided we want 0 - 100, we can make that change.

I'll leave this in a draft until I've implemented both and added testing.

facelessuser:
It should also be noted that Okhsl and Okhsv are only accurate to the 32 bit precision. So with the much more accurate 64 precision Oklab matrices, you get:

```js
> new Color('red').to('okhsl').coords
[ 29.233880279627897, 1.0000000995016394, 0.5680846563197034 ]
> new Color('blue').to('okhsl').coords
[ 264.05202261636987, 1.0000000005848084, 0.3665653391870817 ]
> new Color('green').to('okhsl').coords
[ 142.4953450414439, 0.9999999797952728, 0.4437098579755183 ]
```
Okhsl and Okhsv also are an approximation, but the actual bounds of the model will cut out some sRGB colors and include some that are technically out of gamut. For this reason, I'm opting to set its gamut as itself. For practicality, it is suggested that people choose to gamut map in sRGB, the closest gamut and the one it aimed to target. We could set its gamut to sRGB, but then colors within the model near the edge will complain they are out of gamut.  This is just my opinion and I'll let others decide what is best in this library.

facelessuser:
I believe this is ready for review.

svgeesus:
> For this reason, I'm opting to set its gamut as itself. For practicality, it is suggested that people choose to gamut map in sRGB, the closest gamut and the one it aimed to target. We could set its gamut to sRGB, but then colors within the model near the edge will complain they are out of gamut.

That sounds like a good plan to me.

Issue #437, opened 2024-02-14
New method: `deltas()`


LeaVerou:
When working on the [gamut mapping playground](https://colorjs.io/apps/gamut-mapping) recently, I needed to calculate deltas in OKLCh (differences per component) and ended up doing so by hand.

This seems like a more generally useful function. We have `distance()` and a bunch of `deltaE()` methods, but nothing that is done per-component.

What do you think?

svgeesus:
I realized that we claim to calculate ΔH but are in fact calculating Δh (the angle, not the length of the arc). What this means is that we over-emphasize the difference for near-neutrals and under-emphasize it for high chroma colors.

But yes a general method that gives the set of deltas is probably better and likely faster if the user calculates each one separately.

LeaVerou:
> I realized that we claim to calculate ΔH but are in fact calculating Δh (the angle, not the length of the arc). What this means is that we over-emphasize the difference for near-neutrals and under-emphasize it for high chroma colors.
> 
> But yes a general method that gives the set of deltas is probably better and likely faster if the user calculates each one separately.

Oh that's really interesting. That would explain a lot in the GMA app.

svgeesus:
It also strikes me that, for Οklab, calculating these gets you most the way to DeltaE OK and so there is a risk that these get calculated twice, for the common case where you want ΔΕ ΔL ΔC and ΔH. So I wonder if deltas should return an array of four items rather than 3.

But then, this is not the case at all for complex formulae like ΔE2000 so perhaps not.

Note too the [optimization](https://github.com/color-js/color.js/blob/c3315bcb2dae9e99c758bb178903ce6e6cd9a55a/src/deltaE/deltaECMC.js#L62) in ΔECMC:

```js
let H2 = (Δa ** 2) + (Δb ** 2) - (ΔC ** 2);
```

and then ΔH is `Math.sqrt(H2)` and also

ΔE = Math.sqrt((ΔL ** 2) + (Δa ** 2) + (Δb ** 2))
     = Math.sqrt( (ΔL ** 2) + (ΔC ** 2) + (ΔH** 2))
      = Math.sqrt( (ΔL ** 2) + (ΔC ** 2) + H2)

which can be further optimized by keeping the values of those squares around rather than calculating them twice.

PR #471, opened 2024-03-05
[apps/gamut-mapping] Add gradient tool view

jamesnw:
This adds a way to explore and compare the gamut mapping algorithms across a gradient. 

Essentially, it uses the Color.js steps function to create steps between 2 colors in a space you pick, and then applies the gamut mapping method to each step.

It currently shows the speed of calculating each method, as well as a bar that shows when the gradient is out of the srgb, p3 and rec2020 gamuts.

image

I'm open to feature ideas. Some I'd like to add are sorting or showing a subset of methods, and show different distance calculations (DeltaE, L, H) per step, especially as those are added in #437.

Remaining blocking tasks

- [ ] Get colors from URL
- [ ] Handle invalid color inputs

### DIFF
--- a/apps/gamut-mapping/gradients.css
+++ b/apps/gamut-mapping/gradients.css
@@ -1,0 +1,36 @@
+:root {
+	--font-mono: Consolas, Inconsolata, Monaco, monospace;
+}
+
+body {
+	font: 100%/1.5 system-ui;
+	max-width: 84em;
+	margin: 1em auto;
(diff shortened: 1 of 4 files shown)


facelessuser:
I think this is a great addition.

svgeesus (approved):
Nice!

jamesnw:
I also added a view to compare timing info.

image

facelessuser (approved):

PR #472, opened 2024-03-05
fix: add optional chaining in the event `CSS` is null

michael-siek:
Adding this in the event that `CSS` global is null this doesn't throw an error. There was a user that went thru this code path and `CSS` was null and threw an error.

### DIFF
--- a/src/display.js
+++ b/src/display.js
@@ -13,7 +13,7 @@ defaults.display_space = sRGB;
 
 let supportsNone;
 
-if (typeof CSS !== "undefined" && CSS.supports) {
+if (typeof CSS !== "undefined" && CSS?.supports) {
 	// Find widest supported color space for CSS
 	for (let space of [Lab, REC2020, P3]) {
 		let coords = space.getMinCoords();
--- a/src/display.js
+++ b/src/display.js
@@ -42,7 +42,7 @@ if (typeof CSS !== "undefined" && CSS.supports) {
 export default function display (color, {space = defaults.display_space, ...options} = {}) {
 	let ret = serialize(color, options);
 
-	if (typeof CSS === "undefined" || CSS.supports("color", ret) || !defaults.display_space) {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hey, thank you so much for working on this!
Just curious, in what situation would `CSS` be null?

LeaVerou:
I wonder if a better solution might be to do `globalThis.CSS && CSS.supports` for the check

michael-siek:
Hey @LeaVerou is there any changes that you would like for me to do to this PR to solve the issue?

michael-siek:
Hey @LeaVerou any update on this issue?

LeaVerou:
Hi there! Sorry for the delay. I’m just not sure in what case `CSS` would be present, but `null`. If it was overwritten by the environment somehow, that’s on the environment, we can’t start guarding against that for every global.

If it's something that happens naturally in some case or in some environment, sure, I'd be happy to merge this, but when I asked the response was that you didn’t know why `CSS` was null, it just was. If some part of your code (or some library you're using) is overwriting `CSS`, Color.js is not the place to fix this. It's a slippery slope, we can't start adding optional chaining to every single global just in case it was overwritten.

straker:
@LeaVerou We did some digging and found that users are [following this guide for jest-preset-angular](https://thymikee.github.io/jest-preset-angular/docs/getting-started/installation/#global-mocks ) in which they are being told to override `CSS` with `null`.

svgeesus:
> @LeaVerou We did some digging and found that users are [following this guide for jest-preset-angular](https://thymikee.github.io/jest-preset-angular/docs/getting-started/installation/#global-mocks) in which they are being told to override `CSS` with `null`.

That very much sounds like it should be fixed at the point which is causing this problem, which I see has been raised as an issue.

LeaVerou:
I think I'm tending to agree. @straker as a workaround you can check if it's null and set it to undefined if so.

Issue #473, opened 2024-03-06
Carrying forward analogous components in color interpolation


svgeesus:
@svgeesus So this seems like a surprising revelation. In the past, the resolution of undefined values seemed to be handled during the actual interpolation, not before. Currently, this is exactly how Color.js handles things. Below, we do not get 270deg, but carryforward is also not implemented.

```js
> new Color("hsl(90deg 50% 50%)").mix('hsl(none 50% 50%)', {space: 'hsl', hue: 'longer'});
Color {
  space: <ref *1> ColorSpace {
    id: 'hsl',
    name: 'HSL',
    base: RGBColorSpace {
      id: 'srgb',
      name: 'sRGB',
      base: [RGBColorSpace],
      aliases: undefined,
      fromBase: [Function: fromBase],
      toBase: [Function: toBase],
      coords: [Object],
      white: [Array],
      formats: [Object],
      referred: 'display',
      path: [Array]
    },
    aliases: undefined,
    fromBase: [Function: fromBase],
    toBase: [Function: toBase],
    coords: { h: [Object], s: [Object], l: [Object] },
    white: [ 0.9504559270516716, 1, 1.0890577507598784 ],
    formats: { hsl: [Object], hsla: [Object] },
    referred: undefined,
    path: [ [ColorSpace], [RGBColorSpace], [RGBColorSpace], [Circular *1] ]
  },
  coords: [ 90, 50, 50 ],
  alpha: 1
}
```

But now, the act of carrying forward, which I was under the impression carried forward undefined values only, is also now resolving undefined values with the context of colors before and after it and no longer waiting for interpolation which seemed to be what was implied here?

> If a color with a carried forward [missing component](https://drafts.csswg.org/css-color/#missing-color-component) is interpolated with another color which is not missing that component, the missing component is treated as having the other color’s component value.

Has Color.js always been wrong, or does carryforward fundamentally change this? I have to admit, I can see why browsers may be doing this differently than others. I find it unintuitive that carryforward would resolve the undefined values in this manner.

_Originally posted by @facelessuser in https://github.com/w3c/csswg-drafts/issues/9224#issuecomment-1743834039_

facelessuser:
I'll have a fix-up later today. I had to make changes my lib conform when explicitly requested to do so. I have it working in Color.js locally:

```js
> new Color("hsl(90deg 50% 50%)").mix('hsl(none 50% 50%)', {space: 'hsl', hue: 'longer'}).coords
[...]

PR #474, opened 2024-03-07
Make longer and undefined/same hues have parity with CSS spec

facelessuser:
Ref #473


facelessuser:
This is ready for review

LeaVerou (approved):

Issue #475, opened 2024-03-09
Gray color is not correctly cast to a string


star-over:
I noticed a problem when trying to use the API in the developer tools. 
Specifically, when I tried to convert the color 'gray' to HSL and then to a string, the result displayed 'none'. Other color names do not have this result.

```
(new Color("gray")).to("hsl").toString()
>> 'hsl(none 0% 50.196%)'
```

<img width="446" alt="image" src="https://github.com/color-js/color.js/assets/803639/24b66b9e-9610-4b83-b51f-9fbd453d55f5">

facelessuser:
What hue would you suggest gray has? The reason why you get an undefined hue (`none`) is because the color is achromatic and has no hue.

star-over:
I suggest use `0` instead `none` when color stringify. So, I expect `hsl(0 0% 50.196%)` but not a `hsl(none 0% 50.196%)`

facelessuser:
`none` is now supported in the CSS spec and has been implemented in a number of browsers. Gray doesn't actually have a red (0deg) hue, but regardless of technicalities of what is more correct, I can imagine some people desiring a way to serialize without `none`. There was an option to do so for some amount of time, but it was removed in https://github.com/color-js/color.js/commit/e28cd6a48922c525188f779506f6c204e4aecf40.

Below does look like a bug to me though. Undefined values should be handled before conversion, but that does not appear to be happening currently.

```
new Color("hsl(none 0% 50.196%)").to("srgb").toString()
>>> 'rgb(none none none)'
```

star-over:
I just fixed problem for a while with a `.toString().replaceAll("none", "0")` after any stringify.

facelessuser:
I personally think an option to allow omitting `none` in serialization seems reasonable, but the`none` conversion issue should definitely get fixed.

facelessuser:
I've created a PR (#476) to adjust the bad "none" handling across the library.

@svgeesus and @LeaVerou would need to make a decision on whether it is desired to expose conditional "none" serialization. If they decided that they wanted to move forward with this, I'd be happy to make the PR.

facelessuser:
None bug has been fixed on main.  A new issue is opened that, when implemented, could solve the serialization request: https://github.com/color-js/color.js/issues/479. I will resolve this as a duplicate of that issue. I advise following that issue to stay on top of progress.

PR #476, opened 2024-03-10
Consistently handle none/NaN across library

facelessuser:
This fixes bad conversion due to "none" usage and other "none" related issues.
Ref #475

### DIFF
--- a/src/contrast/APCA.js
+++ b/src/contrast/APCA.js
@@ -4,6 +4,7 @@
 
 import getColor from "../getColor.js";
 import to from "../to.js";
+import {isNone} from "../util.js";
 
 // exponents
 const normBG = 0.56;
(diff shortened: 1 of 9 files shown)


lloydk (changes requested):
LGTM except for the APCA contrast method which doesn't handle colors with `none`.
```
> new Color("red").contrast(new Color("rgb(none 100% 100%)"), "apca");
> NaN
```

facelessuser:
Yeah, I was fixing existing handling, I didn't try to identify new broken cases 🙂

facelessuser:
I'll address it.

facelessuser:
Done

```js
> new Color("red").contrast(new Color("rgb(none 100% 100%)"), "apca");
-53.91362069001541
```

src/contrast/APCA.js
```diff
@@ -54,11 +55,15 @@ export default function contrastAPCA (background, foreground) {
...
-	[R, G, B] = foreground.coords;
+	[R, G, B] = foreground.coords.map(c => {
+		return isNone(c) ? 0 : c;
+	});
 	let lumTxt = linearize(R) * 0.2126729 + linearize(G) * 0.7151522 + linearize(B) * 0.0721750;
 
 	background = to(background, "srgb");
-	[R, G, B] = background.coords;
+	[R, G, B] = background.coords.map(c => {
+		return isNone(c) ? 0 : c;
```
LeaVerou, line 65: Not a blocker, but I think we need something higher level than this, i.e. a function to drop `none`s. If that function does not mutate objects but returns new ones, it will also fix the serialization problem. Can you open a new issue for the API design discussion?

facelessuser: > Not a blocker, but I think we need something higher level than this

I generally agree.

> Can you open a new issue for the API design discussion?

Yep. Issue created: https://github.com/color-js/color.js/issues/479

PR #480, opened 2024-03-12
Port color modification tests

### LINKED ISSUE #395: [chore] Port Color modification tests to the new test format

lloydk:
Closes #395

### DIFF
--- a/test/modifications.js
+++ b/test/modifications.js
@@ -4,9 +4,6 @@ import { check } from "./util.mjs";
 export default {
 	name: "Color modification tests",
 	description: "These tests modify one or more coordinates and check the result.",
-	run (func) {
-		return func();
-	},
 	check: check.deep(check.proximity({epsilon: .005})),
 	tests: [
 		{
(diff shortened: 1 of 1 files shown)


test/modifications.js
```diff
@@ -0,0 +1,100 @@
+import Color from "../src/index.js";
+import { check } from "./util.mjs";
+
+export default {
+	name: "Color modification tests",
+	description: "These tests modify one or more coordinates and check the result.",
+	run (func) {
+		return func();
```
LeaVerou, line 8: That is never needed, you can just have a `run` on each individual test. :)

LeaVerou (approved):
Thanks so much for working on this, you rock!

My only comment is that you don’t actually need `run` there at all, you just need to rename `args` to `run` and it should do what you intend here (and if it doesn't that's a bug to fix). 

There is probably a way to make these a bit less repetitive, but whatevs, I'm just very glad to have them ported! 😄 

Thanks again!

test/modifications.js
lloydk: Fixed, that's way better.

PR #482, opened 2024-03-12
Fix/skip failing conversion tests

lloydk:
Fixes a few conversion tests that were failing and skips all failing ICtCp tests.

Depends on #481 to skip the failing tests.

### DIFF
--- a/test/conversions.js
+++ b/test/conversions.js
@@ -258,7 +258,7 @@ const tests = {
 					expect: [0.22206525, -0.000161, -0.000117],
 				},
 				{
-					args: "color(jzazbz 0.22206525 -0.000161 -0.000117)",
+					args: ["color(jzazbz 0.22206525 -0.000161 -0.000117)", "srgb"],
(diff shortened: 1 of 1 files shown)


facelessuser:
I don't think the ICtCp tests should be skipped, they should be corrected. The tests are bad, not ICtCp or its ability to round trip.

lloydk:
> I don't think the ICtCp tests should be skipped, they should be corrected. The tests are bad, not ICtCp or its ability to round trip.

Wasn't sure if it was the tests or the conversion routines. Fixed.

LeaVerou (approved):
LGTM but I'd like @svgeesus to take a look too

svgeesus:
> I don't think the ICtCp tests should be skipped, they should be corrected. The tests are bad, not ICtCp or its ability to round trip.

I agree.

I haven't yet found a source of known-goot ICtCp conversions of of deltaITP values for given color pairs.

svgeesus (approved):

PR #484, opened 2024-03-13
Port hooks tests

lloydk:


test/hooks.js
```diff
@@ -0,0 +1,28 @@
+import Color from "../src/index.js";
+import { check } from "./util.mjs";
+
+export default {
+	name: "Hooks Tests",
+	description: "TBD",
```
LeaVerou, line 6:
```suggestion
```

It's not a mandatory field, so no need to have a description that just says "TBD".

LeaVerou (approved):
Thanks! See minor comment but otherwise LGTM!

PR #485, opened 2024-03-14
[Gamut mapping App] Add port of Chromium Gamut mapping app

jamesnw:
This ports the algorithm available behind a flag in Chrome Canary, implemented [here](https://chromium-review.googlesource.com/c/chromium/src/+/5147800/13/ui/gfx/color_conversions.cc). 

The approach appears to be different from the others, reach roughly the same results as the CSS algorithm, and is as fast as clip.

I expect that there could be additional cleanup, this was intended to be as straight of a port as possible from the original code.

- [x] Verify what color space it is mapping to - rec2020
- [x] Don't map in-gamut colors
- [x] Can produce colors out of p3 gamut- should it be clipped? Added clip to rec2020.

### DIFF
--- a/apps/gamut-mapping/gradients.js
+++ b/apps/gamut-mapping/gradients.js
@@ -12,7 +12,7 @@ let app = createApp({
 		const urlToColor = params.get("to");
 		const from =  urlFromColor || "oklch(90% .8 250)";
 		const to = urlToColor || "oklch(40% .1 20)";
-		const methods = ["none", "clip", "scale-lh", "css", "raytrace", "edge-seeker"];
+		const methods = ["none", "chromium", "clip", "scale-lh", "css", "raytrace", "edge-seeker"];
 		const runResults = {};
 		methods.forEach(method => runResults[method] = []);
 		return {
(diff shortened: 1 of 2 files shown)


LeaVerou:
Wow, thanks for this! Why the change from `chromium` to `baked-in`? I thought `chromium` was more accurate.

jamesnw:
> Wow, thanks for this! Why the change from `chromium` to `baked-in`? I thought `chromium` was more accurate.

The [issue & description](https://chromium-review.googlesource.com/c/chromium/src/+/5147800) refer to it as "baked in", so I thought I would mirror the other methods, and name this by the methodology rather than source. I suspect there to be a large number of variations on what "baked-in" means, so if we add additional ones, we may want to add the provenance, e.g., `Baked in- Chromium, oklab to rec2020`.

facelessuser:
I do think for comparison purposes, this approach should clip to Display P3. Right now, you are comparing P3 deltas with Chromiums Rec. 2020 deltas, which makes the whole thing unhelpful. Visually, the browser will clip to whatever the display is using, but for raw number comparison purposes, we need actual P3 values.

jamesnw:
@facelessuser Good point- update pushed!

facelessuser:
I have no opinions on name. As a suggestion, you could call it chromium and reference the "baked-in" part in the description. Other than that, it all LGTM at this point.

facelessuser (approved):

jamesnw:
> As a suggestion, you could call it chromium and reference the "baked-in" part in the description.

Good idea- done!

facelessuser:
I think all of the issues everyone had have been addressed. It seems to yield the same results that Chromium does when the GMA flag is enabled as well. Since this doesn't affect core code at all, I will go ahead and merge this.

PR #486, opened 2024-03-14
[deltaE] Try out suggested DeltaE OK2

svgeesus:
Adds the suggested DeltaE OK2 from

 - https://github.com/w3c/csswg-drafts/pull/10063

and tests it against DeltaE OK with the Sharma color difference set.

Should also be tried with the CSS Color 4 GMA.


facelessuser (approved):

facelessuser:
Codewise, LGTM. I haven't tried playing with it integrated into the GMA yet.

PR #497, opened 2024-03-22
[docs] Point typedoc at `.d.ts` files instead of `.js`

### LINKED ISSUE #496: Auto-generated API docs seem not to be using TypeScript files
If you look at the new API page (<https://colorjs.io/api/>), many variables and parameters just have `any` as their types. It looks to me like TypeDoc is pointed at the wrong files (i.e. the JS source instead of `.d.ts` files).

Here's the page for the Color class, for example: <https://colorjs.io/api/classes/color.color>

A screenshot of part of the aforementioned link, showing that many properties have their types documented as "any."

MysteryBlokHed:
Closes #496

TypeDoc was set to use the JavaScript files for source, meaning all JSDoc from the TypeScript files was being ignored. This PR sets it to use the TS files instead, documenting types/interfaces and using the newer JSDoc.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,7 @@
 		"skipLibCheck": true
 	},
 	"include": [
-		"src/**/*.js"
+		"types/src/**/*.d.ts"
 	],
 	"exclude": [
 		"node_modules/**"
--- a/typedoc.json
+++ b/typedoc.json
@@ -3,11 +3,11 @@
 	"json": "./api/docs.json",
 	"name": "Color.js API Docs",
 	"entryPoints": [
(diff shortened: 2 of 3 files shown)


MysteryBlokHed:
If desired, I can also run `build:apidocs` on this PR so that the generated files are up-to-date.

LeaVerou (approved):
Thanks! I’m assuming you’ve checked that the result is better that way?

MysteryBlokHed:
> Thanks! I’m assuming you’ve checked that the result is better that way?

Yeah—using this, everything is typed (unlike the screenshots in #496)

PR #511, opened 2024-05-02
Add OkLrab and OkLrCh

### LINKED ISSUE #505: Add interpolation support for OkLrCH
Hi, I am working on an app to generate shades of colors for game artwork. During development I noticed that when using `steps()` to lerp over OkLCH, the result is too dark at lower luminance:

2024-04-20 15_39_52-Palettejuicer

[It looks like I'm not the only one who noticed this issue.](https://twitter.com/drewgeraets/status/1631423651269709824/photo/1)

It would be nice if OkLrCH were an option for interpolation, as I want to generate shades of color with roughly linear perceptual luminance, and the Lr function [mentioned here](https://bottosson.github.io/posts/colorpicker/#intermission---a-new-lightness-estimate-for-oklab) seems to give results close to that of LAB. The code for the function already seems to be in the library via the OkHSL colorspace. Would it be possible to repurpose this as OkLrCH for better luminance transitions in OkLCH?

facelessuser:
Resolves #505

### DIFF
--- a/test/conversions.js
+++ b/test/conversions.js
@@ -483,7 +483,7 @@ const tests = {
 				{
 					name: "sRGB white (D65) to OKlrch",
 					args: "white",
-					expect: [1.0, 0.0, NaN],
+					expect: [1.0, 0.0, null],
 				},
 				{
 					name: "sRGB red (D65) to OKlrch",
--- a/test/conversions.js
+++ b/test/conversions.js
@@ -430,19 +430,75 @@ const tests = {
 					expect: [1.0, 0.0, null],
 				},
 				{
-					name: "sRGB red (D65) to OKlab",
+					name: "sRGB red (D65) to OKlch",
 					args: "red",
 					expect: [0.6279553639214311, 0.2576833038053608, 29.23388027962784],
 				},
 				{
-					name: "sRGB lime (D65) to OKlab",
+					name: "sRGB lime (D65) to OKlch",
 					args: "lime",
 					expect: [0.8664396175234368, 0.2948272245426958, 142.4953450414439],
 				},
 				{
-					name: "sRGB blue (D65) to OKlab",
+					name: "sRGB blue (D65) to OKlch",
 					args: "blue",
 					expect: [0.45201371817442365, 0.3132143886344849, 264.0520226163699],
+				},
+			],
+		},
+		{
+			name: "OKLrab",
+			data: {
+				toSpace: "oklrab",
+			},
+			tests: [
+				{
+					name: "sRGB white (D65) to OKlrab",
+					args: "white",
+					expect:  [ 1.0000000000000002, -4.996003610813204e-16, 0 ],
+				},
+				{
+					name: "sRGB red (D65) to OKlrab",
+					args: "red",
+					expect: [ 0.5680846563197034, 0.2248630684262744, 0.125846277330585 ],
+				},
+				{
+					name: "sRGB lime (D65) to OKlrab",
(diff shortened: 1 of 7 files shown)


facelessuser:
Only lint issues related to the this change were addressed.

facelessuser:
This will clearly need a rework now that the null changes are in.

facelessuser:
Rebase completed and is ready for review.

test/conversions.js
```diff
@@ -430,22 +430,78 @@ const tests = {
 					expect: [1.0, 0.0, null],
 				},
 				{
-					name: "sRGB red (D65) to OKlab",
```
svgeesus, line 433: good catch, yes

test/conversions.js
```diff
@@ -430,22 +430,78 @@ const tests = {
...
+		{
+			name: "OKLrab",
+			data: {
+				toSpace: "oklrab",
+			},
+			tests: [
+				{
+					name: "sRGB white (D65) to OKlrab",
+					args: "white",
+					expect:  [ 1.0000000000000002, -4.996003610813204e-16, 0 ],
```
svgeesus, line 458: Should this actually be zero and this is cumulative error?

svgeesus (approved):
Approved, one question but not a blocker

svgeesus:
Someone more familiar with ts should probably look at the lint errors

test/conversions.js
facelessuser: Yeah, cumulative floating point error.

Oklab in Color.js currently yields that for white, so OkLrab will also yield that for white as OkLrab just applies a toe function to the lightness, and at white, there is no change.

```js
> new Color('white').to('oklab').coords
[ 1.0000000000000002, -4.996003610813204e-16, 0 ]
```

Floating point errors can vary slightly depending on how the original matrix was generated, the exact steps and in what order. The invert matrix result can be slightly different (floating point error-wise) depending on exactly how the inversion is performed (did it use LU decomposition, some other way, what order were operations done in, etc).

This isn't too different for other Lab-like spaces, especially those that employ more complicated calculations:

```js
> new Color('white').to('ictcp').coords
[ 0.5806888810416109, 1.1102230246251565e-16, 2.914335439641036e-16 ]
```

I think my library has slightly different floating point errors. It yields an exact 1 for lightness, but non-zero results for both a and b.

```py
>>> Color('white').convert('oklab').coords()
[1.0, -4.996003610813204e-16, 1.1102230246251565e-16]
```
But the same things happen for various other Lab-ish spaces (Y is the lightness in XYB):

```py
>>> Color('white').convert('hunter-lab').coords()
[100.0, 1.912989643463523e-14, -7.46260389005987e-15]
>>> Color('white').convert('xyb').coords()
[0.0, 0.8453085619621621, 2.220446049250313e-16]
```

facelessuser:
> Someone more familiar with ts should probably look at the lint errors

As far as this review is concerned, none of the lint errors apply to the code changes made in this PR. I consider these all existing, separate issues. I imagine it would be useful for someone to open a separate review to address these once and for all, but I do not consider any the lint errors to be blocking this review as they are unrelated and were not introduced by these changes.

jgerigmeyer:
> I do not consider any the lint errors to be blocking this review as they are unrelated and were not introduced by these changes

Agreed. If `main` is merged into this branch, CI may pass even with the lint errors.

facelessuser:
Rebased and fixed imports due to recent changes on main. The lint action now passes. I think this can be considered done.

PR #517, opened 2024-05-16
Ensure Okhsl and Okhsv return undefined hues for achromatic colors

### LINKED ISSUE #516: OkHSL/OkHSV does not use NaN for hue-less colors
The OkHSL/OkHSV spaces do not seem to use NaN for hue in their conversions, unlike other colorspaces such as HSL/HSV and LCH/OkLCH. One side effect of this is that interpolation from srgb(0.0, 0.0, 1.0) to srgb(1.0, 1.0, 1.0) produces incorrect results:

2024-05-16 15_03_29-Palettejuicer

The top strip is lerping over OkLCH, and the middle and bottom ones are using OkHSV and OkHSL, respectively. Because srgb(1.0, 1.0, 1.0) isn't converting to a hue of NaN, OkHSV and OkHSL are incorrectly blending their hues toward 0.

Reproduction:

```js
const white = { space: "srgb", coords: [1.0, 1.0, 1.0] };

const whiteOklch = to(white, "oklch"); // [ 1.0000000000000002, 4.996003610813204e-16, NaN ]
const whiteOkhsl = to(white, "okhsl"); // [ 180, 0.6519721306444567, 1.0000000000000002 ]
const whiteOkhsv = to(white, "okhsv"); // [ 180, 1.3189507366749435e-15, 1.0000000000000007 ]
```

facelessuser:
Fixes #516

### DIFF
--- a/src/spaces/okhsl.js
+++ b/src/spaces/okhsl.js
@@ -423,6 +423,10 @@ function okhslToOklab (hsl, lmsToRgb,  okCoeff) {
 function oklabToOkhsl (lab, lmsToRgb, okCoeff) {
 	// Oklab to Okhsl.
 
+	// Epsilon for lightness should approach close to 32 bit lightness
+	// Epsilon for saturation just needs to be sufficiently close when denoting achromatic
+	let εL = 1e-7;
+	let εS = 1e-4;
 	let L = lab[0];
 	let s = 0.0;
 	let l = toe(L);
@@ -458,7 +462,7 @@ function oklabToOkhsl (lab, lmsToRgb, okCoeff) {
 		}
 	}
 
-	if (Math.abs(s) < 1e-4 || l === 0.0 || Math.abs(1 - l) < 1e-7) {
+	if (Math.abs(s) < εS || l === 0.0 || Math.abs(1 - l) < εL) {
 		h = NaN;
 		s = 0.0;
 	}
--- a/src/spaces/okhsl.js
+++ b/src/spaces/okhsl.js
@@ -376,8 +376,8 @@ function okhslToOklab (hsl, lmsToRgb,  okCoeff) {
 
 	let [h, s, l] = hsl;
 	let L = toeInv(l);
-	let a = 0.0;
-	let b = 0.0;
+	let a = null;
+	let b = null;
 	h = constrain(h) / 360.0;
 
 	if (L !== 0.0 && L !== 1.0 && s !== 0) {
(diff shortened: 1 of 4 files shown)


src/spaces/okhsl.js
```diff
@@ -458,6 +458,11 @@ function oklabToOkhsl (lab, lmsToRgb, okCoeff) {
 		}
 	}
 
+	if (Math.abs(s) < 1e-4 || l === 0.0 || Math.abs(1 - l) < 1e-7) {
```
LeaVerou, line 461: Where do these constants come from?

facelessuser: It's an approximation. Considering that Okhsl and Okhsv is mainly 32 bit accurate, I only require the lightness to be close to that.

Generally, everywhere else in this library, close to zero for chroma is not required to be super close to actual zero. I think originally I ran some forward and backwards conversion between various color spaces and settled on 1e-4 for "sufficiently close". I don't recall the exact reason I selected that value, but that is what I did.

Generally, even the CSS spec does not specify what "sufficiently close to zero" is:

> It is intentionally unspecified exactly what "sufficiently close" means for this purpose.

Now that I look into Color.js, OkLCh uses the arbitrary value of 2e-4 to determine achromatic status, so if you prefer that, I can use that: https://github.com/color-js/color.js/blob/main/src/spaces/oklch.js#L31C13-L31C19. The difference is insignificant, but I can see it making sense for the sake of consistency.

facelessuser: I will also add that `l === 0.0` is a more reliable target than `l === 1.0`. This is generally true for _most_ color space algorithms. It is often observable that getting a precise 1 for lightness or 100 for lightness is algorithmically less likely due to floating point math than targeting a lightness of zero during conversion.

LeaVerou: Can we abstract them away to some kind of ε constant?

facelessuser: Sure.

LeaVerou (approved):
LGTM, thanks!

KeyboardDanni:
Will need updating due to the switch from `NaN` to `null`.

facelessuser:
Yeah, I was hoping to get in before all that. I'll rebase and update it probably tomorrow if I have time.

LeaVerou:
@facelessuser let me know if I can help!

facelessuser:
I've rebased and updated to use null.

LeaVerou (approved):
Changes LGTM, thanks!

As a side comment, these spaces appear to be duplicating a lot of logic that they should ideally import from other spaces, but that shouldn't hold this PR up.

facelessuser:
@LeaVerou Okhsl and Okhsv is duplicating logic? Which part? Did you mean to comment in the Oklrab PR? I could see the claim that that PR was duplicating code.

LeaVerou:
Perhaps it's not exactly duplication, but the conversion of these spaces to/from base is very weird, it seems to be going through sRGB Linear somehow? If so, why not just define sRGB linear as the base?

facelessuser:
To be honest, I'm pretty indifferent as to whether the base is Oklab or linear sRGB. If you want the base to be different, I can do it, and instead of having the Oklab to sRGB linear conversion, I'll add the linear sRGB to Oklab conversion. It's all the same to me. It all amounts to pretty much the same thing.

facelessuser:
For context, the original algorithm did conversion back and forth between llnear sRGB and Oklab, but I've made the initial base Oklab and only require a conversion to sRGB linear, not from, simplifying the code.

LeaVerou:
If it needs both color spaces, then I guess up to you.

facelessuser:
I'll leave it as is unless a specific need arises requiring such a change.

PR #532, opened 2024-05-28
`deltas()` function, closes #437

### LINKED ISSUE #437: New method: `deltas()`

LeaVerou:


facelessuser:
Shouldn't this have special handling for hues?

svgeesus:
> Shouldn't this have special handling for hues?

Yes, [as I said earlier](https://github.com/color-js/color.js/issues/437#issuecomment-2002152896)

svgeesus (changes requested):
Needs to calculate delta of the hue arc, not the angular delta

LeaVerou:
I envisioned this as a low-level utility function, if users want to make this optimization for hue it's trivial to do so on the deltas returned. 

That said, it should probably constrain the ΔH returned (possibly by also preserving the original).

LeaVerou:
Ok, I’ve made several improvements:
- A hue interpolation setting, with `"shorter"` being the default.
- Fixed a bunch of issues around `none` handling, and added more tests.

LeaVerou:
We could have another option for scaling the Δh by taking C/L into account, but not sure how that would work in a generic way.

facelessuser:
Looks good to me. I would fix lint issues. related to the files in questions.

facelessuser (approved):
LGTM

Issue #412, opened 2024-02-09
Discord server for Color.js


LeaVerou:
As our core maintainer team grows, we need some kind of more immediate/transient communication than GitHub issues, and also the ability to discuss things privately (e.g. I wanted to propose we invite someone as a new maintainer and it was entirely nontrivial to do so).

Color.js users may also benefit from something like this (we have the [Discussions](https://github.com/color-js/color.js/discussions) section but it's not used much).

I was considering of setting up a Discord server, what do others think? @svgeesus @jgerigmeyer @MysteryBlokHed 
Assuming you’re ok with it, could you either post your Discord usernames here or (for privacy) email them to me? (it's `[my first name]` @ `[my last name]` .me)

MysteryBlokHed:
A Discord server sounds good to me! My username is the same as on GitHub, `@mysteryblokhed`

jgerigmeyer:
Works for me -- `jgerigmeyer` 👍

svgeesus:
Nantonos

LeaVerou:
Could someone add something on the website about the Discord server?

LeaVerou:
Since that has happened as well, I’m now closing this.

PR #533, opened 2024-05-29
[website] Add Discord link to nav links

MysteryBlokHed:
Related issue: #412

This change uses the Discord invite linked from the referenced issue. Currently, the only change is adding a link to the navbar. I could also add some kind of "need help?" section to the homepage that's harder to miss, if that'd be desired.

### DIFF
--- a/_includes/partials/_nav-links.njk
+++ b/_includes/partials/_nav-links.njk
@@ -11,4 +11,5 @@
 <a href="https://elements.colorjs.io/">Elements</a>
 <a href="{{ page | relative }}/test/" class="footer">Tests</a>
 <a href="https://github.com/LeaVerou/color.js">GitHub</a>
+<a href="https://discord.gg/K64FJBznq4" class="footer">Discord</a>
 <a href="https://github.com/LeaVerou/color.js/issues/new" class="footer">File bug</a>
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
Thanks! Let's go with that for now and see how it goes.

MysteryBlokHed:
I know I've already merged this, but I wonder if it might be better to use a unique invite link for the website? That way it's easier to tell where new members are coming from based on the link they used, if that data is of interest

LeaVerou:
Sure, why not? How though?

MysteryBlokHed:
Under the Color.js server > Server Settings > Invites, a full list of invite links and their use count can be seen:

Screenshot of the Invites page, showing a single invite and its use count (23)

If you're interested, then you could create a new link as usual (eg. right-clicking on a channel > Invite People) and then either edit the link in the navbar, or share it with me so I can make the change. The only reason I'd ask you to create the invite instead of doing it myself is so that it shows "Lea Verou invited you to..." instead of my own name while joining.

Also, while creating the invite, make sure to set it not to expire (there's a tiny hyperlink at the bottom of the invite menu to configure the expiration)—we don't want to just have a broken link in the navbar :P

MysteryBlokHed:
You can't actually see which invite a specific user used without a third-party bot, but I think it's still worth doing so that we can see where people are coming from

PR #539, opened 2024-06-04
Update download count in README.md

MysteryBlokHed:
The package has now been downloaded more than 3 million times! :tada:

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ all with sensible defaults
 
 - Has been used to create demos for several W3C specifications
 - Has been used by browsers to test their CSS Color 4/5 implementations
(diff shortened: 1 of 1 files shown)


LeaVerou (approved):
Approving but that's exactly the type of change that would be fine as a direct commit.

MysteryBlokHed:
Fair enough. I just wasn't sure whether you were alright with pushing changes directly to main, even for something small like this :)

LeaVerou:
Yes, if it’s small uncontroversial things like this, no PR is needed. Just please check the diff first to make sure no unrelated changes have creeped in.

PR #540, opened 2024-06-06
[types] Migrate to JSDoc

### LINKED ISSUE #499: How to have a single source of truth for types and docs?

MysteryBlokHed:
Closes #499

This PR is the same as #528, but the source branch is now on the main color.js repo to make direct modifications easier.

<hr>

This PR migrates most `.d.ts` type definitions to JSDoc type annotations instead. This is done in an effort to keep code documentation and implementation in the same place.

Here's a checklist with tasks that I can think of so far:

- [x] Files directly in `src/` converted
- [x] Files in subfolders of `src/` converted (eg. `deltaE/`)
- [x] Typing tests fixed to work with JSDoc annotations (semi-complete? runs just fine but there are type errors)

There are a few files that are currently still using `.d.ts` instead of JSDoc due to their complexity. Namely: color.d.ts, hooks.d.ts, index.d.ts, and space.d.ts.

Since the typings tests are currently not working, there's a chance that I've missed some bugs while doing this conversion. Once I find a way to get dtslint to run with JSDoc types, it'll be easier to spot any mistakes.


MysteryBlokHed:
Should the release draft for v0.6.0 be updated to point to just this PR, or to point to both the original and this one? This doesn't have any of the discussion from #528, but I don't know whether we care about that for the release notes

LeaVerou:
> Should the release draft for v0.6.0 be updated to point to just this PR, or to point to both the original and this one? This doesn't have any of the discussion from #528, if that's something we care about for the release notes

Not sure, would love more opinions. We could mention both, or we could rely on the fact that this links to #528. I’m fine either way.

src/ColorSpace.d.ts
```diff
@@ -5,6 +5,8 @@
  */
 import { White } from "./adapt.js";
 import { ColorConstructor, Coords, ColorTypes } from "./color.js";
+import type FormatClasa from "./Format.js";
```
DmitrySharabin, line 8:
```suggestion
import type FormatClass from "./Format.js";
```

MysteryBlokHed: Whoops. Other references to this typo in the file will also need to be changed

MysteryBlokHed: Nevermind—it looks like the typo is only in the import

types/index.d.ts
```diff
@@ -0,0 +1 @@
+// Minimum TypeScript Version: 5.0
```
jgerigmeyer, line 1: @MysteryBlokHed This fixes dtslint, which I think is now returning only real errors.

jgerigmeyer (commented):
This is great! Definitely an improvement, and not surprising that it now reveals a number of existing type inconsistencies.

src/RGBColorSpace.js
```diff
@@ -3,19 +3,17 @@ import {multiplyMatrices} from "./util.js";
...
 import XYZ_D65 from "./spaces/xyz-d65.js";
 
-/**
- * Convenience class for RGB color spaces
- * @extends {ColorSpace}
- */
+// Type "imports"
+/** @typedef {import("./types.js").RGBOptions} RGBOptions */
+
+/** Convenience class for RGB color spaces */
```
LeaVerou, line 9: Why the change in formating and the removal of `@extends`?

LeaVerou (approved):
LGTM from a quick look. I would lave liked more locality for some of the types.js stuff, but we can explore later, it's imperative to merge this ASAP.

Thanks @DmitrySharabin for reverting the formatting changes, this was so much easier to review!

src/RGBColorSpace.js
MysteryBlokHed: I removed the `@extends` because TypeScript already knows due to `extends ColorSpace`. The formatting change was just because the JSDoc fit on a single line after removing `@extends`

MysteryBlokHed:
> LGTM from a quick look. I would lave liked more locality for some of the types.js stuff, but we can explore later, it's imperative to merge this ASAP.

I put a lot of things in the `types.d.ts` because it made it a lot faster to just copy+paste from the old definitions files, but I agree that it makes sense to move definitions to their own relevant files (later).

And sorry about the formatting 😅 I've turned off that extension for this repo

LeaVerou:
So is there any blocker remaining? Shall we merge and let it ride the trains?

MysteryBlokHed:
> So is there any blocker remaining? Shall we merge and let it ride the trains?

I think we're all good now. `dtslint` tests are still failing, but most errors seem to be due to the tests being out of date with the API rather than actual problems that need fixing.

MysteryBlokHed:
I or whoever else can merge as long as nobody else has any concerns

LeaVerou:
Ship it!

PR #547, opened 2024-06-11
[getAll] Support precision, closes #542

### LINKED ISSUE #542: [`getAll`] Add an option(?) so that formatted coords can be returned
This idea initially came from the need to [show coords in a color swatch](https://github.com/color-js/elements/issues/81) formatted as they are in the parsed color. Something like this:

image

@LeaVerou [pointed out](https://github.com/color-js/elements/pull/93#issuecomment-2158248994) that, unfortunately, there is no way to do it now, but it might be a good feature of `getAll()`.

So, I'm filing this issue to request this feature. 🙃

DmitrySharabin:


DmitrySharabin:
It's based on an outdated main branch. Close it for now so I can resend it later.

PR #548, opened 2024-06-11
[getAll] Support precision, closes #542

### LINKED ISSUE #542: [`getAll`] Add an option(?) so that formatted coords can be returned

DmitrySharabin:
Ensure that it’s something you had in mind, though. My approach is relatively straightforward. We should probably have an optional `options` parameter with `precision` as a property.

### DIFF
--- a/src/getAll.js
+++ b/src/getAll.js
@@ -1,6 +1,6 @@
 import ColorSpace from "./ColorSpace.js";
 import getColor from "./getColor.js";
-import { toPrecision } from "./util.js";
+import { toPrecision, type } from "./util.js";
 
 /** @typedef {import("./types.js").ColorTypes} ColorTypes */
 /** @typedef {import("./types.js").Coords} Coords */
@@ -8,16 +8,23 @@ import { toPrecision } from "./util.js";
 /**
  * Get the coordinates of a color in any color space
  * @param {ColorTypes} color
- * @param {string | ColorSpace} space The color space to convert to. Defaults to the color's current space
- * @param {number} [precision] The number of significant digits to round the coordinates to
+ * @param {string | ColorSpace | object} [options=color.space] If a string or ColorSpace, the color space to convert to. Defaults to the color's current space.
+ *                                       If an object:
+ *                                       [options.space=color.space] — the color space to convert to (defaults to the color's current space);
+ *                                       [options.precision] — the number of significant digits to round the coordinates to.
  * @returns {Coords} The color coordinates in the given color space
  */
-export default function getAll (color, space, precision) {
+export default function getAll (color, options) {
 	color = getColor(color);
 
-	if (typeof space === "number") {
-		precision = space;
-		space = color.space;
+	let optionsType = type(options);
+	let space, precision;
+	if (optionsType === "string" || options instanceof ColorSpace) {
+		space = options;
+	}
+	else if (optionsType === "object") {
+		space = options.space;
+		precision = options.precision;
 	}
 
 	let coords;
--- a/src/getAll.js
+++ b/src/getAll.js
@@ -1,6 +1,6 @@
 import ColorSpace from "./ColorSpace.js";
 import getColor from "./getColor.js";
-import { toPrecision, type } from "./util.js";
+import { toPrecision } from "./util.js";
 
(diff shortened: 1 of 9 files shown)


LeaVerou (changes requested):
Please note that I said:
> Specifying a color space (either as a string or `ColorSpace` object should still work.

Using `space` as an options param should only be needed when we’re *also* specifying precision.

MysteryBlokHed:
Something like this in `toGamut.js` could be used maybe?

<https://github.com/color-js/color.js/blob/3f97a65640ed283251fa125bbd6018ff74633c83/src/toGamut.js#L69-L71>

Where a destructured object is used as a parameter, but then `arguments` is used to check if something else was passed. So if it's a string or instance of a color space, then `space` is just set to `arguments[1]`.

DmitrySharabin:
> Something like this in `toGamut.js` could be used maybe?
> 
> https://github.com/color-js/color.js/blob/3f97a65640ed283251fa125bbd6018ff74633c83/src/toGamut.js#L69-L71
> 
> Where a destructured object is used as a parameter, but then `arguments` is used to check if something else was passed. So if it's a string or instance of a color space, then `space` is just set to `arguments[1]`.

I was thinking of something like this, too. Thank you so much for the hint! 🙏

src/getAll.js
```diff
@@ -1,23 +1,41 @@
...
 /** @typedef {import("./types.js").Coords} Coords */
 
 /**
  * Get the coordinates of a color in any color space
  * @param {ColorTypes} color
- * @param {string | ColorSpace} space The color space to convert to. Defaults to the color's current space
+ * @param {string | ColorSpace | object} [options=color.space] If a string or ColorSpace, the color space to convert to. Defaults to the color's current space.
+ *                                       If an object:
+ *                                       [options.space=color.space] — the color space to convert to (defaults to the color's current space);
+ *                                       [options.precision] — the number of significant digits to round the coordinates to.
```
MysteryBlokHed, line 14: It doesn’t seem to me like TypeScript is properly understanding the type here. Maybe making a `@typedef` for the function’s options would be best? I can add a code suggestion for that you agree

LeaVerou: Yes I think a typedef would make sense here. Though I thought we had found a way to do overloads?

MysteryBlokHed, line 14: Unfortunately I can't select the entire JSDoc and add a suggestion because you can't suggest on deleted lines, but for now I’ll just add a suggestion for this:

```suggestion
 * @param {string | ColorSpace | GetAllOptions} [options=color.space] If a string or ColorSpace, the color space to convert to. Defaults to the color's current space.
```

src/getAll.js
```diff
@@ -1,23 +1,41 @@
 import ColorSpace from "./ColorSpace.js";
 import getColor from "./getColor.js";
+import { toPrecision, type } from "./util.js";
 
 /** @typedef {import("./types.js").ColorTypes} ColorTypes */
 /** @typedef {import("./types.js").Coords} Coords */
```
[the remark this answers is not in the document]
DmitrySharabin: Thank you so much for your help!

src/getAll.js
DmitrySharabin: Yeah, I quite often need this feature to suggest changes to lines that are not in the PR.

DmitrySharabin:
@MysteryBlokHed, could you please check if I get the idea with overloads right?

LeaVerou (approved):

LeaVerou:
@DmitrySharabin do you have access to edit the release notes for v0.6.0 to mention this? (make sure not to accidentally publish!)

DmitrySharabin:
> @DmitrySharabin do you have access to edit the release notes for v0.6.0 to mention this? (make sure not to accidentally publish!)

It looks like I don't — I don't see any drafts of release notes.

PR #554, opened 2024-06-15
Fix serialization of negative percentages

### LINKED ISSUE #552: Incorrect serialization of negative values for sRGB and HSL
Negative values are serialized as `-Infinity`.

https://apps.colorjs.io/convert/?color=oklch(0%2C%200.1%2C%20180)&precision=4

image

lloydk:
Closes #552

Changes the hard coded percentage ranges to a variable range based on whether or not the lower bounds of the range is < 0. Spaces that do not have coordinate ranges (e.g. xyz-d65) will have a percentage range of [0, 1] and [0, 100].

### DIFF
--- a/src/Type.js
+++ b/src/Type.js
@@ -40,7 +40,7 @@ export default class Type {
 			return this.range;
 		}
 		if (this.type === "<percentage>") {
-			return [-1, 1];
+			return this.pctRange();
 		}
 		else if (this.type === "<angle>") {
 			return [0, 360];
@@ -72,7 +72,7 @@ export default class Type {
 		let toRange = this.coordRange;
 
 		if (this.type === "<percentage>") {
-			toRange ??= [-1, 1];
+			toRange ??= this.pctRange();
 		}
 
 		return mapRange(fromRange, toRange, number);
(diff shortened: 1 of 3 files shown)


src/Type.js
```diff
@@ -104,6 +105,15 @@ export default class Type {
 		return ret;
 	}
 
+	/**
+	 * Returns a percentage range for values of this type
+	 * @param {number} scale
+	 */
+	pctRange (scale = 1) {
```
LeaVerou, line 112: I’d rather avoid abbreviations like "pct".
```suggestion
	percentageRange (scale = 1) {
```

LeaVerou (approved):
Thanks! LGTM, see comment wrt naming.

LeaVerou:
Gah, now this introduces an error, since we changed the name of the function but not of the function calls. The suggestion was indicative, sorry if that wasn't clear!

lloydk:
> Gah, now this introduces an error, since we changed the name of the function but not of the function calls. The suggestion was indicative, sorry if that wasn't clear!

Oops, I was watching tv and wasn't really paying attention.

PR #556, opened 2024-06-15
Fix failing constructor test

lloydk:


LeaVerou:
That would have been fine as a direct commit btw :)

Issue #398, opened 2024-02-03
[chore] Port Matrix multiplication tests to the new test format


lloydk:

LeaVerou:
Thanks again @epsilonError!

PR #559, opened 2024-06-20
Port Multiply matrices tests to new format

epsilonError:
An attempt at porting the Matrix Multiplication tests to the new format.

I wanted to try out hTest and thought working on #398 would be a good way to start.

~I was hoping to use the `math.js` matrix multiplication methods as the test oracle, but I couldn't figure out how to get the dynamic loading in the HTML-based testing to properly import it.~

~I've left comments for all of that code I used, just in case anyone else knows how to make that work.~
Feel free to edit this branch (maintainers) or fork it (everyone else).


LeaVerou:
Thank you for working on this!

For mathjs and the HTML testsuite there are two ways to do it:
1. Add an import map on all pages that maps mathjs to a CDN URL
2. Import a specific file from `node_modules` rather than the alias, we can then add a redirect like [this one](https://github.com/color-js/elements/blob/main/_redirects#L5) to make sure `node_modules` resolves on the deployed website.

LeaVerou:
This points out a very interesting use case for hTest: so far it assumes that you know in advance whether a test should pass or produce an error, but if your tests are simply comparing with the output from something else, you may not know. I’ll think of what’s the best way to address this use case in hTest.

LeaVerou:
> To emulate failing being expected for the Incorrect Data tests, the solution I came up with is having `check:` always return true.

Given that the Node runner doesn’t output anything for these, are they really useful? Should we remove them? I'll defer to the others.

LeaVerou:
Oh interesting. `expect` cannot work like that since its actual value may be a function, but perhaps we could have `getExpect()` (akin to `getName()`) that works that way. I wonder what the arguments would be though.

LeaVerou (approved):
LGTM

lloydk (approved):
LGTM

LeaVerou:
> > Oh interesting. `expect` cannot work like that since its actual value may be a function, but perhaps we could have `getExpect()` (akin to `getName()`) that works that way. I wonder what the arguments would be though.
> 
> What I assumed could be the case is the same arguments that are given to `run` would be provided to `getExpect()` and it would have access to all the same data as `run`.
> 
> But I'm only thinking of data being passed as arguments, not functions. I have much better intuitive sense of data manipulation tasks and pass by value arguments, but that is just my bias.
> 
> I think one of the bigger benefits from a `getExpect()` implementation would be deprecation testing of function interfaces, tests for backwards compatibility, and tests for possible lossless/lossy conversions from one function interface to the other.
> 
> `expect` could be v1 arguments, `getExpect()` could use the v1 function interface and `run` & `arg`/`args` would be a newer implementation (or vice versa). But I have no idea how feasible this would be to do in hTest.
> 
> My naive assumption would be `getExpect()` gets the same arguments as `run` if no `expect` is given. But would prioritize `expect` if it exists. This is similar to how I assume `args` and `arg` work. Though I'd bet that pair must have different names to be prioritized correctly, which I haven't accounted for with this `getExpect()` idea.

I have a lot of thoughts, but I think this is not the right place for this discussion — would you mind opening an issue over at hTest so we could discuss this and flesh out the design? Thanks!

PR #564, opened 2024-06-24
[types] Generate declarations from JS

### LINKED ISSUE #563: [types] Generate declarations from JS
Created a separate issue to move discussion away from #555.

Adding the following options to the root `tsconfig.json` allows declaration files to be generated from JS:

```json
"declaration": true,
"emitDeclarationOnly": true
```

The only concern I have is that it clogs up the repository with many `.d.ts` files, which shouldn't be committed. This can be partially solved with `.gitignore` (and then adding exceptions for the `.d.ts` files that _are_ meant to be there), but it's still kind of annoying as a developer.

I tried changing the `outDir` option to put the declaration files into their own directory (like they used to be before the JSDoc migration), but it doesn't seem to copy over the `.d.ts` source files (like `index.d.ts`).

MysteryBlokHed:
Closes #563

Builds declaration files automatically instead of letting editors use JSDoc.

Currently, source maps are also enabled, since it lets developers see exactly where the types for a given file actually come from.


package.json
```diff
@@ -41,11 +54,12 @@
 		"watch:html": "npx @11ty/eleventy --config=_build/eleventy.js --serve",
 		"build:js": "rollup -c _build/rollup.config.js",
 		"watch:js": "rollup -c _build/rollup.config.js --watch",
+		"build:ts": "tsc && node ./scripts/copy-declarations.js",
```
lloydk, line 57: When I run `build:ts` I get compilation errors so the copy-declarations script is never run.

```suggestion
		"build:ts": "node ./scripts/copy-declarations.js && tsc",
```

scripts/copy-declarations.js
```diff
@@ -0,0 +1,31 @@
...
+	"hooks.d.ts",
+	"index.d.ts",
+	"space-coord-accessors.d.ts",
+	"space.d.ts",
+	"types.d.ts",
+];
+
+/** Script file directory */
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
```
lloydk, line 25: Create the destination directory if it doesn't exist. This is needed if we swap the order of operations in `build:ts`.

```suggestion
mkdirSync(resolve(__dirname, DEST), { recursive: true });
```

lloydk (commented):
I still have some compilation errors in my project but I think we're getting closer.

lloydk (approved):
LGTM

scripts/copy-declarations.js
```diff
@@ -0,0 +1,31 @@
+// Copy manually-written `.d.ts` files from `src/` to `types/src/`
+import { copyFileSync } from "node:fs";
```
lloydk, line 2: Sorry, missed the import in my previous suggestion.
```suggestion
import { copyFileSync, mkdirSync } from "node:fs";
```

scripts/copy-declarations.js
```diff
@@ -0,0 +1,31 @@
...
+ * `.d.ts` files to copy.
+ * Paths written relative to the `src/` directory
+ */
+const TO_COPY = [
+	"color.d.ts",
+	"ColorSpace.d.ts",
+	"hooks.d.ts",
+	"index.d.ts",
+	"space-coord-accessors.d.ts",
+	"space.d.ts",
```
lloydk, line 19: `"space.d.ts"` no longer exists.

Issue #551, opened 2024-06-13
JSDoc migration broke some type re-exports


MysteryBlokHed:
The tests `type-accessibility.ts` and `type-fn-accessibility.ts` are currently both failing. The migration to JSDoc seems to have excluded some types that were defined/re-exported in the `index.d.ts` and `index-fn.d.ts` files. These should be re-added to preserve compatibility.

<https://github.com/color-js/color.js/blob/b6e92aa8bf396d7f106a950a240698011a724bce/types/test/type-accessibility.ts#L1-L37>

<https://github.com/color-js/color.js/blob/b6e92aa8bf396d7f106a950a240698011a724bce/types/test/type-fn-accessibility.ts#L1-L32>

PR #565, opened 2024-06-26
Fix types and type tests

lloydk:
Fixes some of the types and type tests.

With these changes plus #563 and #551 was able to get one of my projects to compile and run.


MysteryBlokHed (approved):
Fixes all LGTM!

src/space.d.ts
MysteryBlokHed, line 1: Not even sure how this file got left behind

PR #566, opened 2024-06-26
[types] Remove module augmentation

### LINKED ISSUE #561: Type error: Expected 0 arguments, but got 1 on Color constructor
Super odd one here, on Vercel I see this error:

```
./src/components/form/ColourInput/useColourInput.ts:44:32
Type error: Expected 0 arguments, but got 1.
  42 |
  43 |     try {
> 44 |       const colour = new Color(props.value);
     |                                ^
  45 |
  46 |       const hue = angleToHue(colour.lch["h"] ?? 0);
  47 |
error Command failed with exit code 1.
```

But locally, totally fine.

Source code here: https://github.com/Southclaws/storyden/blob/main/web/src/components/form/ColourInput/useColourInput.ts#L3

This appears to be the only library affected in this codebase so I'm not sure what's happening. I've also tried downgrading to 0.4.5 but still doesn't work. 

Unfortunately this is preventing progress so I might just have to remove color.js until a fix is found.

MysteryBlokHed:
Closes #561
**Note:** This change is for v0.5 of color.js, hence the target branch being `v0.5.x`. It should also be implemented before the release of v0.6.0, but that can be done at a later date.

Removes the module augmentation from `index.d.ts`, instead putting the relevant properties directly into `color.d.ts`. This should the errors being reported on newer versions of TypeScript.


MysteryBlokHed:
@jgerigmeyer Requested review because I'd like to know whether you think there's any better way to do this—imo it seems like module augmentations should just be avoided with the weirdness they seem to have caused.

jgerigmeyer (approved):
Works great! I agree that this is better than the flaky module augmentation we were using before. 👍

PR #567, opened 2024-06-27
[types] Remove module augmentation (v0.6)

MysteryBlokHed:
This is essentially the same change as #566, but for the main branch instead of a patch for v0.5.

I don't think there should be anything missing here—I re-copied and pasted the augmentation contents from `index.d.ts` to `color.d.ts` to make sure I didn't miss any new properties—but review has been requested just in case :)


lloydk:
LGTM

I tested it with #569 and all the type tests passed.

lloydk (approved):

jgerigmeyer (commented):
These specific changes look good, but it looks like since #564 was merged, the entire build process is failing because of errors in the `tsc` step. Even combining this with #569 doesn't help, so it's difficult for me to test. What am I missing?

lloydk:
I don’t think build:ts is being run. On my phone with no reading glasses so
I could be wrong. If that isn’t it then I’ll take a look when I get home.

MysteryBlokHed:
`tsc` is probably reporting type errors from the actual logic of the code. That should probably actually be fixed within the code, but in the meantime, is there some way to force it to exit with code 0? Just so we can actually get the build to run.

lloydk:
> `tsc` is probably reporting type errors from the actual logic of the code. That should probably actually be fixed within the code, but in the meantime, is there some way to force it to exit with code 0? Just so we can actually get the build to run.

I created #571 that should allow the build to run.

MysteryBlokHed:
I'll try merging this and the other active typing PR's and see what the situation is like then

PR #569, opened 2024-06-27
Fix remaining type test failures

lloydk:
Fixes the remaining type test failures. 🤞


MysteryBlokHed:
Would you mind checking whether the errors are from the tests, or the codebase itself? I'm on mobile right now, so I can't.

jgerigmeyer:
The errors are from dtslint -- likely related to: https://github.com/color-js/color.js/pull/571#issuecomment-2195420830

PR #571, opened 2024-06-27
Run build:ts in lint and lint:ci

lloydk:
Due to typescript errors in the code base the `tsc` portion of `build:ts` is hardcoded to always return a 0 exit code.

This should work on Linux and Mac (I only tested it on Linux). If we want to support building on Windows then we need a different strategy for the exit code. There are a number of npm packages that will handle the exit code in a cross platform manner. (e.g. https://github.com/kachkaev/suppress-exit-code).


lloydk:
I'm confused as to why there are so many dtslint errors in CI. I cloned this repository into a new directory, did an npm install added the changes in this PR ran `npm run lint:ci` and got a different set of dtslint errors (errors fixed by #569).

MysteryBlokHed (approved):

MysteryBlokHed (approved):

package.json
```diff
@@ -48,13 +48,13 @@
 		"test": "npx htest ./test",
 		"dtslint": "dtslint --expectOnly types",
 		"eslint": "eslint . --ext .js --ext .ts --ext .cjs",
-		"lint": "run-s build:space-accessors \"eslint -- --fix\" dtslint",
-		"lint:ci": "run-s build:space-accessors \"eslint -- --max-warnings 0\" dtslint",
+		"lint": "run-s build:space-accessors \"eslint -- --fix\" build:ts dtslint",
+		"lint:ci": "run-s build:space-accessors \"eslint -- --max-warnings 0\" build:ts dtslint",
```
MysteryBlokHed, line 52: I think `build:ts` should run in an earlier step to make sure that all the files are in place before `dtslint` runs. Maybe add a build step to the CI file, after `npm install` and before `npm run lint:ci`?

```yaml
            - name: Generate TypeScript definitions
              run: npm run build:ts
```

MysteryBlokHed: Nevermind—I confused `run-s` with `run-p`

PR #574, opened 2024-06-30
[types] Fix type errors in codebase

### LINKED ISSUE #572: [types] Fix type errors in codebase
There are currently 176 type errors coming from the codebase itself. These are definitely worth fixing (not necessarily before v0.6.0, but at some point), so I'd like some feedback as to the best way to fix certain kinds of type-specific errors (i.e. type errors that aren't actually indicative of a problem with the JS logic).

A relatively common one is `TS2322: Type 'number[]' is not assignable to type '[number, number, number]'`. An example of code causing this can be seen here:

<https://github.com/color-js/color.js/blob/c99665cb2dfb9e1cefb1f6c764a41b9af7cc89e2/src/adapt.js#L72-L74>

Since the return value of `multiplyMatrices` is not specific enough, it causes the aforementioned error. There are a few ways I can think of fixing this.

The first would be to simply ignore the error:

```typescript
// @ts-expect-error
return multiplyMatrices(env.M, env.XYZ);
```

The second would be to assert as the proper type (`[number, number, number]`):

```typescript
return /** @type {[number, number, number]} */ (multiplyMatrices(env.M, env.XYZ));
```

And the third would just be to assert as `any`:

```typescript
return /** @type {any} */ (multiplyMatrices(env.M, env.XYZ));
```

I find it annoying that the type assertions with JSDoc are so ugly, but I'm not sure whether it's worth completely ignoring type checks for a line instead with a `@ts-expect-error` directive. Thoughts?

MysteryBlokHed:
Closes #572

Fixes type errors originating from the codebase itself.

In some instances, the logic was changed where it seemed like the type error reflected an actual problem. For example, adding `color = Color.get(color)` where the signature made it seem as though a function could take `ColorTypes`, but actually couldn't.

### DIFF
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -127,8 +127,8 @@ export default class ColorSpace {
 	aliases?: string[] | undefined;
 	base: ColorSpace | null;
 	coords: Record<string, CoordMeta>;
-	fromBase?: ((coords: Coords) => number[]) | undefined;
-	toBase?: ((coords: Coords) => number[]) | undefined;
+	fromBase?: ((coords: Coords) => Coords) | undefined;
+	toBase?: ((coords: Coords) => Coords) | undefined;
 	formats: Record<string, Format>;
 	referred?: string | undefined;
 	white: White;
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -16,6 +16,7 @@ export interface Format {
 	id?: string | undefined;
 	coords?: string[] | undefined;
 	coordGrammar?: (string & { range?: [number, number] })[] | undefined;
+	space?: ColorSpace | undefined;
 	serializeCoords?:
 	| ((coords: Coords, precision: number) => [string, string, string])
 	| undefined;
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -107,7 +107,7 @@ export default class ColorSpace {
 		workingSpace?: string | ColorSpace
 	): CoordMeta & {
 		id: string;
-		index: string | number;
+		index: number;
 		space: ColorSpace;
 	};
 
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -77,6 +77,7 @@ export interface SpaceOptions {
 	 */
 	formats?: Record<string, Format> | undefined;
 	gamutSpace?: "self" | string | ColorSpace | null | undefined;
+	aliases?: string[] | undefined;
 }
 
 export type Ref =
@@ -118,7 +119,8 @@ export default class ColorSpace {
 
 	static registry: Record<string, ColorSpace>;
 
-	get all (): Set<ColorSpace>;
+	static get all (): ColorSpace[];
+
 	/** The ID used by CSS, such as `display-p3` or `--cam16-jmh` */
 	get cssId (): string;
 	get isPolar (): boolean;
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -138,7 +138,7 @@ export default class ColorSpace {
 	from (color: ColorTypes): Coords;
 	from (space: string | ColorSpace, coords: Coords): Coords;
 
-	getFormat (format?: string | Format): Format | null;
+	getFormat (format?: string | Format | FormatClass): FormatClass | null;
 
 	getMinCoords (): Coords;
 
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -14,6 +14,7 @@ export interface Format {
 	/** @default "color" */
 	name?: string | undefined;
 	id?: string | undefined;
+	ids?: string[] | undefined;
 	coords?: string[] | undefined;
 	coordGrammar?: (string & { range?: [number, number] })[] | undefined;
 	space?: ColorSpace | undefined;
--- a/src/Format.js
+++ b/src/Format.js
@@ -21,6 +21,7 @@ export const instance = Symbol("instance");
 
 /**
  * @class Format
+ * @implements {Omit<FormatInterface, "coords" | "serializeCoords">}
  * Class to hold a color serialization format
  */
 export default class Format {
--- a/src/Format.js
+++ b/src/Format.js
@@ -3,6 +3,7 @@ import Type from "./Type.js";
 
 // Type "imports"
 /** @typedef {import("./types.js").ColorSpace} ColorSpace */
+/** @typedef {import("./types.js").Coords} Coords */
 /** @typedef {import("./types.js").Format} FormatInterface */
 
 /**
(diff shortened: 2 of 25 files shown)


src/clone.js
```diff
@@ -1,15 +1,16 @@
 // Type "imports"
+/** @typedef {import("./color.js").default} Color */
 /** @typedef {import("./types.js").PlainColorObject} PlainColorObject */
 
 /**
  *
- * @param {PlainColorObject} color
+ * @param {Color | PlainColorObject} color
```
lloydk, line 7: Color implements the `PlainColorObject` so adding `Color` shouldn't be necessary.

src/display.js
```diff
@@ -7,6 +7,7 @@ import REC2020 from "./spaces/rec2020.js";
 import P3 from "./spaces/p3.js";
 import Lab from "./spaces/lab.js";
 import sRGB from "./spaces/srgb.js";
+import Color from "./color.js";
```
lloydk, line 10: The functional interface should never reference the `Color` class.

lloydk (commented):
Didn't do a thorough review, added a few comments.

src/clone.js
MysteryBlokHed: Right—for some reason I keep forgetting that

src/display.js
MysteryBlokHed: Good point. Should I change the signature of the function to only accept `PlainColorObject`, since that seems to be what the function is meant to handle?

lloydk: `serialize` calls `getColor` on the color so `ColorTypes` should be ok.

MysteryBlokHed: Didn't realize that these lines also expect `PlainColorObject`:

<https://github.com/color-js/color.js/blob/c0a48bf283f7858cf589e81f5566a485299aa344/src/display.js#L49-L52>

So I just put the call at the beginning of the function instead.

MysteryBlokHed:
Not 100% sure that the commit I just pushed for `Format` is entirely correct, so any comments would be appreciated

lloydk:
> Not 100% sure that the commit I just pushed for `Format` is entirely correct, so any comments would be appreciated

The `Format` stuff is complicated, I took a look at it last week and gave up after a few hours.

I'll take a closer look after I take the dog for a walk but some stuff that's missing:

- `ColorSpace.getFormat` should return the `Format` class not the `Format` interface.
- The `Format` class should have almost every property from the `Format` interface. see https://github.com/color-js/color.js/blob/c0a48bf283f7858cf589e81f5566a485299aa344/src/Format.js#L36

In the `Format` class I didn't like how the `coords` property changed from an array of strings to `Type[][]` so in the PR that I abandoned I renamed `Format.coords` to `coordTypes`.

lloydk:
The Format changes look good although I'm not 100% sure about the type for `Format.coords`. If `format.type === "custom"` it's `[string, string, string]` otherwise it's `Type[][]`. I think it would be better to have two different properties.

MysteryBlokHed:
> If `format.type === "custom"` it's `[string, string, string]` otherwise it's `Type[][]`. I think it would be better to have two different properties.

Ah, I didn't even notice that. I don't think there's a great way to type it without separating the two properties. We could ask Lea for her thoughts in regards to making a new property, since she wrote the class?

lloydk:
> > If `format.type === "custom"` it's `[string, string, string]` otherwise it's `Type[][]`. I think it would be better to have two different properties.
> 
> Ah, I didn't even notice that. I don't think there's a great way to type it without separating the two properties. We could ask Lea for her thoughts in regards to making a new property, since she wrote the class?

We could just leave it as `Type[][]` since you probably won't have a `coords` property for a custom format.

src/ColorSpace.d.ts
```diff
@@ -16,6 +16,7 @@ export interface Format {
 	id?: string | undefined;
```
lloydk, line 16:
```suggestion
	id?: string | undefined;
	ids?: string[] | undefined;
```

The zyz-d65 space uses the `ids` property.

MysteryBlokHed:
Sorry I haven't worked on this for a while, I've been busier than usual. If anybody else wants to push commits to fix errors then feel free—this branch is on the main repository instead of a fork so permissions shouldn't be an issue.

LeaVerou:
Sorry I’ve been absent as well, my thesis defense is in a month so I've been directing all my resources towards it. Should this PR block releasing v0.6.0?

lloydk:
> Sorry I’ve been absent as well, my thesis defense is in a month so I've been directing all my resources towards it. Should this PR block releasing v0.6.0

At the very least we should merge what we have so far in this PR before releasing v0.6.0. I had to switch my project to this branch because I needed some of the fixes in this PR.

LeaVerou:
Does anyone have a sense of what % done this PR is?

lloydk:
> Does anyone have a sense of what % done this PR is?

maybe 80%

MysteryBlokHed:
> At the very least we should merge what we have so far in this PR before releasing v0.6.0.

I agree. We might as well merge what we have so far and worry about the rest of the errors some other time, since most errors don't actually seem to indicate any logical problems in the code.

LeaVerou:
Agreed

LeaVerou:
@MysteryBlokHed should I just merge this then? It's still marked as draft.

MysteryBlokHed:
Oops, I forgot to mark it ready—fixed

src/Format.js
```diff
@@ -98,7 +110,7 @@ export default class Format {
 			if (!type) {
 				// Type does not exist in the grammar, throw
 				let coordName = coordMeta.name || id;
-				throw new TypeError(`${ providedType ?? arg?.raw ?? arg } not allowed for ${coordName} in ${this.name}()`);
+				throw new TypeError(`${ providedType ?? /** @type {any} */ (arg)?.raw ?? arg } not allowed for ${coordName} in ${this.name}()`);
```
LeaVerou, line 113: Excuse my ignorance, why do we need inline `@type` comments? Is this information we cannot capture in the block jsdoc comments? (a link is fine, I just googled it and couldn't figure it out)

MysteryBlokHed: This is how type assertions work when using JSDoc.
Type assertions: <https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions>
JSDoc syntax for assertions: <https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#casts>

It's needed whenever TypeScript can't properly determine the type of something within the body of a function (eg. for variables that were defined within that function).

For the line you commented on, TypeScript would complain that the property `raw` doesn't exist on `number`. The easiest way to stop it from complaining was to assert `arg` as `any`, effectively disabling type checking for `arg` on that particular line.

It's also sometimes needed when a function's return type isn't specific enough, like here:

<https://github.com/color-js/color.js/blob/9661dc697426a3983e7d390daea5e49e80291f05/src/RGBColorSpace.js#L42>

LeaVerou: Ah, I see, thank you!

LeaVerou (approved):
LGTM, I think this is a net improvement. Since this is a large one, I suggest we merge soon before stuff gets out of sync.

PR #585, opened 2024-08-13
Add specialized matrix transform for Vector3 and Matrix3x3

lloydk:
Adds a specialized matrix multiplication function (`transform`) that transforms an array of length 3 with a 3x3 matrix. The specialized function can also transform the array in place which can eliminate one array allocation.

The https://github.com/texel-org/color color library has a benchmark comparing itself to Color.js.

Profiling the conversion benchmark showed that matrix multiplication was one of the areas that was taking the most amount of time.

This PR speeds up the conversion benchmark by more than 3x and some of the other benchmarks by 2-3x.

### Benchmark numbers before this PR:

```
conversion (Colorjs.io procedural API) --
Colorjs.io: 6213.74 ms
Ours: 733.75 ms
Speedup: 8.5x faster

conversion (Colorjs.io main API) --
Colorjs.io: 25444.47 ms
Ours: 703.50 ms
Speedup: 36.2x faster

gamut mapping OKLCH - sRGB (Colorjs.io procedural API) -- Colorjs.io: 13302.80 ms
Ours: 72.97 ms
Speedup: 182.3x faster

gamut mapping OKLCH - sRGB (Colorjs.io main API) -- Colorjs.io: 14900.11 ms
Ours: 67.78 ms
Speedup: 219.8x faster

gamut mapping all spaces to P3 (Colorjs.io procedural API) -- Colorjs.io: 10878.26 ms
Ours: 201.85 ms
Speedup: 53.9x faster

gamut mapping all spaces to P3 (Colorjs.io main API) -- Colorjs.io: 13731.21 ms
Ours: 201.76 ms
Speedup: 68.1x faster
```
### After this PR:
```
conversion (Colorjs.io procedural API) --
Colorjs.io: 1793.97 ms
Ours: 713.03 ms
Speedup: 2.5x faster

conversion (Colorjs.io main API) --
Colorjs.io: 19884.18 ms
Ours: 704.59 ms
Speedup: 28.2x faster

gamut mapping OKLCH - sRGB (Colorjs.io procedural API) -- Colorjs.io: 4981.04 ms
Ours: 74.01 ms
Speedup: 67.3x faster

gamut mapping OKLCH - sRGB (Colorjs.io main API) -- Colorjs.io: 6477.12 ms
Ours: 68.02 ms
Speedup: 95.2x faster

gamut mapping all spaces to P3 (Colorjs.io procedural API) -- Colorjs.io: 4205.36 ms
Ours: 202.27 ms
Speedup: 20.8x faster

gamut mapping all spaces to P3 (Colorjs.io main API) -- Colorjs.io: 6137.99 ms
Ours: 199.91 ms
Speedup: 30.7x faster
```

There are still some color spaces that haven't been switched over to use the `transform` function. Assuming we want to proceed with this PR I'll create a separate PR to deal with the remaining spaces.


facelessuser (approved):
Looks good to me.

src/multiply-matrices.js
```diff
@@ -85,3 +90,58 @@ export default function multiplyMatrices (A, B) {
...
+ * @param {Vector3} [out]
+ * @returns {Vector3}
+*/
+export function transform (input, matrix, out = [0, 0, 0]) {
+	const x = dot3(input, matrix[0]);
+	const y = dot3(input, matrix[1]);
+	const z = dot3(input, matrix[2]);
+	out[0] = x;
+	out[1] = y;
+	out[2] = z;
```
LeaVerou, line 145: Presumably something like `matrix.map(m => dot3(input, m))` was slower due to the function?

LeaVerou (approved):
Thanks for working on this! LGTM, one nit is that the name `transform()` seems a bit too generic.

src/multiply-matrices.js
facelessuser: Direct access via array index is most certainly faster than a generalized iteration function. There is no doubt overhead for such convenience.

lloydk: There is some overhead to using `map` and using `map` won't allow you to transform the input in place resulting in an extra array allocation in the places where the transform is done in place (oklab color space).

I played around a bit with replacing some of the `map` calls in the RGB color spaces  with direct array access and got a modest performance improvement. Speedup for the procedural API conversion benchmark dropped to around 2.2 or 2.3 (although that might have included some optimizations to the `adapt` function, I can't remember).

lloydk:
> Thanks for working on this! LGTM, one nit is that the name `transform()` seems a bit too generic.

I agree but wasn't sure what to call it, I was hoping someone would suggest a better name 😃

LeaVerou:
> > Thanks for working on this! LGTM, one nit is that the name `transform()` seems a bit too generic.
> 
> I agree but wasn't sure what to call it, I was hoping someone would suggest a better name 😃

It multiplies a vector of length 3 with a 3x3 matrix, right? What about `multiply3()` or `multiply_3x3()` or even `multiply_3_3x3()`?

src/multiply-matrices.js
LeaVerou: Fair enough. I suspected but figured I'd check.

PR #591, opened 2024-08-26
Add color space for Linear Rec2100

### LINKED ISSUE #443: Support `rec2100-linear`?
I wonder if we should support [`rec2100-linear`](https://drafts.csswg.org/css-color-hdr/#valdef-color-rec2100-linear) from CSS Color HDR. It was  [requested](https://github.com/w3c/csswg-drafts/issues/553#issuecomment-1809200438) by the Color on the Web CG, primarily for Canvas, and adopted by [CSSWG resolution](https://github.com/w3c/csswg-drafts/issues/553#issuecomment-1864865513).

Although it is an HDR space, it is defined with `white` === `color(rec2100-linear 1 1 1)` (so the actual HDR values are above 1). That makes it an **exact alias** of `color(--rec2020-linear)`, because BT.2100 uses the BT.2020 primaries and white point, differing only in the transfer function.

lloydk:
Closes #443

I wasn't 100% sure if I should use the `toBase` and `fromBase` functions from Linear Rec2020 or import the matrices from Linear Rec2020 and set the `toXYZ_M` and `fromXYZ_M` options.

### DIFF
--- a/src/space-coord-accessors.d.ts
+++ b/src/space-coord-accessors.d.ts
@@ -35,6 +35,7 @@ declare class SpaceAccessors {
 	prophoto_linear: SpaceAccessor;
 	rec2020: SpaceAccessor;
 	rec2020_linear: SpaceAccessor;
+	rec2100_linear: SpaceAccessor;
 	rec2100hlg: SpaceAccessor;
 	rec2100pq: SpaceAccessor;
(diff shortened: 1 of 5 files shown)


svgeesus:
Looks good overall, not sure what to do about the linting errors though.

svgeesus (approved):

Issue #587, opened 2024-08-23
An unnecessary full style recalculation is done every scroll event on many pages of `colorjs.io`


Inwerpsel:
When checking dev tools for the site, I noticed that every scroll event the `--scrolltop` custom property of the `html` element is updated.

While at a glance this might not look expensive, it actually results in a style recalculation of all elements on the page. I'm pretty sure it's the only reason the scrolling is choppy, and removing this property update is enough to make the page scroll normally.
<details>
<summary><h3>Performance recording during scrolling</h3></summary>

<img src='https://github.com/user-attachments/assets/6415686a-21cc-4d6e-86d1-4f87988561e6' id='perf-during-scroll' />

Details of a single recalc, taking 25.18ms on desktop.

![colorjs-scrolling-recalc-detail](https://github.com/user-attachments/assets/794c40f3-d7b2-443f-900a-0a2ad971455c)

</details>

This is triggered by an event listener that only sets a custom property on the root element.

https://github.com/color-js/color.js/blob/f50d95a542b6a487d0ffdd004b62879f200c148d/assets/js/index.js#L16-L18

The only reference to this custom property is in the `#toc` element, where it's used to move a `position: fixed` element depending on the scroll position.

https://github.com/color-js/color.js/blob/f50d95a542b6a487d0ffdd004b62879f200c148d/assets/css/docs.css#L13-L24

It looks a lot like this CSS can be replaced with `position: sticky`, completely removing the need to update the root property and thereby removing all the time spent on the purple stuff in above graphs.

svgeesus:
Scrolling is indeed choppy; feel free to send a pull request!

Inwerpsel:
I created a second PR with a much simpler solution that should have no risk and still get us a similar performance improvement by reducing the scope of the recalculation.

PR #592, opened 2024-08-27
Avoid style recalculation of all elements on each scroll event

Inwerpsel:
## Description

Addresses[ laggy scrolling](https://github.com/color-js/color.js/issues/587) resulting from unnecessary style recalculations.

The most important thing about this PR is that this listener is removed.

```js
document.addEventListener("scroll", evt => {
	root.style.setProperty("--scrolltop", root.scrollTop);
}, {passive: true});
```

Without the listener, the main thread is almost completely idle during scrolling. With the listener, the main thread is [fully occupied with tasks that take hundreds of milliseconds on lower end devices](https://github.com/color-js/color.js/issues/587#user-content-perf-during-scroll). That means scrolling at a few FPS, 100% of the time.

The only thing this listener is doing is make an element behave as though it has `position: sticky`. So the goal of the PR is to convert it to `position: sticky` and remove the need for any style recalculation.

Using `position: sticky` also has better behavior in other ways. The previous code results in the bottom items not being reachable on long lists (such as [/spaces](https://colorjs.io/docs/spaces)).

## Probably the best solution

The markup should probably be changed to be similar to the [/api page](https://colorjs.io/api/).

tocoutside

For `position: sticky` to work, you need to have a container that spans the height you want the element to travel. So you'd want a container which is positioned next to the content and is equally high.

However, in the current markup the TOC element is inside of the content, so `position: sticky` cannot produce the intended effect of flowing next to the content.

tocinside

## This PRs CSS only solution

When trying out this markup change, I noticed it would cascade into too many other CSS changes and associated regression risk. I preserved [a simple attempt at modifying the markup](https://github.com/Inwerpsel/color.js/commit/ce28accff85c60422381a4084f026ab7d365d9b4) on a separate branch.

So it seemed warranted to try removing the performance issue with an as minimal change as possible while only changing the visual appearance for the better (i.e. it does the same except when the previous behavior was unintended, like the behavior near the bottom of the page).

The "trick":
* position `#toc` absolutely, to allow putting it next to the content
* give it 100% the height of its container
* also push it 100% to the right.
* 
* This should reliably put the element into the position the previous code was putting it.

## Performance improvement

The following screenshot compares performance during scrolling on the /spaces pages, [before](https://colorjs.io/docs/spaces) and [after ](https://deploy-preview-592--colorjs.netlify.app/docs/spaces) these changes. Both are on desktop, with no CPU throttling, so the improvement on lower end devices will be much greater.

* Time spent on style recalculation: 90% -> 0% (remaining purple is other rendering tasks)
* About 1 major GC / 200ms scrolling -> no major GC due to scrolling
* 25fps and main thread fully occupied -> 144fps and >50% idle time left

colorjs-before-after

### DIFF
--- a/assets/css/docs.css
+++ b/assets/css/docs.css
@@ -15,13 +15,13 @@ main {
 		position: absolute;
 		right: 100%;
 		height: 100%;
-		/* This is a quick fix for `fit-content` below not working anymore. */
-		min-width: 220px;
+		/* Still sorting what to use for the 3em. */
+		max-width: calc(var(--page-margin) - 3em);
+		width: max-content;
 
 		& > ul {
 			position: sticky;
 			top: 1em;
-			width: fit-content;
 		}
 	}
 
--- a/assets/css/docs.css
+++ b/assets/css/docs.css
@@ -11,15 +11,17 @@ main {
 	}
 
 	@media (min-width: 1480px) {
-		position: fixed;
-		top: 11rem;
-		right: calc(var(--page-width) + var(--page-margin) + 1em);
-		width: fit-content;
-		max-width: calc(var(--page-margin) - 1em);
-		margin-left: 1em;
+		/* Stretch out #toc in the margin, so it can be a container for the sticky list. */
+		position: absolute;
+		right: 100%;
+		height: 100%;
+		/* This is a quick fix for `fit-content` below not working anymore. */
+		min-width: 220px;
 
-		@supports (top: max(1em, 1px)) {
-			top: max(0em, 11rem - var(--scrolltop) * 1px);
+		& > ul {
+			position: sticky;
+			top: 1em;
+			width: fit-content;
 		}
 	}
 
--- a/assets/css/docs.css
+++ b/assets/css/docs.css
@@ -11,17 +11,19 @@ main {
 	}
 
 	@media (min-width: 1480px) {
+		--_gap: 1rem;
+		box-sizing: border-box;
 		/* Stretch out #toc in the margin, so it can be a container for the sticky list. */
 		position: absolute;
 		right: 100%;
 		height: 100%;
-		/* Still sorting what to use for the 3em. */
-		max-width: calc(var(--page-margin) - 3em);
+		padding-inline: var(--_gap);
+		max-width: calc(var(--page-margin) - 2 * var(--_gap));
 		width: max-content;
 
 		& > ul {
 			position: sticky;
-			top: 1em;
+			top: var(--_gap);
 		}
 	}
 
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -9,10 +9,5 @@ if (location.pathname.indexOf("/docs/") > -1 && window.toc) {
 	import("./docs.js");
 }
 
-let root = document.documentElement;
-
 styleCallouts();
 
-document.addEventListener("scroll", evt => {
-	root.style.setProperty("--scrolltop", root.scrollTop);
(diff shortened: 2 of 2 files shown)


LeaVerou:
If we have to set a fixed width, maybe set it in a font-relative unit so we can make CSS changes without having to adjust it.

Inwerpsel:
> If we have to set a fixed width, maybe set it in a font-relative unit so we can make CSS changes without having to adjust it.

Yeah I tried to preserve the previous behavior as much as possible, the previous fixed value was a quick fix.

I was now able to make it have the same behavior as before. Not by making `fit-content` work, but by using `max-content` and also giving it a `max-width`. I didn't quite figure out yet which spacing value to reference for the 3em, or if there is any.

Inwerpsel:
@LeaVerou Are the current changes are an acceptable way to address the performance issue? I added some info about the improvement in the description.

I did not find any regressions due to these changes. So it seems like a safe way to improve performance for all users, while removing the urgency of changing the markup to be similar to the /api page so it can accommodate sticky elements in a cleaner way.

Personally, I wouldn't worry too much about it using this 3em, or the exact spacing values, if the current result works well. If any later CSS changes need other dynamic values / variables to be in that place, that is something that can be changed at that time, and it will be more obvious what needs to go there.

Of course, feel free to solve this in any other way. I mostly created this PR to show what the problem is and that it can be addressed.

DmitrySharabin:
I really like the proposed fix—It solves the issue elegantly. 👍

I would suggest minor changes to make the TOC even more responsive to the different screen sizes. Something like this:

```diff
@media (min-width: 1480px) {
+	--_gap: 1rem;

+	box-sizing: border-box;

	/* Stretch out #toc in the margin, so it can be a container for the sticky list. */
	position: absolute;
	right: 100%;
	height: 100%;
+	padding-inline: var(--_gap);
-	/* Still sorting what to use for the 3em. */
-	max-width: calc(var(--page-margin) - 3em);
+	max-width: calc(var(--page-margin) - 2 * var(--_gap));
	width: max-content;

	& > ul {
		position: sticky;
-		top: 1em;
+		top: var(--_gap);
	}
}
```

I didn't want to suggest them in the code so that we could discuss them here, but if it's more convenient for you, I can do that. I hope you don't mind. 🙃

https://github.com/user-attachments/assets/cbbe1b66-36a2-40f5-9cfd-50256dbb3837

Inwerpsel:
Thanks @DmitrySharabin! Those changes look good to me, I added them to the PR.

I guess there are still some reasons to (eventually) use the same approach (or same code) as the TOC elements on the `/api` page. Those are independently scrollable and also work on smaller screens. They come from an API generator, so perhaps you can find and extract the template it uses for that.

DmitrySharabin:
> Thanks @DmitrySharabin! Those changes look good to me, I added them to the PR.

Great! 

> I guess there are still some reasons to (eventually) use the same approach (or same code) as the TOC elements on the [`/api` page](https://colorjs.io/api/).

I would do it in another PR—it seems a bit orthogonal to the issue we are trying to solve here. Even with the changes you made, the website works and feels much better. So, I wouldn't hold this PR and ask @LeaVerou to review it. Let's see what she will tell us. 🙂

Inwerpsel:
> I would do it in another PR

Yes definitely, just mentioning it as a suggestion in case future changes happen to the TOC 🙂

LeaVerou (approved):
LGTM, thank you all!!

PR #594, opened 2024-09-06
[tests] Add more matrix multiply tests

epsilonError:
Added more tests for matrix multiply. 
And utility functions so additional tests for `multiply_v3_matrix3x3` will be easy to add.

The new tests are skipped, because they are provisional, and I just want to highlight them.
The expected results are still visible in the HTML runner version of the tests.

If these tests seem reasonable for generic matrix multiplication, I'd be happy to make minimal changes to `multiplyMatrices` that would pass these tests in a future pull request.

I have some broader questions around how matrix multiplication should work for `color.js`, but I started a [discussion thread](https://github.com/color-js/color.js/discussions/584) to have that broader conversation.


lloydk (approved):
LGTM, sorry for the delay.

PR #595, opened 2024-09-07
parse.js fixes to allow "custom" ColorSpace formats

sidewayss:
Two very simple fixes that come out of this discussion: https://github.com/color-js/color.js/discussions/593

1) The code was finding a match for a custom `parse()` function, but only exiting the inner of two loops.  I added a new variable: `found`, that the inner loop sets to true so that the outer loop knows to exit.  This was not only causing unnecessary iterations of the loop, it was causing issues with custom user ColorSpaces that have formats with `{type:"custom"}`.
2) The addition of a single `?` in this line: `let color = format.parse?(env.str);` so that "custom" ColorSpace formats don't require a `parse` property.  The discussion linked above creates a custom ColorSpace with a default format that has a custom `serialize()` property/function, but doesn't need a `parse` property.  The current code requires it to have a dummy `parse()` property/function that returns falsy.  This change removes that requirement and prevents the throwing of unnecessary errors.

It is worth noting that this outside loop on `ColorSpace.all` is inefficient relative to its contents, even after fix 1.  There is only one built-in space with custom formats: sRGB, and it's the near the end of `ColorSpace.all`.  I'm no expert, but based on what I've seen I can't imagine any other built-in spaces doing custom parsing.  As for user color spaces, I don't know if any other users have created any, and mine doesn't parse.  Are there pending proposals for new CSS color strings that parse in other spaces?  Is there a need for a user color space that parses a custom, non-CSS color string?

If you want to ensure that user spaces and new built-in spaces can do custom parsing, you can create a new `ColorSpace.custom` array containing sRGB all other registered spaces that have custom formats.  You could add spaces to that array in `register()`.  Otherwise it's just sRGB, no need to iterate over any other color spaces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,12 +29,12 @@ all with sensible defaults
 </header>
 
 <section>
-	
+
 ## Impact
 
 - Has been used to create demos for several W3C specifications
 - Has been used by browsers to test their CSS Color 4/5 implementations
-- Over [9 million total npm downloads](https://limonte.dev/total-npm-downloads/?package=colorjs.io)!
+- Over [10 million total npm downloads](https://limonte.dev/total-npm-downloads/?package=colorjs.io)!
 - Used by several [high impact projects](https://www.npmjs.com/browse/depended/colorjs.io), including [Sass](https://sass-lang.com/), [Open Props](https://open-props.style/), [axe](https://www.deque.com/axe/) accessibility testing engine, and [OddContrast](https://www.oddcontrast.com/) and [CSS HD Gradients](https://gradient.style/) color tools
 - Parts of Color.js’s API are used as a testing ground for the design of a [native `Color` object for the Web platform](https://github.com/wicg/color-api).
 
@@ -72,8 +72,8 @@ import deltaE200 from "https://colorjs.io/src/deltaE/deltaE2000.js";
 
 Warning: To use `import` statements in a browser, your `<script>` needs `type="module"`
 
-Are you old school and prefer to simply have a global `Color` variable? 
-We’ve got you covered! 
+Are you old school and prefer to simply have a global `Color` variable?
+We’ve got you covered!
 Just include the following script in your HTML:
 
 ```html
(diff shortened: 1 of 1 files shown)


sidewayss:
It is worth noting that this outside loop on `ColorSpace.all` is highly inefficient relative to its contents.  There is only one built-in space with custom formats: `srgb`.  I'm no expert, but based on what I've seen I can't imagine any other built-in spaces doing custom parsing.  As for user color spaces, I don't know if any other users have created any, and mine doesn't parse.

If you want to ensure that user spaces can do custom parsing, you could create a new `ColorSpace.custom` array containing `srgb` and any registered user spaces that have custom formats.  You could add to the array in `register()`.  Otherwise it's just `srgb`, no need to iterate over any other color spaces.

LeaVerou:
Oh wow, this completely fell through the cracks, so sorry @sidewayss! Do you want to rebase so we can review? I think the type info in this is out of date, which is why it's showing so many type issues.

LeaVerou:
Btw if the inner loop needs to break the outer loop, you can do that with labels: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label

LeaVerou:
Did you mean to close this? Is the issue no longer present?

sidewayss:
No, I mean to create a new PR.  I didn't think you'd be so quick to notice.

This PR is based on my fork's main branch.  That's a mistake on my part.  I needed to rebase/reset my main branch prior to creating a new branch and implementing these tiny changes.  I have implemented it as a `break label` command.

Now my problem is that in testing the updated code I have run into this issue: #627,  which is impeding my testing in the moment.  I need a workaround for getting the registry if I want to test this, and I'd prefer to test it before submitting the new PR.

sidewayss:
Replacement PR here: #628

It fails the Lint & Test Types / Lint & Test Types (pull_request) tests, but that's everywhere, in lots of files I haven't touched.

PR #597, opened 2024-10-02
Avoid full style recalculation by setting custom property only on element that uses it

Inwerpsel:
Addresses https://github.com/color-js/color.js/issues/587

This PR is a safer alternative to https://github.com/color-js/color.js/pull/592 as it does not change any of the CSS logic, it only sets the custom property at a lower level in the tree.

Since the `--scrolltop` custom property is only used on the `#toc` element, we can set its value on that element instead of the root element. This avoids a *full* style recalculation of all elements on the page which takes very long, and causes massive framerate drops during scrolling.

This should not result in any changed behavior or appearance as no other element uses the custom property. It should be a pure optimization with no risk for regressions.

Comparing the performance recordings before and after this change, you can observe the time spent on style recalculation is reduced to a fraction. The long purple chunks shrunk became almost too small to see on the right.

colorjs-before-after-toc-optim

https://colorjs.io/docs/spaces vs https://deploy-preview-597--colorjs.netlify.app/docs/spaces

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -9,10 +9,13 @@ if (location.pathname.indexOf("/docs/") > -1 && window.toc) {
 	import("./docs.js");
 }
 
-let root = document.documentElement;
-
 styleCallouts();
+let toc = document.getElementById("toc");
 
-document.addEventListener("scroll", evt => {
-	root.style.setProperty("--scrolltop", root.scrollTop);
-}, {passive: true});
+if (toc) {
+	let root = document.documentElement;
+
+	document.addEventListener("scroll", () => {
+		toc.style.setProperty("--scrolltop", root.scrollTop);
+	}, {passive: true});
+}


LeaVerou:
Thanks! Though ideally, we need to fix this even for pages that _do_ have a TOC. @DmitrySharabin is actually investigating a potential fix right now.

Inwerpsel:
This does also address the performance issue on pages with a TOC. On those pages, in the previous version it would have to style recalculate all elements on the page before it was able to paint the scroll. With these changes, it only recalculates the TOC itself, which speeds the recalculation up by a factor of about 20 to 50 (see screenshot above of tasks run during scrolling on a page with a TOC). This effectively removes the performance issue.

So it might be a quick win if the required changes would take a bit longer.

DmitrySharabin:
@Inwerpsel [Your fix with CSS](https://github.com/color-js/color.js/pull/592) does the job nicely, to my mind. We don't need the `scroll` event with that fix, so we can safely ditch it. I would follow that route. I left a [comment](https://github.com/color-js/color.js/pull/592#issuecomment-2391356570) for you in that PR. I would love to know your thoughts on it.

PR #598, opened 2024-10-03
Fix return type for deltas function

lloydk:
The inferred type generated by Typescript for the result of the deltas function was incorrect. The `space` property had a type of `string | ColorSpace` when it should have been `ColorSpace`.

A new type called `DeltasReturn` has been added to fix the issue.

@LeaVerou not sure why we're returning the `space` and the `spaceId`. Can `spaceId` be removed?

### DIFF
--- a/src/deltas.js
+++ b/src/deltas.js
@@ -39,7 +39,7 @@ export default function deltas (c1, c2, {space, hue = "shorter"} = {}) {
 
 	let alpha = subtractCoords(c1.alpha, c2.alpha);
 
-	return { space: /** @type {ColorSpace} */ (c1.space), spaceId: ColorSpace.get(c1.space).id, coords, alpha };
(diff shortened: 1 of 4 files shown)


LeaVerou:
> @LeaVerou not sure why we're returning the `space` and the `spaceId`. Can `spaceId` be removed?

Search the codebase for `spaceId` and if you find no compelling reason, I'm fine to remove it.

lloydk:
> > @LeaVerou not sure why we're returning the `space` and the `spaceId`. Can `spaceId` be removed?
> 
> Search the codebase for `spaceId` and if you find no compelling reason, I'm fine to remove it.

The `deltas()` tests used it so I changed the tests and removed the property from the return type.

MysteryBlokHed (approved):
LGTM

PR #599, opened 2024-10-04
Fix typo

DmitrySharabin:
At the same time, we fix the build process on Netlify.

For now, we have this:

image

### DIFF
--- a/data/spaces.js
+++ b/data/spaces.js
@@ -5,7 +5,7 @@ let whitePoints = Object.entries(WHITES);
 
 // 11ty chokes on this
 // import modules from "./modules.json" with { type: "json" };
-const modules = await (await fetch(new URL("data/modules.json", import.meta.url)).json());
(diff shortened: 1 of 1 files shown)


DmitrySharabin:
Even with this change, we have the same issue: `fetch()` fails on the build because we get a URL with a non-supported protocol. Why? We are building the URL using `import.meta.url` as a base: https://github.com/color-js/color.js/blob/3e20e82d774641871bc13780d1cb13a03108f756/data/spaces.js#L8 In Node, it returns the protocol that `fetch()` doesn't support.

Even if the site works fine for now, it's because the previous build is being served. All the builds [starting from this commit](https://github.com/color-js/color.js/commit/3e20e82d774641871bc13780d1cb13a03108f756) will fail.

LeaVerou (approved):

DmitrySharabin:
Since this PR doesn't fix the build on Netlify, I will close it in favor of #600, which fixes both issues.

PR #605, opened 2024-10-28
Replace `instanceof` checks to work across color.js sources

### LINKED ISSUE #602: OOP API should work with Color objects imported from a different Color.js URL
I keep running into these issues right now and it's incredibly frustrating. If I try to use `Color` objects that have been created by a different Color.js (even if the versions are identical), I keep getting errors like: 

image

Most of them originate in `ColorSpace.get()`, though it's possible once we fix that other issues may creep up. But given that we're compatible even with plain object literals, I'd consider this a bug.

MysteryBlokHed:
Closes #602

Instead of relying solely on `instanceof` to check if an argument is an instance of some class (eg. Color or ColorSpace), also checks the name of the constructors of the instance to see if any of them have the same name as the target class.

This method isn't perfect, but it should work well enough given that it seems unlikely for a user of color.js to define their own unrelated class whose name conflicts with the library.

I'm open to changing this way of checking inheritance if anybody has better ideas.

### DIFF
--- a/releases/v0.6.0.md
+++ b/releases/v0.6.0.md
@@ -112,6 +112,10 @@ There is also a new app:
 - API docs that are actually up to date, using typedoc! You can find them in (by @leaverou with help from @MysteryBlokHed in #498 #497 #549)
 - Updated color space diagram in https://colorjs.io/docs/spaces which is now dynamically generated via https://d2lang.com/ (by @leaverou)
 
+### Bug fixes
+
+- Object-oriented functions now work between different sources of Color.js (#605)
+
 ### For contributors
 
 * Document how to serve in development, and add `--serve` to `watch:html`. by @jamesnw in https://github.com/color-js/color.js/pull/467
--- a/releases/v0.6.0.md
+++ b/releases/v0.6.0.md
@@ -114,7 +114,7 @@ There is also a new app:
 
 ### Bug fixes
 
-- Object-oriented functions now work between different sources of Color.js (#605)
+- Object-oriented functions now work between different sources of Color.js (by @MysteryBlokHed in #605)
 
 ### For contributors
 
--- a/src/color.d.ts
+++ b/src/color.d.ts
@@ -159,6 +159,10 @@ declare class Color extends SpaceAccessors implements PlainColorObject {
 		| Record<string, DefineFunctionHybrid>
 	): void;
 
+	static isColorInstance (arg: any): arg is Color;
+
+	static isColor (arg: any): arg is (Color | { coords: Coords });
+
 	get space (): ColorSpace;
 	get spaceId (): string;
 
--- a/src/color.d.ts
+++ b/src/color.d.ts
@@ -159,10 +159,6 @@ declare class Color extends SpaceAccessors implements PlainColorObject {
 		| Record<string, DefineFunctionHybrid>
 	): void;
 
-	static isColorInstance (arg: any): arg is Color;
-
-	static isColor (arg: any): arg is (Color | { coords: Coords });
-
 	get space (): ColorSpace;
 	get spaceId (): string;
 
--- a/src/color.js
+++ b/src/color.js
@@ -122,7 +122,7 @@ export default class Color {
 	 * Basically gets us the same result as new Color(color) but doesn't clone an existing color object
 	 */
 	static get (color, ...args) {
-		if (color instanceof Color) {
+		if (Color.isColorInstance(color)) {
 			return color;
 		}
 
@@ -181,6 +181,12 @@ export default class Color {
 				Color.defineFunction(name, exports[name]);
 			}
 		}
+	}
+
+	static isColorInstance = util.isInstance(this);
+
+	static isColor (arg) {
+		return Color.isColorInstance(arg) || Array.isArray(arg?.coords);
 	}
 }
 
--- a/src/color.js
+++ b/src/color.js
@@ -122,7 +122,7 @@ export default class Color {
 	 * Basically gets us the same result as new Color(color) but doesn't clone an existing color object
 	 */
 	static get (color, ...args) {
-		if (Color.isColorInstance(color)) {
+		if (util.isInstance(color, this)) {
 			return color;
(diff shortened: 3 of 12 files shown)


src/getColor.js
```diff
@@ -37,7 +37,7 @@ export default function getColor (color, options) {
 	// Object fixup
 	let space = color.space || color.spaceId;
 
-	if (!(space instanceof ColorSpace)) {
+	if (!ColorSpace.isColorSpace(space)) {
 		// Convert string id to color space object
 		color.space = ColorSpace.get(space);
```
LeaVerou, line 42: We could actually refactor this check to check if `space` is a string.

LeaVerou (changes requested):
Hey, thanks so much for working on this!

I love the idea of a generic `isInstance()` factory! Do we still need the public-facing APIs? Not sure we have authors that need to check if an object is a ColorSpace or a Format (I'd wager in most cases `ColorSpace.get()` would suffice). It seems more useful for `Color` since it does more than the instance check, though even there a static `isColor()` could be misconstrued as to also account for parseable color strings.

Since the API needs some more discussion, can we skip the public-facing APIs for now and just use private `const`s in each corresponding module? It seems in most cases the `instanceof` check is in the same module as the class it's checking, with one exception that can actually be refactored to a string check.

src/color.js
```diff
@@ -182,6 +182,12 @@ export default class Color {
 			}
 		}
 	}
+
+	static isColorInstance = util.isInstance(this);
+
+	static isColor (arg) {
+		return Color.isColorInstance(arg) || Array.isArray(arg?.coords);
```
lloydk, line 189: I think the second part of the check should test to see if `arg` conforms to the `PlainColorObject` interface.

LeaVerou:
Or actually — not use a const at all. Instead of `isInstance()` being a factory forcing people to call it in two steps, we could make that an implementation detail by using a WeakMap to associate constructors with their isInstance functions and every time it's called, we could check if we have already generated a function for this constructor, and if not, generate it and store it in the cache. Or, a regular object literal that keys the constructor name to the function. 

Though wait a second, why are we generating functions at all? There is no one-time work that we are saving that way, it's literally function currying, that users of this function can do with `function.bind()`.

src/color.d.ts
```diff
@@ -159,6 +159,10 @@ declare class Color extends SpaceAccessors implements PlainColorObject {
 		| Record<string, DefineFunctionHybrid>
 	): void;
 
+	static isColorInstance (arg: any): arg is Color;
+
+	static isColor (arg: any): arg is (Color | { coords: Coords });
```
lloydk, line 164:
```suggestion
	static isColor (arg: any): arg is (Color | PlainColorObject);
```

lloydk (changes requested):

lloydk:
> I love the idea of a generic `isInstance()` factory! Do we still need the public-facing APIs? Not sure we have authors that need to check if an object is a ColorSpace or a Format (I'd wager in most cases `ColorSpace.get()` would suffice). It seems more useful for `Color` since it does more than the instance check, though even there a static `isColor()` could be misconstrued as to also account for parseable color strings.

I'm not convinced we need the static `isColor()` function mostly for the reason you mentioned.

If we think it may be useful then I think we should add a standalone function that checks to see if an object conforms to the `PlainColorObject` interface so that it can be used from the procedural API. `Color.isColor()` can then do the `isInstance()` check and call the new standalone function.

MysteryBlokHed:
> Though wait a second, why are we generating functions at all? There is no one-time work that we are saving that way, it's literally function currying, that users of this function can do with `function.bind()`.

I wrote `isInstance` that way just as a convenience when adding the static methods for other classes—I didn't mean for it to be part of the public API that library consumers would use, since I figure it would be more intuitive for developers to call eg. `ColorSpace.isColorSpace` rather than something like `isInstance(arg, ColorSpace)`.

LeaVerou:
Yeah, I think for now let's punt the static methods and just use `IsInstance()` directly.

lloydk (approved):
LGTM

src/getColor.js
lloydk: the `util.isString()` is "slow" and came up as a hotspot when I was running the [texel](https://github.com/texel-org/color) benchmarks so I don't think we should add an additional call to `isString()`.

LeaVerou: Oh interesting. Maybe just `typeof space === "string"` then?

MysteryBlokHed: Changed in 0123830e005e08e9511d8b8e63c6734dc7490e30

LeaVerou:
Is this ready for a re-review? Also, could you please add an entry to the changelog in the same PR?

MysteryBlokHed:
> Is this ready for a re-review?

Should be ready now

> Also, could you please add an entry to the changelog in the same PR?

Done!

releases/v0.6.0.md
```diff
@@ -112,6 +112,10 @@ There is also a new app:
 - API docs that are actually up to date, using typedoc! You can find them in (by @leaverou with help from @MysteryBlokHed in #498 #497 #549)
 - Updated color space diagram in https://colorjs.io/docs/spaces which is now dynamically generated via https://d2lang.com/ (by @leaverou)
 
+### Bug fixes
+
+- Object-oriented functions now work between different sources of Color.js (#605)
```
LeaVerou, line 117:
```suggestion
- Object-oriented functions now work between different sources of Color.js (by @MysteryBlokHed in #605)
```

LeaVerou: Pretty sure there's an existing section on bugfixes btw

LeaVerou (approved):
LGTM, see comments on changelog

releases/v0.6.0.md
MysteryBlokHed: I tried to Ctrl+F for "bug" or "fix" but didn't find anything (besides the list of PR's that still need to be added to the log)

PR #606, opened 2024-11-03
[build] Build color-fn with ESM exports

### LINKED ISSUE #589: ESM export for `color-fn`
Hey there!

It seems that `color-fn` is only exported as `cjs`. Would it make sense to export it as `esm` too?
Would love to use it in [observablehq](https://observablehq.com).

Or is there a way to use it there without the Class API?

Thanks in advance!

Benny

MysteryBlokHed:
Closes #589

Also builds the color-fn API with ESM exports, which was previously missing.

### DIFF
--- a/_build/rollup.config.js
+++ b/_build/rollup.config.js
@@ -27,6 +27,11 @@ const fnBundles = [
 		"sourcemap": true,
 		"exports": "named",
 	},
+	{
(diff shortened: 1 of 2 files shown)


LeaVerou (approved):
LGTM, thanks!

Issue #553, opened 2024-06-14
REC.21000-HLG doesn't handle negative values when converting from rec2020-linear


lloydk:
`rec2100hlg` converts negative values to `NaN` when converting from `rec2020-linear`.

https://apps.colorjs.io/convert/?color=color(--rec2020-linear%20-0.0001%200%200)&precision=4

![image](https://github.com/color-js/color.js/assets/65072/6b0d452a-7d38-4f3a-84f9-b4c1de2bfd86)

facelessuser:
This issue is caused by the algorithm implementation in its current form. When a channel is less than 0.5, a square root is applied to the value. Negative square roots will return an undefined value. If a sane value is desired, we could utilized a signed square root using "signed power (spow)" to at least return a predictable value:

```diff
diff --git a/src/spaces/rec2100-hlg.js b/src/spaces/rec2100-hlg.js
index ba497ed..899016e 100644
--- a/src/spaces/rec2100-hlg.js
+++ b/src/spaces/rec2100-hlg.js
@@ -1,5 +1,6 @@
 import RGBColorSpace from "../RGBColorSpace.js";
 import REC2020Linear from "./rec2020-linear.js";
+import {spow} from "../util.js";

 // FIXME see https://github.com/LeaVerou/color.js/issues/190

@@ -41,7 +42,7 @@ export default new RGBColorSpace({
                        // ITU-R BT.2390-10 p.23
                        // 6.1 The hybrid log-gamma opto-electronic transfer function (OETF)
                        if (val <= 1 / 12) {
-                               return Math.sqrt(3 * val);
+                               return spow(3 * val, 0.5);
                        }
                        return a * Math.log(12 * val - b) + c;
                });
```

facelessuser:
I've issued a PR for review if such a fix is desired.

PR #607, opened 2024-11-12
Remove duplicates and mentions of tiny PRs

DmitrySharabin:
Those are the changes to the list at the end of the release notes.

### DIFF
--- a/releases/v0.6.0.md
+++ b/releases/v0.6.0.md
@@ -139,24 +139,16 @@ Remove those already mentioned above (or so small they don't need to be):
 
 
 * Fix types for `setAll()` by @lloydk in https://github.com/color-js/color.js/pull/520
-* Fix toGamut() types by @lloydk in https://github.com/color-js/color.js/pull/544
-* Fix RGBColorSpace type tests by @lloydk in https://github.com/color-js/color.js/pull/543
-* Add type for keywords by @lloydk in https://github.com/color-js/color.js/pull/545
-* Fix getAll() types by @lloydk in https://github.com/color-js/color.js/pull/546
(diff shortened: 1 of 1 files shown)


LeaVerou (changes requested):
I would tend to just group these under larger items (e.g. there's one for improving types) because you still want to give the appropriate credit to people.

DmitrySharabin:
I moved the PRs in the sections they belong to (from my perspective). Please take a look.

facelessuser:
I think I'm generally okay with the consolidations here, but I do believe that unique bugs need to be called out specifically for those experiencing issues and want to know the issue has been resolved. So with that said, I think https://github.com/color-js/color.js/pull/575 should be called out under fixes as it resolves the issue noted here: https://github.com/color-js/color.js/issues/553, which is a real issue that could affect users.

LeaVerou:
> I think I'm generally okay with the consolidations here, but I do believe that unique bugs need to be called out specifically for those experiencing issues and want to know the issue has been resolved. So with that said, I think #575 should be called out under fixes as it resolves the issue noted here: #553, which is a real issue that could affect users.

Agreed, under a "Bugfixes" section

LeaVerou (changes requested):
Thanks! Almost there, see feedback by @facelessuser

DmitrySharabin:
@facelessuser @LeaVerou Thank you so much for your review. I tried to address your feedback. Could you please take another look?

LeaVerou (approved):
LGTM though let's wait a day or two in case anyone else sees issues

lloydk (approved):
LGTM

facelessuser (approved):

facelessuser:
LGTM

PR #608, opened 2024-11-13
[oklch] Rename `Oklch` → `OKLCh`

DmitrySharabin:
The same name is already used in the test suite, so I decided to go with it.

### DIFF
--- a/src/spaces/oklch.js
+++ b/src/spaces/oklch.js
@@ -4,7 +4,7 @@ import lch from "./lch.js";
 
 export default new ColorSpace({
 	id: "oklch",
-	name: "Oklch",
+	name: "OKLCh",
 	coords: {
 		l: {
 			refRange: [0, 1],


LeaVerou (approved):

LeaVerou:
Thx!

lloydk:
The name was changed from `OKLCh` to `Oklch` in https://github.com/color-js/color.js/pull/300. Is there a reason we're reverting the change?

facelessuser:
I think it was changed in #300 to reflect how it is named in CSS. So I am surprised to see this get renamed here as well. With that said, the creator mentions it as OkLCh, not OKLCh (see [here](https://bottosson.github.io/posts/colorpicker/#okhsv)). It is quite common to mention LCh in this manner. I think, because this was meant to match how CSS refers to the space, it should not have been reverted as how the author mentioned it had already been taken into consideration IIRC when the change was originally made.

LeaVerou:
Happy to go with the version the creator uses — which IMO wins over what CSS is using. @DmitrySharabin could you make the change please? Thx!

DmitrySharabin:
> Happy to go with the version the creator uses — which IMO wins over what CSS is using. @DmitrySharabin could you make the change please? Thx!

Done!

Issue #609, opened 2024-11-30
8.7 million is rather larger than 3 million


svgeesus:
Colorjs.io [says](https://colorjs.io/):

> Over [3 million total npm downloads](https://limonte.dev/total-npm-downloads/?package=colorjs.io)!

This remains true, although the link currently says 

> [colorjs.io](https://npmjs.com/package/colorjs.io) was downloaded [8,788,840](https://npm-stat.com/charts.html?package=colorjs.io) times.

I suggest we update the text to say Over 8 million.

LeaVerou:
I just updated it to say over 9 (since it's currently at 8.9 so it will only take 1-2 more days to cross 9 given we're at 500k installs per week)

Ideally we'd want to find an automated way to do this (maybe even a build step that scrapes https://limonte.dev/total-npm-downloads/?package=colorjs.io )

MysteryBlokHed:
This could probably be done with GitHub workflows and the `https://api.npmjs.org/downloads/` endpoint. I can look into writing a workflow sometime tomorrow

PR #610, opened 2024-12-02
[ci] Add workflow to auto-update download count in README

MysteryBlokHed:
Briefly discussed in #609

This workflow runs a bash script to pull the latest download count from npm, and update the README.md number if the new value is larger.

Currently, it's set up to run daily.

### DIFF
--- a/.github/workflows/download_count.yaml
+++ b/.github/workflows/download_count.yaml
@@ -22,7 +22,7 @@ jobs:
           downloads=$(curl https://api.npmjs.org/downloads/point/1970-01-01:$(date -d '+1 day' +"%Y-%m-%d")/colorjs.io | jq -r .downloads)
           echo "Latest download count: $downloads"
           # into millions
-          downloads=$((downloads / 1000000))
+          downloads=$(( (downloads + 500000) / 1000000 ))
           # try to find current download count in readme
           readme=$(<README.md)
           if [[ "$readme" =~ ([0-9]+)" million total npm downloads" ]]; then
--- a/.github/workflows/download_count.yaml
+++ b/.github/workflows/download_count.yaml
@@ -1,0 +1,47 @@
+name: Update Download Count
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  update-downloads:
+    name: Update Download Count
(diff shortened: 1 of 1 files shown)


.github/workflows/download_count.yaml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get download count
+        run: |
+          # check downloads
+          downloads=$(curl https://api.npmjs.org/downloads/point/1970-01-01:$(date -d '+1 day' +"%Y-%m-%d")/colorjs.io | jq -r .downloads)
+          echo "Latest download count: $downloads"
+          # into millions
+          downloads=$((downloads / 1000000))
```
LeaVerou, line 25: Does this also round? I'm not very familiar with these and can't tell where the number of downloads is rounded.

MysteryBlokHed: This code will just truncate the number instead of rounding to the nearest million (eg. 5,999,999 downloads would become be 5 million). I can make it round instead, if preferred

LeaVerou: Truncating is probably fine, rounding might be better if easy.

MysteryBlokHed: I just changed it to round instead of truncate, but I've also just realized that the README says "_over_ 10 million total downloads," so it might be a bit misleading to round? I'm not sure how important that is, but figured it's worth mentioning

MysteryBlokHed: @LeaVerou Any updates?

LeaVerou: At the rate of increase, I think it's fine if it says "over" and rounds, unless we run it pretty frequently. No strong opinion though.

LeaVerou (approved):

MysteryBlokHed:
The `workflow_dispatch` trigger should already add a button somewhere to run it manually. To run it less frequently we can adjust the crontab here:

<https://github.com/color-js/color.js/blob/a5e94f9e2bea0ae18e11aace9d778a4dca35ef7e/.github/workflows/download_count.yaml#L5-L6>

e.g. `0 0 * * 1` to run every Monday at midnight instead of every day at midnight

PR #619, opened 2024-12-25
Fix parsing coords that have a minimum range < 0

lloydk:
The acescc color space has coords with a range of `[-0.358, 1.468]`.

The `percentageRange` function of the `Type` class was returning a value of `[-1, 1]` which caused coords specified as percentages to be calculated incorrectly.

The `percentageRange` function will now return `[0, 1]` if the Type has `coordMeta` with a `range` property.


lloydk:
With this fix all tests are green.

LeaVerou:
Thx! Could we add a changelog entry too? Otherwise this would become a blocker for the upcoming release.

lloydk:
> Thx! Could we add a changelog entry too? Otherwise this would become a blocker for the upcoming release.

done

LeaVerou (approved):

LeaVerou:
Thanks!

PR #622, opened 2024-12-26
Fix cam16 and hct type errors

lloydk:
Add additional types, @ts-expect-error comments and some minor code changes to fix all typescript errors in the cam16 and hct modules.

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -205,3 +205,38 @@ export interface DeltasReturn {
 	coords: [number, number, number];
 	alpha: number;
 }
+
+// cam16.js
+export type LightnessOrBrightness = { J: number; Q?: never; } | { J?: never; Q: number; };
+export type ChromaOrColorfulnessOrSaturation = { C: number; M?: never; s?: never; } | { C?: never; M: number; s?: never; } | { C?: never; M?: never; s: number; };
+export type HueOrHueQuadrature = { h: number; H?: never; } | { h?: never; H: number; };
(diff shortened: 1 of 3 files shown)


LeaVerou (approved):

LeaVerou:
LGTM though I'm not the best person to review type defs.

src/types.d.ts
```diff
@@ -205,3 +205,38 @@ export interface DeltasReturn {
...
+	c: number;
+	nc: number;
+	fl: number;
+	flRoot: number;
+	n: number;
+	z: number;
+	nbb: number;
+	ncb: number;
+	dRgb: number[];
+	dRgbInv: number[];
```
MysteryBlokHed, line 240:
```suggestion
	dRgb: [number, number, number];
	dRgbInv: [number, number, number];
```

I think these could be narrowed to `[number, number, number]` since they map over a 3-length array:

<https://github.com/color-js/color.js/blob/148af07d1af66f5c9b22d2e50ef5a2eee6a77cc7/src/spaces/cam16.js#L145-L146>

<https://github.com/color-js/color.js/blob/148af07d1af66f5c9b22d2e50ef5a2eee6a77cc7/src/spaces/cam16.js#L175-L180>

MysteryBlokHed (changes requested):
Just one change and everything else LGTM

src/types.d.ts
lloydk: done

MysteryBlokHed (approved):

PR #625, opened 2024-12-27
chore: upgrade to eslint 9.x

43081j:
Upgrades ESLint and moves to the new flat config style.

Also fixes up some newly found lint errors/warnings and enables TSESLint's strict config.

I was roaming around the repo so figured I'd do this while I'm in there. if you don't want it, all good just let me know and we can close this 👍 

### Notes

- Locally, for whatever reason, `npm run build:ts` fails with a _lot_ of errors. is that known to be broken? (in `main`)
- Do we need our own full blown config still? maybe we could use the recommended configs from ESLint, stylistic and tseslint
- could we introduce a formatter like prettier or dprint, and drop stylistic?


notebook/color-notebook.js
```diff
@@ -202,7 +203,7 @@ export default class Notebook {
 		}
 		catch (e) {
 			// Syntax error
-			var error = e;
```
43081j, line 205: I feel like there may have been a reason this used `var`. let me know if we cant make this change

LeaVerou:
Hi there, sorry this slipped through the cracks. Looks like it's now in need of rebasing. Is the change still current?

LeaVerou:
Thanks! If it wouldn't be too much hassle, would you mind separating the Prettier run from the substantive changes to the config? Currently the commit log will display all of them under the commit "chore: upgrade to eslint 9.x". Fine to just separate into two commits in this PR, then we can just rebase instead of squashing when merging.

LeaVerou (approved):
Thanks again! LGTM but someone who has worked more with types like @MysteryBlokHed should also review this.

eslint.config.js
```diff
@@ -0,0 +1,352 @@
+import eslintConfig from "@eslint/js";
+import { configs as tseslintConfigs } from "typescript-eslint";
+import * as globals from "globals";
+
+export default [
+	{
+		ignores: ["!.*", "assets/js/prism.js", "api/", "/dist/", "docs/", "types/src/"],
+	},
```
MysteryBlokHed: It doesn't seem to me like these ignores are properly affecting the entire ESLint configuration. When I run `npm run eslint` locally, I get almost 2,000 errors from the `dist` directory:

```text
> colorjs.io@0.6.0 eslint
> eslint .


/Users/adam/Git/color.js/dist/color-fn.cjs
   677:13  error  'name' is already defined                                                                                                                                       no-redeclare
  1496:3   error  Include a description after the "@ts-expect-error" directive to explain why the @ts-expect-error is necessary. The description must be 10 characters or longer  @typescript-eslint/ban-ts-comment
  1619:4   error  Include a description after the "@ts-expect-error" directive to explain why the @ts-expect-error is necessary. The description must be 10 characters or longer  @typescript-eslint/ban-ts-comment
  7083:1   error  Expected an assignment or function call and instead saw an expression                                                                                           @typescript-eslint/no-unused-expressions
  7084:1   error  Expected an assignment or function call and instead saw an expression                                                                                           @typescript-eslint/no-unused-expressions
  7085:1   error  Expected an assignment or function call and instead saw an expression                                                                                           

(etc)

✖ 1975 problems (1945 errors, 30 warnings)
```

43081j: Very strange, I do not, and do have a dist directory. 

This shouldn't be happening, I'll have a look into it

43081j: should be sorted 👍

MysteryBlokHed (approved):
LGTM!

svgeesus:
Please remove `package-lock.json` which is no longer in this repo. After that, this PR can be merged.

MysteryBlokHed (approved):
Changes look good to me, thank you!

PR #626, opened 2024-12-30
[style] Use Prettier

MysteryBlokHed:
Adds Prettier to format code for the project.

Since some Prettier rules conflict with eslint preferences, the lint script is currently set to run `eslint --fix` right after Prettier. I've tried to look for Prettier plugins to customize it to the way our code is formatted, but I couldn't find anything.

Because of these differences, the `prettier --check` command will always return warnings, so we wouldn't be able to enforce Prettier through CI that way. The best idea I have to check formatting with CI would be to just run `prettier --write; eslint --fix` and make sure that it doesn't produce any unstaged changes.

Also, I've quickly gone through the code and added some `// prettier-ignore` comments to maintain custom formatting (eg. spacing for matrices), but lmk if I've missed anything 😄


LeaVerou (changes requested):
So yeah, rather than calling ESLint fix after Prettier, let's remove ESLint (or dramatically restrict it to only the things that Prettier doesn't touch at all) in this PR and see how we're doing. 

E.g. it looks like the space before the opening paren is kept, but I don't know if that's ESLint or our Prettier config.
Same with stroustrup for braces.

MysteryBlokHed:
> is there any way we could use comments that don’t depend on any particular tooling?

Afaik there isn't any sort of tooling-agnostic ignore comment, so we'd just need to replace the comments if ever we wanted to change formatters. It's a bit annoying, but a global find+replace isn't the worst.

> So yeah, rather than calling ESLint fix after Prettier, let's remove ESLint (or dramatically restrict it to only the things that Prettier doesn't touch at all) in this PR and see how we're doing.

I just pushed a commit that has the code formatted without running `eslint --fix`, so the diff can be used to see how the code looks with/without ESLint: <https://github.com/color-js/color.js/commit/672c9a1e8ce0d42389f696fdc303aed893e54790>

LeaVerou:
Are there any Prettier plugins that could help reduce the difference? 

Googling a bit everything that comes up seems to imply that Prettier adds a space between function paren by default (e.g. https://prettier.io/blog/2020/03/21/2.0.0.html#always-add-a-space-after-the-function-keyword-3903httpsgithubcomprettierprettierpull3903-by-j-f1httpsgithubcomj-f1-josephfrazierhttpsgithubcomjosephfrazier-sosukesuzukihttpsgithubcomsosukesuzuki-thorn0httpsgithubcomthorn0-7516httpsgithubcomprettierprettierpull7516-by-bakkothttpsgithubcombakkot ) so I'm puzzled why it doesn't here.

And there's a plugin for brace style: https://www.npmjs.com/package/prettier-plugin-brace-style

MysteryBlokHed:
> Googling a bit everything that comes up seems to imply that Prettier adds a space between function paren by default

I think that's just for anonymous functions:

```javascript
//                       v space here
const identity = function (value) {
  return value;
};
//               v no space here
function identity(value) {
  return value;
}
```

It doesn't seem to me like there are any plugins currently available to add the space for regular functions.

> And there's a plugin for brace style: <https://www.npmjs.com/package/prettier-plugin-brace-style>

Added in the latest commit. Here's a new link to compare with/without `eslint --fix`: <https://github.com/color-js/color.js/compare/8e00024c7fe62344e1de63fc083e792e67161b99...use-prettier>

LeaVerou:
Oof. I just went down a rabbit hole investigating building a Prettier plugin for this 😅

What if we keep ESLint only for rules that a) can be enforced automatically b) are completely unambiguous (no exceptions) c) are NOT already enforced by Prettier?
Which may well end up being just this one rule.

Then, we add a workspace-specific project config that:
1. Enables Prettier (for folks that have it disabled by default)
2. Enables format on save
3. Any formatting action calls Prettier and then ESLint --fix.

What do you think?

MysteryBlokHed:
Seems good to me. Would we still want to keep ESLint for any non-formatting-related rules,  though? Its main purpose is less about code style and moreso about preventing pitfalls.

src/CATs.js
```diff
@@ -52,11 +53,12 @@ export function adapt (W1, W2, id = "Bradford") {
 	// console.log({scale});
 
 	let scaled_cone_M = multiplyMatrices(scale, method.toCone_M);
-	let adapt_M	= multiplyMatrices(method.fromCone_M, scaled_cone_M);
+	let adapt_M = multiplyMatrices(method.fromCone_M, scaled_cone_M);
 	// console.log({scaled_cone_M, adapt_M});
 	return adapt_M;
 }
 
+// prettier-ignore
```
LeaVerou, line 61: Should we just ignore the whole file?

LeaVerou:
> Seems good to me. Would we still want to keep ESLint for any non-formatting-related rules, though? Its main purpose is less about code style and moreso about preventing pitfalls.

If there are pitfalls it prevents that Prettier doesn't, we could. What are these?

MysteryBlokHed:
> If there are pitfalls it prevents that Prettier doesn't, we could. What are these?

Most of our ESLint config seems to be for this sort of rule. Here's a section of it:

<https://github.com/color-js/color.js/blob/db0479df4a4614eb8e37b1c2bbaef552e2e4077b/.eslintrc.cjs#L26-L43>

Only a few of the rules are formatting-specific. I can go through and remove those that are (except for the bracket spacing one we talked about)

LeaVerou:
Personally, I've failed to write the most basic of Prettier plugins. @DmitrySharabin is investigating this now.

DmitrySharabin:
> Personally, I've failed to write the most basic of Prettier plugins. @DmitrySharabin is investigating this now.

I made some progress. I just need Lea to look at the changes I made first.

Do I get it right that we don't consider using ESLint as a formatter? When I work on something in this repo, ESLint does an excellent job formatting my code on save. Don't get me wrong; I have nothing against Prettier (I use it on my projects, too); I'm just curious. 🙂

MysteryBlokHed:
> Do I get it right that we don't consider using ESLint as a formatter?

I think we just don't want two tools with overlapping functionality. So Prettier should be just formatting, and ESLint should only be code quality checks

LeaVerou:
> > Personally, I've failed to write the most basic of Prettier plugins. @DmitrySharabin is investigating this now.
> 
> I made some progress. I just need Lea to look at the changes I made first.
> 
> Do I get it right that we don't consider using ESLint as a formatter? When I work on something in this repo, ESLint does an excellent job formatting my code on save. Don't get me wrong; I have nothing against Prettier (I use it on my projects, too); I'm just curious. 🙂

Unfortunately, there's a pretty serious bug with this approach: https://github.com/LeaVerou/prettier-plugin-space-before-function-paren/issues/2

But we're making progress!!

MysteryBlokHed:
> Unfortunately, there's a pretty serious bug with this approach: https://github.com/LeaVerou/prettier-plugin-space-before-function-paren/issues/2

FYI this repo seems to be private, I can't access it

LeaVerou:
Ah, that's probably for the best given how bad this bug is 

> > Unfortunately, there's a pretty serious bug with this approach: [LeaVerou/prettier-plugin-space-before-function-paren#2](https://github.com/LeaVerou/prettier-plugin-space-before-function-paren/issues/2)
> 
> FYI this repo seems to be private, I can't access it

Just made it public: https://github.com/LeaVerou/prettier-plugin-space-before-function-paren

LeaVerou:
Okay, @DmitrySharabin made huge progress on this over the weekend! Could you try this out and see how it works? https://www.npmjs.com/package/prettier-plugin-space-before-function-paren

MysteryBlokHed:
It seems like the plugin doesn't work with comments. I tried running it on this source file:

```javascript
function someFunc () {}

// Function call
someFunc();
```

And got this error:

```sh
$ npx prettier --version
3.4.2
$ npx prettier demo.js
demo.js
[error] demo.js: Error: Comment "Function call" was not printed. Please report this error!
[error]     at ensureAllCommentsPrinted (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/index.mjs:19189:13)
[error]     at printAstToDoc (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/index.mjs:19986:3)
[error]     at async coreFormat (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/index.mjs:20393:14)
[error]     at async formatWithCursor (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/index.mjs:20598:14)
[error]     at async formatFiles (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/internal/cli.mjs:3235:18)
[error]     at async main (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/internal/cli.mjs:3887:5)
[error]     at async Module.run (file:///home/adam/Git/prettier-plugin-space-before-function-paren/node_modules/prettier/internal/cli.mjs:3833:5)
```

DmitrySharabin:
> It seems like the plugin doesn't work with comments.

Thank you so much for pointing this out. It turns out plugins should handle comments separately (who might have thought 🤦‍♂️). I have a [fix](https://github.com/LeaVerou/prettier-plugin-space-before-function-paren/pull/5), but first, I'd like @LeaVerou to review it.

MysteryBlokHed:
> @MysteryBlokHed, I published a new version of the plugin with the fix. Could you please verify that the issue with comments is gone? 🤞

Just pulled the latest version of the repo and it seems to work now!

LeaVerou:
Whee! let's try this again? Just published to npm!

MysteryBlokHed:
Alright, new problem...it seems like adding this plugin undoes the Stroustrup bracket style: 353398dbf50eb67ed5db8e171360d92e9ae2778b

I tried flipping the order of the plugins in the Prettier config file, but that just causes the inverse problem—the bracket style stays, but the spacing before parentheses does not.

DmitrySharabin:
> Alright, new problem...it seems like adding this plugin undoes the Stroustrup bracket style: [353398d](https://github.com/color-js/color.js/commit/353398dbf50eb67ed5db8e171360d92e9ae2778b)

I suspected something like this might happen. 😭 I’ll try to fix it. Thanks for trying things out.

DmitrySharabin:
> > Alright, new problem...it seems like adding this plugin undoes the Stroustrup bracket style: [353398d](https://github.com/color-js/color.js/commit/353398dbf50eb67ed5db8e171360d92e9ae2778b)
> 
> I suspected something like this might happen. 😭 I’ll try to fix it. Thanks for trying things out.

I could find a fix; we don't need to change anything in our plugin. Due to Prettier limitations [1], to use our plugin with other plugins formatting JS code, we need to install the [prettier-plugin-merge](https://github.com/ony3000/prettier-plugin-merge) plugin and **add it to the end** of the `plugins` array. This plugin will be used last and preserve changes made by the previous plugins.

I wrote some tests to check if it works. It does. The order of the plugins doesn't matter (except `prettier-plugin-merge` should always be last).

I'll add a note about this to the plugin's README.

[1] If two or more plugins are configured to format a particular language, Prettier will only use the last of those plugins.

MysteryBlokHed:
Alright, the formatting works now. But it seems like there are a few cases where the bracket spacing doesn't get applied: <https://github.com/color-js/color.js/compare/8e00024c7fe62344e1de63fc083e792e67161b99...use-prettier>

DmitrySharabin:
> But it seems like there are a few cases where the bracket spacing doesn't get applied

That means we need to improve our plugin. 😁

> I see differences in the typedef files

Those are TypeScript files, and we don't support them yet. So, it's probably high time to add the support.

> As well as places where functions are defined in an object

Those cases definitely should be covered.

Thanks for spotting the issues! I'll work on them.

DmitrySharabin:
@MysteryBlokHed, do you mind trying this again? Just published on npm.

> Not sure whether there are any other cases I've missed

Since this is our first plugin, finding new issues on every iteration is our fate. 😅 Thanks for helping with this!

MysteryBlokHed:
Strangely, there still seems to be one case that the extension misses 🤔: <https://github.com/color-js/color.js/compare/8e00024c7fe62344e1de63fc083e792e67161b99...use-prettier#diff-621ae9fa7ab8ed0f19aab11f0312aefb2bc9d66072e46f65f14555b3393dba1b>

DmitrySharabin:
> I'm honestly a bit confused by this, since it seems to work on everything else—even elsewhere within this file. I would assume that these are `FunctionDeclaration` nodes, like any other function, which look like they should be handled by the plugin.

That's indeed weird. I added corresponding tests, and they all passed without any changes to the current version of the plugin. That said, I added support for some more cases, including the one you mentioned. Could you please try it (the new release is already on npm)?

I already tested it with your branch, and it seems to work. However, I noticed one thing I don't know how to explain. When I ran `npm run prettier -- --write`, the `color.d.ts` file didn't change. But if I format the same file using the Prettier extension or running `prettier **/*.{js,ts,cjs} --write`, the expected changes appeared. Is there a chance that it's somehow related to the `--cache` parameter? 🤷‍♂️

MysteryBlokHed:
> But if I format the same file using the Prettier extension or running `prettier **/*.{js,ts,cjs} --write`, the expected changes appeared.

For some reason I can't replicate this behaviour locally. Running `--write` (with and without `--cache`) will correct other formatting issues in the file, but it doesn't touch the spacing in that one area.

I guess it's an issue with my local setup? It's definitely strange

LeaVerou:
If this is only about these two changes, I think it's fine. Is this good to merge then? (once we address the conflicts)

MysteryBlokHed:
> If this is only about these two changes, I think it's fine. Is this good to merge then? (once we address the conflicts)

Should be. I'll take a look at the conflicts once I'm home

MysteryBlokHed:
I've tried running Prettier with and without `--cache` on two different computers, but nothing gets it to format the namespace block properly.

I've just added `// prettier-ignore`  comments to the functions there for the time being.

Assuming that that's an acceptable solution for now, I think this is ready to merge?

MysteryBlokHed:
I've gone ahead and removed the bracket spacing ESLint rule we had.

If we're migrating fully to Prettier, is it safe to assume we also want to get rid of all the other style-only rules?

MysteryBlokHed:
> If we're migrating fully to Prettier, is it safe to assume we also want to get rid of all the other style-only rules?

Thoughts?

LeaVerou:
> > If we're migrating fully to Prettier, is it safe to assume we also want to get rid of all the other style-only rules?
> 
> Thoughts?

I'm 99% sure the answer to this should be "yes", but what are the specific styles we have that are NOT covered by Prettier?

MysteryBlokHed:
> I'm 99% sure the answer to this should be "yes", but what are the specific styles we have that are NOT covered by Prettier?

Looking through the `@stylistic` plugins in the ESLint config, I think that the only place where it diverges from Prettier is for indentation in IIFE's:

<https://github.com/color-js/color.js/blob/48e06585528d9fd4e00d654bb1687a0e96800945/.eslintrc.cjs#L201-L203>

Here's an example of the difference (note how GitHub Actions constantly says "expected indentation of 0"):

<https://github.com/color-js/color.js/compare/8e00024c7fe62344e1de63fc083e792e67161b99...use-prettier#diff-d2d4c1b75aa80167dc1f2fd2212e1661f3107aca953e62c10aca2bbfe5547bb8>

Is this a style choice that would be important for us to keep?

LeaVerou:
I can't quite parse that diff, but I don't think we have that many IIFEs that this actually matters, do we?

MysteryBlokHed:
> I can't quite parse that diff

Yeah the GitHub Actions prompts sort of get in the way, but the only change is the level of indentation.

> I don't think we have that many IIFEs that this actually matters, do we?

Seems like it's just that one file, so probably shouldn't matter. I assume that the ESLint config was only set up that way to match the existing formatting in that file.

LeaVerou:
So is this ready to merge?

MysteryBlokHed:
Should be now

LeaVerou (approved):
Alright, let's do this!

Issue #613, opened 2024-12-05
jz**** color spaces have NaN where all other spaces have numbers


sidewayss:
https://codepen.io/sidewayss/pen/JoPGdQy

Open the CodePen console and you'll see a list of space ids followed by coordinates. The only `NaN` coordinates are in the jzczhz and jzazbz spaces.

This is a bad set of coordinates to begin with, but the fact that it's just these two spaces makes me think that it's a bug.

This issue is based on this Q&A: #612

svgeesus:
Sorry not to have noticed this issue until now.

I agree that this is a bug.

Looking again at the code for Jzazbz, I confirmed that the matrix values and the equations are the same as in the [Sadfar et al paper](https://opg.optica.org/oe/fulltext.cfm?uri=oe-25-13-15131&id=368272). They also seem to be the same as in the [Jacob Rus observable](https://observablehq.com/@jrus/jzazbz).

I agree with @facelessuser that Absolute XYZ should not be clamped to positive values, and that the code for Jzazbz should use the sign-preserving `spow()` function. I made those changes in a color.js branch.

Looking now at the conversion from `sRGB white`, in the (modified) colorjs I get `color(jzazbz 0.22207 -0.00016 -0.00012)` and `color(jzczhz 0.22207 0.0002 216.08)`. The chroma for pure white strikes me as way too high.

The same conversion in Coloraide:

```python
chris@SuperNomad:~$ python
Python 3.12.3 (main, Jan 17 2025, 18:03:48) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from coloraide import Color
>>> w = Color('white').convert('jzazbz')
>>> print(w)
color(jzazbz 0.22207 -0.00016 -0.00012 / 1)
>>> w2 = w.convert('jzczhz')
>>> print(w2)
color(jzczhz 0.22207 0.0002 216.08 / 1)
```

The observable notebook (after changing the color to be converted to whte, and the media white luminance to 203) gives

```
user_jab = 
    
  Array(3) [
  0: 0.22206398164774332
  1: -0.00016398964837013863
  2: -0.00010233622023381184
]
```

So these are all giving the same values (to 5 significant figures) and yet the chroma for white looks suspiciously high.

If a simple monochromatic color is giving bad values, I certainly don't trust conversions of wildly high chroma values.

svgeesus:
It looks like the brighter the neutral color, the greater the chroma shift, and all at a consistent hue:

[...]

PR #629, opened 2025-01-11
Fix jzczhz check for achromatic colors

lloydk:
The default `ε` (0.00001) calculated for `jzczhz` is to low for achromatic colors. The highest `az` or `bz` value for an achromatic color is `-0.00023625807422922307` for [color(rec2100-pq 1 1 1)](https://apps.colorjs.io/convert/?color=color(rec2100-pq%201%201%201)&precision=4).

The `ColorSpace` class has been modified to allow setting `ε` when constructing a `ColorSpace` and the `ε` for `jzczhz` has been set to `0.0002363`.

This new `ε` ensures that all achromatic colors up to the brightest white in `req2100-pq` have a null hue when converted to `jzczhz`.

However, with the new `ε` there are some non-achromatic colors near black in `rec2100-pq` that will be considered to be achromatic when converted to `jzczhz`. I think this is ok but I wanted to point them out.
```
// The non zero values in this list are either 1 / 255, 2 / 255 or 3 / 255.
color(rec2100-pq 0 0 0.00392156862745098)
color(rec2100-pq 0 0 0.00784313725490196)
color(rec2100-pq 0 0.00392156862745098 0)
color(rec2100-pq 0 0.00392156862745098 0.00392156862745098)
color(rec2100-pq 0 0.00392156862745098 0.00784313725490196)
color(rec2100-pq 0 0.00784313725490196 0)
color(rec2100-pq 0 0.00784313725490196 0.00392156862745098)
color(rec2100-pq 0 0.00784313725490196 0.00784313725490196)
color(rec2100-pq 0.00392156862745098 0 0)
color(rec2100-pq 0.00392156862745098 0 0.00392156862745098)
color(rec2100-pq 0.00392156862745098 0 0.00784313725490196)
color(rec2100-pq 0.00392156862745098 0.00392156862745098 0)
color(rec2100-pq 0.00392156862745098 0.00392156862745098 0.00784313725490196)
color(rec2100-pq 0.00392156862745098 0.00784313725490196 0)
color(rec2100-pq 0.00392156862745098 0.00784313725490196 0.00392156862745098)
color(rec2100-pq 0.00392156862745098 0.00784313725490196 0.00784313725490196)
color(rec2100-pq 0.00784313725490196 0 0)
color(rec2100-pq 0.00784313725490196 0 0.00392156862745098)
color(rec2100-pq 0.00784313725490196 0 0.00784313725490196)
color(rec2100-pq 0.00784313725490196 0.00392156862745098 0)
color(rec2100-pq 0.00784313725490196 0.00392156862745098 0.00392156862745098)
color(rec2100-pq 0.00784313725490196 0.00392156862745098 0.00784313725490196)
color(rec2100-pq 0.00784313725490196 0.00784313725490196 0)
color(rec2100-pq 0.00784313725490196 0.00784313725490196 0.00392156862745098)
color(rec2100-pq 0.00784313725490196 0.00784313725490196 0.011764705882352941)
color(rec2100-pq 0.00784313725490196 0.011764705882352941 0.00784313725490196)
color(rec2100-pq 0.00784313725490196 0.011764705882352941 0.011764705882352941)
color(rec2100-pq 0.011764705882352941 0.00784313725490196 0.00392156862745098)
color(rec2100-pq 0.011764705882352941 0.00784313725490196 0.00784313725490196)
color(rec2100-pq 0.011764705882352941 0.00784313725490196 0.011764705882352941)
color(rec2100-pq 0.011764705882352941 0.011764705882352941 0.00784313725490196)
```

I rounded the `ε`, I can change it to the exact number if that's a more appropriate value.

### DIFF
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -78,6 +78,7 @@ export interface SpaceOptions {
 	formats?: Record<string, Format> | undefined;
 	gamutSpace?: "self" | string | ColorSpace | null | undefined;
 	aliases?: string[] | undefined;
+	ε?: number | undefined;
 }
 
 export type Ref =
@@ -137,6 +138,7 @@ export default class ColorSpace {
 	referred?: string | undefined;
 	white: White;
 	gamutSpace: ColorSpace;
+	ε?: number | undefined;
 
 	from (color: ColorTypes): Coords;
 	from (space: string | ColorSpace, coords: Coords): Coords;
--- a/src/ColorSpace.js
+++ b/src/ColorSpace.js
@@ -85,6 +85,7 @@ export default class ColorSpace {
 
 		// Other stuff
 		this.referred = options.referred;
+		this.ε = options.ε;
 
 		// Compute ancestors and store them, since they will never change
 		Object.defineProperty(this, "path", {
--- a/src/spaces/jzczhz.js
+++ b/src/spaces/jzczhz.js
@@ -24,4 +24,5 @@ export default new ColorSpace({
 	base: Jzazbz,
 	fromBase: lch.fromBase,
 	toBase: lch.toBase,
+	ε: 0.0002363,
 });


facelessuser:
> However, with the new ε there are some non-achromatic colors near black in rec2100-pq that will be considered to be achromatic when converted to jzczhz. I think this is ok but I wanted to point them out.

Yeah, that means you lose those colors, unfortunately.

This kind of thing crops up with a number of color spaces, especially ones that allow you to adjust the viewing environment and assuming the eye is not fully adapted, such as CAM16, ZCAM, HCT, RLAB, etc. In such spaces, the achromatic colors, those with zero chroma, don't necessarily align with what one would traditionally consider neutral in sRGB due to the influence of the adapting luminance and background luminance.

I think the question becomes is achromatic what is considered neutral in the given space (with the given viewing environment and when chroma is zero) or is it strictly what we would consider achromatic in a given RGB space?

In Jzabz's case, chroma doesn't get anywhere near as big as it does in CAM16 and HCT near white, so you can probably get away with giving it a more relaxed threshold (like is done in this PR) at the expense of some color. Maybe the small loss of colors won't matter?

In spaces like CAM16 and HCT, such an approach simply can't work. You need something more complicated as the traditional achromatic line is more dynamic in these spaces.

I'm not really sure what is preferable. I experimented with such things earlier, and at least in my own color library, I decided to make "achromatic" relative to the space, and when chroma is near zero in those spaces, it is considered achromatic. I think I preferred not losing colors if it could be helped. I at least think that with JzCzhz, if this is the direction we want to go, the impact might not be too large.

svgeesus:
See also https://github.com/color-js/color.js/issues/613#issuecomment-2666872133

svgeesus:
None of these typescript lint errors relate to the files I touched in this PR

MysteryBlokHed (approved):

svgeesus:
So it is now clear that ε should be chosen to encompass the narrow range of values close to the central axis where the hue angle is ill-conditioned and replaced with `none`.

In many spaces (but not all) that is also where achromatic colors converted from some RGB space where R=G=B fall. For the spaces where that is not true, that is because of color appearance; the eye is not fully adapted and so neutrals gradually diverge from the central axis as the neutral gets brighter. But the hue angle is not ill-conditioned and the ε should not be made artificially large so as to encompass these colors.

svgeesus:
So it is now clear that ε should be chosen to encompass the narrow range of values close to the central axis where the hue angle is ill-conditioned and replaced with `none`.

In many spaces (but not all) that is also where achromatic colors converted from some RGB space where R=G=B fall. For the spaces where that is not true, that is because of color appearance; the eye is not fully adapted and so neutrals gradually diverge from the central axis as the neutral gets brighter. But the hue angle is not ill-conditioned and the ε should not be made artificially large so as to encompass these colors.

svgeesus:
Closing without merging

PR #631, opened 2025-01-12
[multiplyMatrices] Update implementation to pass all tests

epsilonError:
Updates implementation and passes all tests for `multiplyMatrices`.

I attempted to make the smallest changes per commit to pass skipped tests. Once a test was passed, I didn't skip it again (resulting in one regression that was also fixed).

Note: To see that an error is no longer thrown by multiplying empty arrays, you'll have to comment out the
```
check: (..._args) => {
    return true; // Treat these tests as passed
},
```
block in the Incorrect Data test when running `htest`.

@lloydk, let me know how this looks to you. And I can also fit some refactoring into this pull request before it's merged, if needed. 

I wanted to prioritize minimal changes to pass the tests which would make a future refactor easier to understand before I started it.


facelessuser:
I think the output generally looks okay to me. The results behave like I would expect. I do find the output in the HTML a bit frustrating as it strips all the brackets (which isn't something you are doing or can control). I would expect the multiplication of `[[3]]` and `[[4]]` to be `[[12]]`, but the output strips all of that context when it is displayed and it looks like the scalar `12`. So I assume it is being calculated with the appropriate dimensions, but I would have to checkout the commit locally to be sure. But generally, I think it looks good to me.

facelessuser (approved):

MysteryBlokHed:
Is there anything blocking this from being merged?

facelessuser:
I have no objections to anything in the review, though I'm the only one who seems to have reviewed it.

LeaVerou (approved):

PR #633, opened 2025-02-08
Added isValid function

caoimghgin:


LeaVerou (changes requested):
Hi there @caoimghgin, thanks for submitting a PR!

I think a helper for this makes sense, but we absolutely cannot replicate all color parsing logic in it. The helper should most likely just call `parse()` and catch any thrown errors, i.e. a much, much smaller thing 😁 

And since it would be calling the existing `parse()` machinery, it won't need that many tests either.

I'm also not sure what `apps/tracer.js` is for?

PR #640, opened 2025-03-12
Ensure Prophoto and Rec2020 spaces handle negative values

lloydk:


src/spaces/prophoto.js
```diff
@@ -11,9 +11,27 @@ export default new RGBColorSpace({
 	base: ProPhotoLinear,
 	toBase (RGB) {
 		// Transfer curve is gamma 1.8 with a small linear portion
-		return RGB.map(v => (v < Et2 ? v / 16 : v ** 1.8));
+		return RGB.map(v => {
+			let sign = v < 0 ? -1 : 1;
+			let abs = v * sign;
```
LeaVerou, line 16: Just to make sure I follow, we're not using `Math.sign()` for performance, right?

facelessuser:
I thought/assumed this was already handled 😅. Generally, LGTM.

facelessuser (approved):

src/spaces/prophoto.js
lloydk: This is how negative values were handled in the `sRGB` space so I used the same technique to keep things consistent. I don't know if there would be that much of a performance difference.

svgeesus (approved):

PR #642, opened 2025-03-14
Fix types and type errors

lloydk:
- changed Color* types to allow null for alpha
- added `findFormat` and equals functions to the `ColorSpace` type
- added types for the `id` and `alpha` properties of the `Format` class to fix type errors in the `parse` function
- minor code changes to eliminate type errors in the `parse` function.
- added a `ParseMeta` interface to `types.d.ts` for the metadata stored on Color objects
- fixed type errors in the `serialize` function related to `parseMeta`.
- Other miscellaneous fixes to correct types and type errors.

### DIFF
--- a/scripts/generate-space-accessor-types.js
+++ b/scripts/generate-space-accessor-types.js
@@ -20,14 +20,14 @@ for (const spaceModule in spaces) {
 let template = `// DO NOT EDIT. This file is generated with \`npm run build:space-accessors\`.
 
 /** Proxy used for space accessors */
-export type SpaceAccessor = Record<string, number> & number[];
+export type SpaceAccessor = Record<string, number | null> & (number | null)[];
 
 declare class SpaceAccessors {`;
 for (const spaceId of Array.from(spaceIds).sort()) {
 	template += `\n\t${spaceId}: SpaceAccessor;`;
 }
 for (const coord of Array.from(coords).sort()) {
-	template += `\n\t${coord}: number;`;
+	template += `\n\t${coord}: number | null;`;
 }
 template += `
 }
--- a/src/ColorSpace.d.ts
+++ b/src/ColorSpace.d.ts
@@ -27,7 +27,7 @@ export interface Format {
 	alpha?: boolean | undefined;
 	test?: ((str: string) => boolean) | undefined;
 	/** Function to parse a string into a color */
-	parse?: ((str: string) => ColorConstructor) | undefined;
+	parse?: ((str: string) => ColorConstructor | undefined | null) | undefined;
 	/**
 	 * Serialize coordinates and an alpha channel into a string.
 	 * Must be defined for a format to support serialization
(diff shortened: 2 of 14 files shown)


kleinfreund:
What I most commonly run into with regards to types in colorjs.io are `SpaceAccessor` channels missing `null` leading to code that accesses things like `color.hsl.h` (which has type `number`) and not taking into account that `h` can be null. The following

https://github.com/color-js/color.js/blob/a236bc3d04958731994e18758b663bc2e1eb4beb/scripts/generate-space-accessor-types.js#L30

should use `number | null` as far as I can tell.

lloydk:
> What I most commonly run into with regards to types in colorjs.io are `SpaceAccessor` channels missing `null` leading to code that accesses things like `color.hsl.h` (which has type `number`) and not taking into account that `h` can be null. The following
> 
> https://github.com/color-js/color.js/blob/a236bc3d04958731994e18758b663bc2e1eb4beb/scripts/generate-space-accessor-types.js#L30
> 
> should use `number | null` as far as I can tell.

Good catch, I've changed the space accessor types to allow `null` for `color.hsl.h` and `color.h`.

LeaVerou:
> Good catch, I've changed the space accessor types to allow `null` for `color.hsl.h` and `color.h`.

All channels can be null, not just those. Even alpha :)

lloydk:
> > Good catch, I've changed the space accessor types to allow `null` for `color.hsl.h` and `color.h`.
> 
> All channels can be null, not just those. Even alpha :)

I thought the space accessors only returned coords?
```
> c = new Color("red")
> Color {parseMeta: {…}, space: RGBColorSpace, coords: Array(3), alpha: 1, …}
> c.oklch
> Proxy(Array) {0: 0.6279553639214311, 1: 0.2576833038053608, 2: 29.23388027962784}
> c.oklch.alpha
> undefined
```

LeaVerou:
> I thought the space accessors only returned coords?

Yup, the "even alpha" was a general comment. But every coord can be `null`, not just hue.

kleinfreund:
That alpha can be `null` was a bit more surprising to me, but it's definitely the case currently. At the very least through this code:

https://github.com/color-js/color.js/blob/a236bc3d04958731994e18758b663bc2e1eb4beb/src/parse.js#L160

LeaVerou:
Yup, hue and chroma can be `null` as a result of other operations, but `null` is a valid value that can be passed manually too, just like `none` in CSS. It represents the absence of a value, and can be quite useful for interpolation.

facelessuser:
I guess maybe chroma could be undefined if the space's transformation algorithm just can't handle chroma with zero lightness...but often in those cases it is still just assumed chroma is then zero.

MysteryBlokHed (approved):
Typings changes LGTM

There are a lot of errors related to the `Format` type (or more specifically, errors with `string | Format`) throughout `src/serialize.js`, but those were there before this PR. Probably worth fixing, but doesn't necessarily need to be right now.

lloydk:
> 
> There are a lot of errors related to the `Format` type (or more specifically, errors with `string | Format`) throughout `src/serialize.js`, but those were there before this PR. Probably worth fixing, but doesn't necessarily need to be right now.

I made a half hearted attempt at fixing those `Format` related errors but ended up with some failing tests. I'll try again in a separate PR on a day when I've had a good night's sleep and I'm fully caffeinated.

MysteryBlokHed:
Is there anything else that needs to be done for this PR before merging?

PR #644, opened 2025-03-28
Set CAM16 and HCT hues to null when chroma is near zero

facelessuser:


facelessuser:
I thought that LCh was using chroma as the reference change for achromatic, but I looked a bit closer today while doing this, and it appears it uses Lab reference range, which is confusing to me.

While I think LCh and LCh-like spaces should use chroma as reference range for determining achromatic threshold, not the difference between a - b, for CAM16 and HCT we have no choice and they must use chroma/colorfulness as the polar coordinate system is the base, so we use chroma/colorfulness to determine the achromatic threshold.

svgeesus (approved):

Issue #648, opened 2025-06-05
parse.js throws unnecessary error for hls()


robinscholz:
When parsing colors like `hsl(0 0 0 / 100%)` or `hsl(0, 0%, 0%, 1)` the following check [here](https://github.com/color-js/color.js/blob/89bfe6e7fce55c05fb15f13d8acae4750d8a5c72/src/parse.js#L114) is wrongly throwing an error.

This is the relevant code:

```
if (coords.length !== coordCount) {
			throw new TypeError(
				`Expected ${coordCount} coordinates for ${space.id} in ${env.str}), got ${coords.length}`,
			);
		}
```

Running `0.6.0-alpha.1`

LeaVerou:
Is it a regression? Does it still throw a bug in 0.5.0?

facelessuser:
Well, it worked prior to 6.x, so it seems like a regression.

facelessuser:
Admittedly, I don't know at what point it broke, whether we made parsing more rigid, or what. Locally, 0.5.2 parsed the two problematic cases fine. That doesn't mean that the behavior cannot be logically explained in 0.6.x and we just need to update the format spec for HSL, but either by mistake or on purpose, form the user perspective, it seems like a regression.

kleinfreund:
Part of the parsing issue is that for the hsla format definition (and possibly others), the `alpha` property is `true` which causes the expression `env.parsed.args.pop()` to remove the third coord. So I think colorjs.io assumes `hsla` is (1) always legacy format and (2) always has an alpha component, but that's not what the spec says. Without looking too much into it, it doesn't look like the current format definition could capture that HSL for its modern formats can use both `hsl(` and `hsla(` with `<percentage> | <number>` on saturation and lightness while the legacy formats can use both `hsl(` and `hsla(` with `<percentage>` on saturation and lightness. Alpha seems to always be optional, too.

PR #650, opened 2025-06-09
fix: hsl parsing not being spec-compliant

### LINKED ISSUE #428: Ensure HSL is spec compliant
Additional formats need to be added to the `hsl` color space in order to be [spec compliant](https://drafts.csswg.org/css-color/#the-hsl-notation).

For example: `hsla(240deg 100% 50%)` fails to parse.

kleinfreund:
## Changes

Add the `<number>` type to the saturation and lightness coordinates (i.e. changing them from `<percentage>` to `<percentage> | <number>`) for the HSL color space to match the specification (see https://www.w3.org/TR/css-color-4/#funcdef-hsl).

Make parsing independent from the `format.alpha` configuration of color spaces and instead rely on the parser to identify whether there is a "last alpha" element in a color string using functional notation. This way, colors that are being parsed with a format that sets `format.alpha` to `true` (for serialization purposes) can still omit the alpha value which is valid in both legacy and modern syntax for hsl/hsla colors.

Closes #428, #648.

## Notes

- **Important**: Currently (both before and with this PR's changes), HSL colors in the legacy functional format parse fine with `<number>` values for saturation/lightness. Two skipped tests are added for this.
- This pull request contains a significant change to the parsing logic that needs careful review. Specifically, there's one assumption I'm making about the structure of color strings in functional notation: I assume that all color strings in functional notation **except `color()`** that have four "elements" contain the alpha channel in the last element. Now I know this is certainly not true for relative CSS colors (e.g. `lch(from blue calc(l + 20) c h)`), but I'm not sure there's ever a chance for the current parsing logic to handle such colors. In other words, I assume that the parsing logic I changed deals exclusively with `<color-function>` excluding specifically `color()`.
- The exclusion of `color()` is possibly something that can be addressed, but I wanted to keep the scope of this PR clear. From a quick look at https://www.w3.org/TR/css-color-4/#predefined, a similar assumption as above could be made for `color()`: Perhaps one can assume that all `color()` strings that have five "elements" contain the alpha channel in the last element.
- I used the currently failing examples from the issues #428 and #648 in the tests, feel free to suggest more cases that are useful to test.
- Only related on a meta level: I found working with the test suite a little tricky because logging expressions was difficult. For one, logging synchronously would often lead to nothing being logged. I eventually resorted to using the following timeout-based technique which isn't ideal:

  ```js
  let messages = [JSON.stringify({ lastAlpha: env.parsed.lastAlpha, args: env.parsed.args })]
  global.setTimeout(() => console.log(messages.join(':')), 15)
  ```

  Furthermore, I found myself commenting out a lot of the tests to reduce the noise while logging. htest now has support for `skip` which is useful, but while working on code, I regularly use `only` flags for _only_ running tests within a test suite flagged as such without the need to disable them otherwise.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -99,7 +99,7 @@ export default function parse (str, options) {
 
 		let alpha = 1;
 
-		if (format.alpha === true || env.parsed.lastAlpha) {
+		if (env.parsed.lastAlpha) {
 			alpha = env.parsed.args.pop();
 
 			if (meta) {
@@ -245,11 +245,17 @@ export function parseFunction (str) {
 		let args = [];
 		let argMeta = [];
 		let lastAlpha = false;
+		let name = parts[1].toLowerCase();
 
 		let separators = parts[2].replace(regex.singleArgument, ($0, rawArg) => {
 			let { value, meta } = parseArgument(rawArg);
 
-			if ($0.startsWith("/")) {
+			if (
+				// If there's a slash here, it's modern syntax
+				$0.startsWith("/")
+				// If there's still elements to process after there's already 3 in `args` (and the we're not dealing with "color()"), it's likely to be a legacy color like "hsl(0, 0%, 0%, 0.5)"
+				|| (name !== "color" && args.length === 3)
+			) {
 				// It's alpha
 				lastAlpha = true;
 			}
@@ -260,7 +266,7 @@ export function parseFunction (str) {
 		});
 
 		return {
-			name: parts[1].toLowerCase(),
+			name,
 			args,
 			argMeta,
 			lastAlpha,
--- a/test/parse.js
+++ b/test/parse.js
@@ -565,6 +565,60 @@ const tests = {
 					args: "hsl(230.6 179.7% 37.56% / 1)",
 					expect: { spaceId: "hsl", coords: [230.6, 179.7, 37.56], alpha: 1 },
 				},
+				{
+					name: "legacy syntax, <percentage> saturation/lightness, no alpha (#428, #648)",
+					args: ["hsl(0, 0%, 0%)"],
+					expect: { spaceId: "hsl", coords: [0, 0, 0], alpha: 1 },
+				},
+				{
+					name: "legacy syntax, <percentage> saturation/lightness, alpha (#428, #648)",
+					args: ["hsl(0, 0%, 0%, 0.5)"],
+					expect: { spaceId: "hsl", coords: [0, 0, 0], alpha: 0.5 },
+				},
+				{
+					name: "legacy syntax, <number> saturation/lightness, no alpha (#428, #648)",
+					args: ["hsl(0, 0, 0)"],
+					throws: true,
+					// TODO: #428. This currently parses successfully but shouldn't because the legacy syntax doesn't allow `<number>` for saturation or lightness.
+					skip: true,
+				},
+				{
+					name: "legacy syntax, <number> saturation/lightness, alpha (#428, #648)",
+					args: ["hsl(0, 0, 0, 0.5)"],
+					throws: true,
+					// TODO: #428. This currently parses successfully but shouldn't because the legacy syntax doesn't allow `<number>` for saturation or lightness.
+					skip: true,
+				},
+				{
+					name: "modern syntax, <percentage> saturation/lightness, no alpha (#428, #648)",
+					args: ["hsl(0 50% 25%)"],
+					expect: { spaceId: "hsl", coords: [0, 50, 25], alpha: 1 },
+				},
+				{
+					name: "modern syntax, <percentage> saturation/lightness, alpha (#428, #648)",
+					args: ["hsl(0 50% 25% / 50%)"],
+					expect: { spaceId: "hsl", coords: [0, 50, 25], alpha: 0.5 },
+				},
+				{
+					name: "modern syntax, <number> saturation/lightness, no alpha (#428, #648)",
+					args: ["hsl(0 50 25)"],
+					expect: { spaceId: "hsl", coords: [0, 50, 25], alpha: 1 },
+				},
+				{
+					name: "modern syntax, <number> saturation/lightness, alpha (#428, #648)",
+					args: ["hsl(0 50 25 / 50%)"],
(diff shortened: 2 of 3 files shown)


src/parse.js
```diff
@@ -245,11 +245,17 @@ export function parseFunction (str) {
...
 
 		let separators = parts[2].replace(regex.singleArgument, ($0, rawArg) => {
 			let { value, meta } = parseArgument(rawArg);
 
-			if ($0.startsWith("/")) {
+			if (
+				// If there's a slash here, it's modern syntax
+				$0.startsWith("/")
+				// If there's still elements to process after there's already 3 in `args` (and the we're not dealing with "color()"), it's likely to be a legacy color like "hsl(0, 0%, 0%, 0.5)"
+				|| (name !== "color" && args.length === 3)
```
facelessuser, line 257: I would note that if you ever implement a color space with more than 3 color channels, this will be a problem. It is true that most color spaces are usually 3 channels, and it may be that there is no intention of implementing color spaces with more than 3 channels, but I feel I need to at least point it out.

LeaVerou:
> Only related on a meta level: I found working with the test suite a little tricky because logging expressions was difficult. For one, logging synchronously would often lead to nothing being logged.

Just saw this. Typically, expressions are logged in the collapsed tree of the test, to avoid noise (i.e. you need to expand them). Is that the case or are they not logged in at all? I’m thinking that perhaps this should have been opt-in rather than just wrapping all console methods. 🤔 

cc @DmitrySharabin (not urgent)

src/parse.js
facelessuser: If moving forward in the current state is preferred, I would consider at least going ahead and generating a technical debt issue describing the issue to be solved in the future. I agree though that it might not be an immediate concern.

kleinfreund: Yes, that's an obvious flaw in my approach. It's an assumption that currently holds true, but that can break with time.

The first thing I thought was to store the number of possible channels on the relevant format. This also opens up the possibility for users to hook into individual parsing rules _per_ channel for custom formats should that be something you're interested in.

Since in `parseFunction`, we don't have knowledge of the `Format` instance just yet, that would be a bit more heavy-handed to change. I can create a ticket for this as it would at least be good to have an answer for the project's intention with this.

kleinfreund:
> Just saw this. Typically, expressions are logged in the collapsed tree of the test, to avoid noise (i.e. you need to expand them). Is that the case or are they not logged in at all? I’m thinking that perhaps this should have been opt-in rather than just wrapping all console methods. 🤔

Having now tried this on a macOS computer, I can confirm that logs happen. I originally worked on this using Windows where I'm fairly sure I tried this and it didn't work. I think what didn't work was expanding the leave nodes in the interactive report. So the timeout-based logging allowed me to "break" out of the reported nodes, so to speak. For what it's worth, I didn't thoroughly look into htest's options for whether perhaps either the "console log wrapping" or interactive reporter can be disabled. I reckon either would've "revealed" the logs to me.

src/parse.js
kleinfreund: Draft for such a follow up issue should this get merged:

## Remove parser dependance on all CSS colors in functional notation (except color()) having 4 elements (3 color coordinates; 1 alpha)

Currently, the colorjs.io parsing logic has a dependance on color strings in functional notation (except `color()`) having 4 elements: 3 color coordinates and 1 alpha. This was introduced in #650 and discussed in https://github.com/color-js/color.js/pull/650#discussion_r2138733791.

This dependance would lead to errors if support for color strings in functional notation with a different number of elements is introduced.

One such case is `color()` which has 5 elements (https://www.w3.org/TR/css-color-4/#predefined). This also illustrates that it's plausible that this becomes a problem as the elements in between the parentheses aren't constrained to be just color coordinates and an alpha value. For `color()`, the first element is the color space parameter and effectively shifts the color and alpha elements "to the right".

To remove this dependance, each color space format could hold a definition of each element in the corresponding functional notation. The introduction of a new format and/or color space would then have to only take care of providing an accurate definition and not need to mess with the general parsing logic.

LeaVerou: One thing to note is that even though the *official* color spaces don't use more than 3 channels yet, an author-defined color space can use any number of channels.

I’m not sure storing the number of channels on the format is the right way to go, since it could be variable.
It definitely seems like the right API shape is not obvious, so in the interest of progress, I'm still leaning towards merging this and filing the issue above.

Perhaps you could improve it somewhat by adding a check for commas.

svgeesus, line 257: Yeah not super keen on that assumption, but the nature of legacy functional syntax makes it hard to avoid.
Having read the discussion I am in favor of merging it and logging the issue, as suggested.

test/parse.js
```diff
@@ -565,6 +565,60 @@ const tests = {
 					args: "hsl(230.6 179.7% 37.56% / 1)",
 					expect: { spaceId: "hsl", coords: [230.6, 179.7, 37.56], alpha: 1 },
 				},
+				{
```
svgeesus, line 568: Thanks both for the tests and the notes on missing/invalid tests!

svgeesus (approved):

src/parse.js
lloydk: There are a lot of places in the types where we assume only 3 coordinates, the most notable being the `Coords` type.

```
export type Coords = [number | null, number | null, number | null];
```

So we'll need to make a number of changes to the types to support an arbitrary number of coordinates.

kleinfreund: > I’m not sure storing the number of channels on the format is the right way to go, since it could be variable.

I was thinking more along the lines of capturing the same information as the specs show for the syntax of colors: https://www.w3.org/TR/css-color-4/#funcdef-hsl

This would allow you to accurately describe where which kind of element can be expected and also whether its optional. In any case, that'll be part of the discussion in the follow up issue.


> Perhaps you could contain the scope of the change somewhat by adding a check for commas.

Would you mind clarifying that for me? Checking `parts[2]` for commas like `name !== "color" && args.length === 3 && parts[2].includes(",")`?

kleinfreund: Opened the technical debt tracking issue for the dependance on number of channels here: https://github.com/color-js/color.js/issues/651

PR #654, opened 2025-06-18
Fix return types of the darken/lighten functions on the Color class

### LINKED ISSUE #652: types are incorrect for Color.darken and Color.lighten
Current type definitions for Color state that darken() and lighten() return PlainColorObject, when they actually return the Color object.

This can be demonstrated with the following code:

```typescript
import Color, { type PlainColorObject } from "colorjs.io";

function isColor(value: unknown): value is Color {
  return value instanceof Color;
}

export const green = new Color("green");
console.log("green", isColor(green));

export const darkGreen = green.darken(0.5);
console.log("darkGreen", isColor(darkGreen));

export const lightGreen = green.lighten(0.5);
console.log("lightGreen", isColor(lightGreen));

export const plainColorObject: PlainColorObject = {
  space: Color.spaces.sRGB,
  coords: [0, 0, 0],
  alpha: 1,
};
console.log("plainColorObject", isColor(plainColorObject));
```

When compiled and run, the output is:

```
green true
darkGreen true
lightGreen true
plainColorObject false
```

This shows that `darkGreen` and `lightGreen` are Color objects.

While the generated .d.ts shows that both have the PlainColorObject type

```typescript
import Color, { type PlainColorObject } from "colorjs.io";
export declare const green: Color;
export declare const darkGreen: PlainColorObject;
export declare const lightGreen: PlainColorObject;
export declare const plainColorObject: PlainColorObject;
```

lloydk:
Changes the types of the `lighten` and `darken` instance and static functions to return `Color` instead of `PlainColorObject`.

Also fixes an issue with the static functions returning a `PlainColorObject` instead of a `Color` when anything but a `Color` object was passed to them. For example:

`new Color("red").lighten()` would return a `PlainColorObject` instead of a `Color` object.

Fixes #652

### DIFF
--- a/src/color.d.ts
+++ b/src/color.d.ts
@@ -284,8 +284,8 @@ declare class Color extends SpaceAccessors implements PlainColorObject {
 	// variations
 	lighten: ToColorPrototype<typeof lighten>;
 	darken: ToColorPrototype<typeof darken>;
-	static lighten: typeof lighten;
-	static darken: typeof darken;
+	static lighten: ToColorNamespace<typeof lighten>;
+	static darken: ToColorNamespace<typeof darken>;
 }
 
 export default Color;
--- a/src/variations.js
+++ b/src/variations.js
@@ -27,3 +27,9 @@ export function darken (color, amount = 0.25) {
 	let /** @type {Ref} */ lightness = [space, "l"];
 	return set(color, lightness, l => l * (1 - amount));
 }
+
+/** @type {"color"} */
+lighten.returns = "color";
+
+/** @type {"color"} */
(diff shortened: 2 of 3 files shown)


LeaVerou:
Approved because it's an improvement over the current state, but ideally we'd want a structure where functions like `lighten()` and `darken()` can be modular. Otherwise how can plugins define correct types?
I wonder if we can use an `interface` for `Color` objects, since that can be extended in other files. 🤔

lloydk:
> @lloydk should I merge this? Do you want to?

Forgot about this one.

PR #663, opened 2025-08-06
Improve procedural API for all color spaces use case, rel #661

### LINKED ISSUE #661: Using the procedural API + importing all color spaces is painful
I think a common use case for the procedural API is a bit of a hybrid: still be able to access all known color spaces, but support tree-shaking in terms of the actual functionality.

We don't currently handle this well at all. There is no export in the package that gives you all color spaces. They are exported as individual exports, together with everything else, so one needs to loop over all exports to see which ones are a color space and register them.

As a first step, we should have an export that includes all color spaces, so they can be registered in one go.

I.e. instead of that:
```js
export *                                  from "./spaces/index-fn.js";
```

Perhaps do:

```js
import * as spaces from "./spaces/index-fn.js";
export *                                  from "./spaces/index-fn.js";
export {spaces};
```

Then, to make it even easier, we could overload `ColorSpace.register()` to take an array of color spaces.

Then all authors would need to do is:

```js
import { spaces, ColorSpace } from "colorjs.io/fn"
ColorSpace.register(spaces);
```

We could also follow a similar pattern for other extensibility points, e.g. deltaE methods.

LeaVerou:
- Single export for all color spaces in `/fn` so they can be registered in one go
- `/spaces` export that exports all color spaces and automatically registers them
- `/src` export as a catch-all because we are bound to have missed more things

Doesn't close #661, but addresses the most pressing pain point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
 				"default": "./dist/color-fn.cjs"
 			}
 		},
-		"./dist/*": "./dist/*"
+		"./spaces": "./src/spaces/index.js",
+		"./dist/*": "./dist/*",
+		"./src/*": "./src/*"
 	},
 	"typesVersions": {
 		"*": {
--- a/src/index-fn.js
+++ b/src/index-fn.js
@@ -33,6 +33,10 @@ export *                                  from "./deltaE/index.js";
 export { default as deltaEMethods }       from "./deltaE/index.js";
 export *                                  from "./variations.js";
 export { mix, steps, range, isRange }     from "./interpolation.js";
+
+// Export all color spaces as a single object as well so they can be registered in one go (#661)
+import * as spaces from "./spaces/index-fn.js";
+export { spaces }
 export *                                  from "./spaces/index-fn.js";
 
 // Type re-exports


facelessuser (approved):

lloydk:
I tried importing something from `src` in a Typescript project and the type for the function couldn't be found.

```
import { toPrecision } from "colorjs.io/src/util.js";
```

results in this error:

```
Could not find a declaration file for module 'colorjs.io/src/util.js ...'
```

Not sure what needs to be added to `package.json` so that types are picked up for the `./src/*` export.

LeaVerou:
> I tried importing something from `src` in a Typescript project and the type for the function couldn't be found.
> 
> ```
> import { toPrecision } from "colorjs.io/src/util.js";
> ```
> 
> results in this error:
> 
> ```
> Could not find a declaration file for module 'colorjs.io/src/util.js ...'
> ```
> 
> Not sure what needs to be added to `package.json` so that types are picked up for the `./src/*` export.

Can you file an issue so we're not tracking this in a merged PR?

Issue #636, opened 2025-02-26
Add support for relative colors


uhafner:
In the Jenkins project we are using oklch colors. In order to use such colors in Echarts I wanted to use your library to convert these oklch colors to rgb colors (see https://github.com/jenkinsci/echarts-api-plugin/pull/383). This works quite well for absolute colors like `oklch(60% 0.2601 12.28)`, but for relative colors like `oklch(from oklch(55% 0.2308 256.91deg) 2% 0.075 h)` the call `new Color(color).to("sRGB").toString({format: "hex"});` fails with an exception. 

Is there a plan to support such [relative colors](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors) as well?

LeaVerou:
Hi there @uhafner!
This is definitely in scope, though no idea when we might get to it.
Meanwhile, as a workaround, you can use the color to set the `color` on a dummy element and then read back the computed style.

LeaVerou:
> I'm curious on exactly how much 'relative' stuff Color.js plans to support. Does it plan on implementing its own support and logic for things like `calc()`? Or does it plan to keep relative support fairly simple?

That’s the million dollar question 😁 I don't have an answer right now. It's unclear how much utility that provides without variables, and the implementation effort involved is pretty significant. I'd be inclined to support this as a plugin, rather than part of core Color.js.

svgeesus:
> I'm curious on exactly how much 'relative' stuff Color.js plans to support. Does it plan on implementing its own support and logic for things like `calc()`? Or does it plan to keep relative support fairly simple?

Most real-world usage is indeed using `calc()`, `var()` and custom properties. That seems to be out of scope here.

However I would have no objection to supporting the plain nested relative colors. But I feel it could be a lot of work

Issue #316, opened 2023-06-06
Would be good to have an isValid helper


fatih-erikli:
Currently I call `Color(string)` and treat it as valid if an exception was not thrown. It would be good to have an `isValid` function exported in the module.

LeaVerou:
I wonder if we could feed two birds with one scone by making `Color.get()` / `getColor()` not error (see #518 ). Then you can just check if the return value `=== null`.

gandhiShepard:
Exceptions are not thrown for all invalid colors however. If using `oklch("z" 100 230)` the returned `.toString` value would be `oklch(NaN 100 230)`. Obviously, "z" is not acceptable input and 100 is out of range as well, yet, no error is returned. I think a `Color.isValid({format: 'oklch'})` would be a better api. I think it would be better to validate a known input, than having a validation that checked the input against ALL known color spaces.

ColorJS is amazing btw! Much resepct.

LeaVerou:
> Exceptions are not thrown for all invalid colors however. If using `oklch("z" 100 230)` the returned `.toString` value would be `oklch(NaN 100 230)`. Obviously, "z" is not acceptable input and 100 is out of range as well, yet, no error is returned.

Interesting point. `NaN` is a valid coordinate value (corresponds to the CSS `calc(NaN)`), but perhaps we should not be treating all non-numeric strings as equivalent to `NaN`.

> I think a `Color.isValid({format: 'oklch'})` would be a better api. I think it would be better to validate a known input, than having a validation that checked the input against ALL known color spaces.

Functional formats zero in pretty quickly to the color space.
It would be helpful to know the use cases that prompted this. If it's just to facilitate fault tolerance, see #518, I wonder if that would be an acceptable solution?

> ColorJS is amazing btw! Much resepct.

Thanks! <3

gandhiShepard:
I was trying to validate colors strings within an input while a user types.

bakert:
I'm curious why `new Color("rgb(1000, 'echo', null)")` throws an exception but `new Color("srgb", [-1000, "echo", null])` does not. If I have these values separately rather than as a CSS string, how can I get the exception throwing behavior?

(My hacky idea of `new Color(new Color("srgb", [-1000, "echo", null]).toString())` doesn't work because the invalid values are set to 0 when creating the inner Color.)

bakert:
```
> new Color('rgb(-1, -1000, -1)').toString()
'rgb(0% 0% 0%)'
[...]

PR #664, opened 2025-08-06
Add `tryColor()`/`Color.try()` for graceful parsing and complex colors

LeaVerou:
- Optionally resolves complex CSS colors via the DOM, closes [#636](https://github.com/color-js/color.js/issues/636)
- Graceful parsing closes [#316](https://github.com/color-js/color.js/issues/316)

Still needs tests, at least for the graceful bit, but perhaps we can add them later. We can mock the DOM bit so that the tests run in Node. While looking at tests, I also noticed we don't seem to have any for `getColor()`/`Color.get()` either.
Though we may want to also have some client-side only tests to see how different browsers behave so we can describe what this can safely resolve in the docs.

### DIFF
--- a/src/color.d.ts
+++ b/src/color.d.ts
@@ -8,7 +8,7 @@ import hooks from "./hooks.js";
 import * as util from "./util.js";
 import ColorSpace, { Ref } from "./ColorSpace.js";
 import SpaceAccessors from "./space-coord-accessors.js";
-import { ToGamutOptions } from "./types.js";
+import { ToGamutOptions, ParseOptions, TryColorOptions } from "./types.js";
 
 import {
 	to as toFn,
@@ -153,10 +153,10 @@ declare class Color extends SpaceAccessors implements PlainColorObject {
 	constructor (space: string | ColorSpace, coords: Coords, alpha?: number | null);
 
 	// These signatures should always be the same as the constructor
-	static get (color: ColorTypes): Color;
+	static get (color: ColorTypes, options?: ParseOptions): Color;
 	static get (space: string | ColorSpace, coords: Coords, alpha: number | null): Color;
 
-	static try (color: ColorTypes): Color | null;
+	static try (color: ColorTypes, options?: TryColorOptions): Color | null;
 
 	static defineFunction (name: string, code: DefineFunctionHybrid): void;
 	static defineFunction (


src/tryColor.js
```diff
@@ -0,0 +1,77 @@
...
+				}
+			}
+			else {
+				// Still not resolved
+				error = {message: "Color value is a valid CSS color, but it could not be resolved :("};
+			}
+		}
+	}
+
+	if (error) {
```
DmitrySharabin, line 70: Why do we check for `error`? If I get everything correctly, it is always there if we couldn’t parse the color. We can only update it. What am I missing?

DmitrySharabin: Also, if there is no `error`, will the function return `undefined`? If so, it isn’t reflected in the function signature. Right?

LeaVerou: Oooh both good points, thx!

src/color.js
```diff
@@ -131,6 +132,24 @@ export default class Color {
 		return new Color(color, ...args);
 	}
 
+	/**
+	 * Like {@link get} but returns `null` if resolution fails instead of throwing an error.
+	 * Additionally, supports passing an element to resolve complex CSS colors through the DOM (slow).
+	 */
+	static try (color, ...args) {
```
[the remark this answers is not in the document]
LeaVerou: Yup, I was trying to avoid coupling their signatures more than is necessary, but perhaps that's too much. But since, as you point out `get()` is the same, I think we can deal with that later.

lloydk (changes requested):
Need to add the static `try()` method to `color.d.ts`.

If I pass an array of colors to `tryColor()`there doesn't seem to be a way get parsing metadata for each color, you'll get the metadata for the last color. I think `getColor()` has the same problem.

DmitrySharabin:
Oh, and one more thing. Shouldn't we start using `tryColor()` instead of `getColor()` in the `Color` constructor to avoid unnecessary throws?
https://github.com/color-js/color.js/blob/e70420e5c76961e8548802e0b3b89b99801717e5/src/color.js#L56

lloydk:
> > If I pass an array of colors to `tryColor()`there doesn't seem to be a way get parsing metadata for each color, you'll get the metadata for the last color. I think `getColor()` has the same problem.
> 
> Ooh, nice catch. Though given that `getColor()` has the same problem, I suggest solving them both together separately, after this PR. What do you think?

I think that's fine.

src/color.js
```diff
@@ -131,6 +132,25 @@ export default class Color {
 		return new Color(color, ...args);
 	}
 
+	/**
+	 * Get a color instance from the argument passed or `null` if resolution fails (instead of throwing an error).
+	 * Additionally, it supports passing an element to resolve complex CSS colors through the DOM (slow).
+	 * @see {@link tryColor} for more details
+	 */
+	static try (color, options) {
```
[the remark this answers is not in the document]
LeaVerou: `get()` is the same, so if we change this, we could change it in both.

lloydk (approved):
LGTM

PR #667, opened 2025-08-08
Update changelog

DmitrySharabin:
Includes the changes after `v0.0.6-alpha` is released until 2025-08-~08~10.


LeaVerou (changes requested):
While we probably want a combined changelog for the stable v0.6.0, I think we should separate by alpha while we're still doing pre-releases. Otherwise folks following the pre-releases don't know what's new.

LeaVerou (changes requested):
Nice work!

I don't think it's sustainable to maintain a separate v0.6.0 changelog that needs to be manually kept in sync. I'd just keep the alphas and combine them into a v0.6.0 when the stable is out.

PR #668, opened 2025-08-10
[types] Add types for `spaces` and `src` exports

### LINKED ISSUE #666: Types not found for ./src/* and ./spaces exports
I tried importing something from the new `./src*` export in a Typescript project and the type for the function couldn't be found.

```
import { toPrecision } from "colorjs.io/src/util.js";
```

results in this error:

```
Could not find a declaration file for module 'colorjs.io/src/util.js ...'
```

Same problem with the `./spaces` export. Not sure about the `./dist/*` export.

MysteryBlokHed:
Closes #666

This commit doesn't touch the exports for `dist`, it just fixes the types for `spaces` and `src`. I only specified the `import` property for each of the two exports, since the files use ESM exports and wouldn't work with `require`.

Example (from #666):

```typescript
import { toPrecision } from "colorjs.io/src/util.js";
```

Is now typed properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,19 @@
 				"default": "./dist/color-fn.cjs"
 			}
 		},
-		"./spaces": "./src/spaces/index.js",
-		"./dist/*": "./dist/*",
-		"./src/*": "./src/*"
+		"./spaces": {
+			"import": {
+				"types": "./types/src/spaces/index.d.ts",
+				"default": "./src/spaces/index.js"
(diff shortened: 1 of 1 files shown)


lloydk (approved):
LGTM

LeaVerou (approved):
Ooh nice, thanks! 

You'd expect TypeScript would be smart enough to figure this out (at least for `src`) but nooope…

Issue #669, opened 2025-08-28
Deploying changes to `color(rec2020 ...)`


svgeesus:
TLDR; Do we just make this backwards-incompatible change to remain in sync with Color 4, or continue to provide the old implementation as well (with some different name)?

An issue was raised on CSS Color 4 to say that we were using the wrong transfer function for the BT.2020 color space. Instead of the OETF in BT.2020 we should be using the EOTF in BT.1886, because we say that `rec2020` colorspace is display-referred.

 - https://github.com/w3c/csswg-drafts/issues/12574

I have prototyped this change in a fork of color.js and the changes are [clearly visible](https://github.com/w3c/csswg-drafts/issues/12574#issuecomment-3156578544)

Last week, CSS WG [resolved](https://github.com/w3c/csswg-drafts/issues/12574#issuecomment-3200792172) to make the change. CSS Color 4 has not been updated _yet_ but will be soon. WPT will also need to be updated, which will make all browsers fail [two of those five tests](https://wpt.fyi/results/css/css-color?label=master&label=experimental&aligned&q=2020) (tests on component values 0 or 1 are unaffected by a change in transfer function).

I can see four ways to make this change (on a scale from cautious to move-fast-break-things):

1. Announce an upcoming breaking change in the next color.js beta release notes, see if anyone complains, meanwhile do nothing for a while
2. Add a new `--rec2020-proposed` color space, leaving the old one as-is to not break anything
3. Change the existing `rec2020` but add a new color space called  `rec2020-oetf` or something, with the existing implementation
4. Just change the existing `rec2020`, because apart from tests and demos there is little real-world usage

facelessuser:
Yeah, EOTF in BT.1886 definitely has a different response.

<img width="2480" height="1048" alt="Image" src="https://github.com/user-attachments/assets/50c1c377-c4ec-4f9e-b212-483c693b49c4" />

PR #672, opened 2025-09-03
[spaces/rec2020] Use gamma 2.40 for display-referred rec2020 #669

svgeesus:
rename previous implementation as scene-referred --rec2020-oetf so people can still use the older implementation if they want

See

 - https://github.com/w3c/csswg-drafts/issues/12574#issuecomment-3156578544

### DIFF
--- a/data/modules.json
+++ b/data/modules.json
@@ -77,6 +77,13 @@
 			"description": "A very wide gamut RGB color space used in 4k and 8k UHDTV. It is currently difficult to make a screen which displays the entire REC.2020 gamut; current movies broadcast in this colorspace do not use the full gamut, although they do use a little more than P3."
 		},
 		{
+			"name": "REC.2020 Scene Referred",
+			"id": "--rec2020-oetf",
+			"dependencies": ["srgb"],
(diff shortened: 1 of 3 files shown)


facelessuser (commented):
As far as implementation, everything looks fine. But related Rec. 2020 tests would need to be updated as some are failing:

- Rec. 2020 <-> Linear Rec. 2020
- APCA contrast, sRGB gamut in other color spaces

facelessuser:
I don't see anything obviously wrong with gamut mapping or stringifying the output, so I suspect gamut mapping tests (and a few others) are broken due to calling some special htest check functions. This isn't an area I'm familiar with, but I guess we should open a separate issue to figure out what's going wrong there.

svgeesus:
> But related Rec. 2020 tests would need to be updated as some are failing:

Agreed, I will fix those

svgeesus:
One reason so few `rec2020` tests were failing is that an astonishing number use colors with either `0` or `1` coordinates which are insensitive to the actual transfer function.

facelessuser (approved):

facelessuser:
LGTM

PR #686, opened 2025-10-19
[types] Use `@import` instead of `@typedef`

### LINKED ISSUE #685: Use TypeScript 5.5's `@import` for JSDoc type imports
Whenever we needed to import types from another file with JSDoc, we've been using this syntax with `@typedef`:

<https://github.com/color-js/color.js/blob/83bd1f887ca398ef3845890e384434f9311c6c6b/src/tryColor.js#L4-L7>

TypeScript 5.5 introduced support for a new `@import` keyword, which would allow us to rewrite that example as:

```typescript
/**
 * @import { ColorTypes, PlainColorObject, TryColorOptions } from "./types.js";
 */
```

A comparison between the generated `.d.ts` outputs:

```typescript
// Before (@typedef)
export type ColorTypes = import("./types.js").ColorTypes;
export type PlainColorObject = import("./types.js").PlainColorObject;
export type TryColorOptions = import("./types.js").TryColorOptions;
// After (@import)
import type { ColorTypes } from "./types.js";
import type { TryColorOptions } from "./types.js";
import type { PlainColorObject } from "./types.js";
```

Using this syntax has a few advantages. Namely:

1. It avoids polluting the exports of many files, since the `@typedef` syntax would actually _re-export_ every imported type (as can be seen from the output above)
2. It is less verbose/easier to type than repeated `@typedef`s
3. It makes LSP functionality like "jump to definition" work properly. With `@typedef`, trying to jump to an imported type's definition would just go to the `@typedef` line, rather than the actual source

This change also _shouldn't_ cause color.js to be incompatible with consumers' older TypeScript versions, since everything type-related in the package gets compiled to `.d.ts` anyway. We'd just need to upgrade color.js's TypeScript dependency from our current pinned v5.4.5 to v5.5.0 or later.

See: <https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#the-jsdoc-import-tag>

MysteryBlokHed:
Closes #685

There are some files for which the `@typedef`s' re-exporting behavior was deliberate (e.g. everything in `index-fn`), while for most cases it was just a way of mimicking an import. For any files where the re-export behaviour was deliberate, the `@typedef`s have been left alone.

### DIFF
--- a/src/clone.js
+++ b/src/clone.js
@@ -1,4 +1,3 @@
-/** @import Color from "./color.js" */
 /** @import { Coords, PlainColorObject } from "./types.js" */
 
 /**
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -15,6 +15,8 @@ export {
 	White,
 	// Re-exported from src/CATs.d.ts
 	CAT,
+	// Re-exported from src/deltas.d.ts
+	DeltasReturn,
 	// Re-exported from src/display.d.ts
 	Display,
 	// Re-exported from src/interpolation.d.ts
--- a/src/to.js
+++ b/src/to.js
@@ -2,10 +2,7 @@ import getColor from "./getColor.js";
 import ColorSpace from "./ColorSpace.js";
 import toGamut from "./toGamut.js";
 
-/** @import { ColorTypes, PlainColorObject } from "./types.js" */
-
-// Type re-exports
-/** @typedef {import("./types.js").ToGamutOptions} ToGamutOptions */
+/** @import { ColorTypes, PlainColorObject, ToGamutOptions } from "./types.js" */
 
 /**
  * Convert to color space and return a new color


MysteryBlokHed:
The majority of the re-exported types here are types that would make more sense to define directly within the re-exporting files, rather than in `types.d.ts`. At some point, it might be worth going through and migrating any such type definitions from the `types.d.ts` file to `@typedef` directives directly within the source file. That way we'll both reduce duplicated exports, and make sure that the types are defined within the file to which they're actually relevant.

For example: any functions that take complex options as parameters currently have their options interfaces defined in `types.d.ts` and then re-export those, when it would probably make more sense to define them directly within the source file.

src/clone.js
```diff
@@ -1,7 +1,5 @@
-// Type "imports"
-/** @typedef {import("./color.js").default} Color */
-/** @typedef {import("./types.js").Coords} Coords */
-/** @typedef {import("./types.js").PlainColorObject} PlainColorObject */
+/** @import Color from "./color.js" */
```
lloydk: I don't think this import is necessary.

src/deltas.js
```diff
@@ -4,8 +4,9 @@ import to from "./to.js";
 import { adjust } from "./angles.js";
 import { isNone } from "./util.js";
 
-// Type "imports"
-/** @typedef {import("./types.js").ColorTypes} ColorTypes */
+/** @import { ColorTypes } from "./types.js" */
+
+// Type re-exports
 /** @typedef {import("./types.js").DeltasReturn} DeltasReturn */
```
lloydk: This type is exported in `index-fn.js` but it's missing from `index.d.ts`.

lloydk (approved):
LGTM

MysteryBlokHed:
@LeaVerou requesting review in case you want to look at the dependency version changes in `package.json`

AviVahl:
This PR completely broke the generated .d.ts files. See issue: https://github.com/color-js/color.js/issues/692

PR #693, opened 2025-11-17
First stab at fixing the primary nav on small viewports

DmitrySharabin:
### Summary

- Introduce a responsive navigation system that “moves” overflowed items into a hamburger menu. These changes make the header adapt to smaller viewports and
prevent wrapping/overflow of nav items.
- Tweak the corresponding CSS rules and header image sizing (so it doesn’t shrink to 0 width and entirely disappear).

After some experiments, I’ve decided to duplicate menu items in two places (instead of programmatically moving them) to avoid DOM manipulations as much as possible. For me, it seems a reasonable trade-off since our nav menu is not huge, and RO might fire quite often.

### DIFF
--- a/_includes/partials/_nav-links.njk
+++ b/_includes/partials/_nav-links.njk
@@ -1,16 +1,34 @@
-<a href="{{ page | relative }}/get/">Get<span class="wide"> Color.js</span></a>
-<div class="menu">
+<a href="{{ page | relative }}/get/" data-nav-item="1">Get<span class="wide"> Color.js</span></a>
+<div class="menu" data-nav-item="2">
 	<a href="{{ page | relative }}/docs/">Docs</a>
 	<ul>
 		{% include "./_docs-nav.njk" %}
 	</ul>
 </div>
-<a href="{{ page | relative }}/api/">API</a>
-<a href="{{ page | relative }}/notebook/">Play!</a>
-<a href="https://apps.colorjs.io/">Demos</a>
-<a href="https://elements.colorjs.io/">Elements</a>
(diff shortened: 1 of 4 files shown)


_includes/partials/_nav-links.njk
```diff
@@ -1,16 +1,34 @@
...
+		<li data-nav-item="0" hidden><a href="{{ page | relative }}/get/">Get Color.js</a></li>
+		<li data-nav-item="1" hidden><a href="{{ page | relative }}/docs/">Docs</a></li>
+		<li data-nav-item="2" hidden><a href="{{ page | relative }}/api/">API</a></li>
+		<li data-nav-item="3" hidden><a href="{{ page | relative }}/notebook/">Play!</a></li>
+		<li data-nav-item="4" hidden><a href="https://apps.colorjs.io/">Demos</a></li>
+		<li data-nav-item="5" hidden><a href="https://elements.colorjs.io/">Elements</a></li>
+		<li data-nav-item="6" hidden><a href="https://github.com/LeaVerou/color.js">GitHub</a></li>
+		<li data-nav-item="7" hidden><a href="https://discord.gg/K64FJBznq4">Discord</a></li>
+		<li data-nav-item="8" hidden><a href="https://opencollective.com/color">♡&nbsp;Sponsor</a></li>
+	</ul>
```
DmitrySharabin: As a TODO, I'd try to DRYify this part, but unfortunately, those items are not entirely identical to the nav items. So, for now, I have decided to leave it as is and revisit it in the next iteration.

PR #703, opened 2025-12-24
[types] Move ColorSpace types from `.d.ts` to JSDoc

MysteryBlokHed:
Interfaces and type aliases were moved to the `types.d.ts` file and imported/re-exported from `ColorSpace.js` as needed. The definitions for the class itself (properties and methods) now come from JSDoc comments. The `ColorSpace.d.ts` file has been removed.

### DIFF
--- a/src/getAll.js
+++ b/src/getAll.js
@@ -31,9 +31,10 @@ export default function getAll (color, options) {
 
 	let coords, space;
 
-	if (options?.space) {
-		space = ColorSpace.get(options, options.space);
+	try {
+		space = ColorSpace.get(options, options?.space);
 	}
+	catch {}
 
 	let precision = options?.precision;
 


MysteryBlokHed:
Note: there _is_ actually a change for the logic of one function here; namely, the `ColorSpace.get` static method. It seems to me that the original function here has a bug:

<https://github.com/color-js/color.js/blob/6ea537a03272bfb0967cff0f8870b282170ea16f/src/ColorSpace.js#L297-L304>

If a caller passes an empty string, then the function will return back that empty string, rather than either returning a space object or throwing an error. Compare this to e.g. the `Color.get` static method, where there isn't any check for falsiness:

<https://github.com/color-js/color.js/blob/6ea537a03272bfb0967cff0f8870b282170ea16f/src/color.js#L127-L133>

The only change I made is removing the check for `!space`:

```diff
 	static get (space, ...alternatives) {
-		if (!space || isInstance(space, this)) {
+		if (isInstance(space, this)) {
 			return space;
 		}
```

LeaVerou (changes requested):
Thanks for this!

My only comment is that if we're re-exporting `Format`, it's part of the API and it should have a good name. Calling constructors `XxxClass` is a bit of an antipattern (imagine a function called `fooFunction()`, like, we don't do that). So perhaps it's the TS `Format` that should change (`FormatType`?). Or, we can use a different name like `ColorFormat`, `StringFormat`, anything that describes what it does, rather than its syntax.

MysteryBlokHed:
> My only comment is that if we're re-exporting `Format`, it's part of the API and it should have a good name. Calling constructors `XxxClass` is a bit of an antipattern (imagine a function called `fooFunction()`, like, we don't do that).

Yeah, in this case I just did that to disambiguate between the `Format` interface and `Format` class within the `ColorSpace.js` file; the name `FormatClass` is never used outside of that file. It would probably make sense to just give the interface a unique name, though. Maybe `FormatObject` to match other -Object names we've used, like with `ColorObject`?

<https://github.com/color-js/color.js/blob/6ea537a03272bfb0967cff0f8870b282170ea16f/src/color.d.ts#L53-L64>

src/ColorSpace.js
```diff
@@ -146,8 +187,8 @@ export default class ColorSpace {
 
 	/**
 	 * Lookup a format in this color space
-	 * @param {string | object | Format} format - Format id if string. If object, it's converted to a `Format` object and returned.
-	 * @returns {Format}
+	 * @param {string | FormatObject | Format} format - Format id if string. If object, it's converted to a `Format` object and returned.
+	 * @returns {FormatObject | Format | null}
```
lloydk:
```suggestion
	 * @returns { Format | null}
```

src/ColorSpace.js
```diff
@@ -296,10 +397,13 @@ export default class ColorSpace {
...
 	 * Lookup ColorSpace object by name
-	 * @param {ColorSpace | string} name
+	 * @param {ColorSpace | string} space
+	 * @param {Array<ColorSpace | string>} alternatives
+	 * @returns {ColorSpace}
+	 * @throws {TypeError} If no matching color space is found
 	 */
 	static get (space, ...alternatives) {
-		if (!space || isInstance(space, this)) {
+		if (isInstance(space, this)) {
```
lloydk: This change breaks a bunch of tests as the `resolveCoord` method depends on the old behavior.

lloydk (changes requested):
Mostly LGTM, just a couple of issues.

MysteryBlokHed:
I've updated all the places I found that explicitly depend on the old behaviour by doing something explicit along these lines:

```javascript
let space = ColorSpace.get(/* ... */);
if (!space) {
    // ...
}
```

With that, I think this PR should be good?

Though note that it would need to be released with a minor version bump, since this is a breaking change.

lloydk (changes requested):

lloydk:
A bunch of tests are failing.

MysteryBlokHed:
Looks like the errors come from the suggested change to `getAll` in 29d0d90bc2374bc3c19d32dab891a26da2563ae0. I'll put back the try/catch for the moment

PR #709, opened 2026-01-15
[types] Move DOM lib inclusion to types definition file

### LINKED ISSUE #708: Explicitly include `lib=["dom"]` in types declaration.
I'm trying to use `colorjs.io` in a TypeScript project that does not have DOM types imported by default (mostly runs in node, not in a browser), and I run into a bunch of
```
node_modules/colorjs.io/types/src/types.d.ts:234:12 - error TS2304: Cannot find name 'Element'.

234  element?: Element;
```

I'm not really keen on adding `lib = ["dom"]` to my top-level compiler options (it breaks many stuff all over the place), a workaround is to do it on a file by file basis, which is a bit annoying (or to liberally nuke `skipLibCheck` everywhere, something else I'd rather avoid). Maybe I could also selectively import what I need (since this dependency only appears in `trycolor.js` which I cannot use out of a browser anyway), but I didn't manage to make it work (`import Color from "colorjs.io/src/color.js";` was not the solution).

---
Adding a triple-slash directive `/// <reference lib="dom" />` at the top of `node_modules/colorjs.io/types/index.d.ts` solves the problem for me (by telling TS to use these types for this library).

So, I assume this is as simple as adding the directive to [`types/index.d.ts`](https://github.com/color-js/color.js/blob/main/types/index.d.ts), however, given the latest commit on that file, it looks like it is somewhat autogenerated, so there may be more machinery I'm unaware of that needs to handle that.

Jym77:
Closes #708.

### DIFF
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -1,3 +1,4 @@
+/// <reference lib="dom" />
 /**
  * @packageDocumentation
  * Defines and re-exports many types for use throughout the library.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"target":"ES2018",
 		"checkJs": true,
-		"lib": ["DOM", "ESNext"],
+		"lib": ["ESNext"],
 		"skipLibCheck": true,
 		"declaration": true,
 		"emitDeclarationOnly": true,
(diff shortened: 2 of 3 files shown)


lloydk (approved):
LGTM

MysteryBlokHed (approved):

Issue #713, opened 2026-01-29
Add remaining CSS gamut mapping algorithms


jamesnw:
The EdgeSeeker and Ray Trace gamut mapping algorithms are now part of the [CSS spec](https://drafts.csswg.org/css-color-4/#css-gamut-mapping). They are currently available in the [Apps repo](https://github.com/color-js/apps/tree/main/gamut-mapping), but are not available as methods in color.js.

facelessuser:
This is interesting. In order to do this in Color.js, I think a couple of decisions would need to be made.

1. Do we hard code them to only work with `srgb-linear`, `display-p3-linear`, and `rec2020-linear` like CSS or do we add enough support that you could make it generic in most cases? This can be relative to the algorithm, but should be decided.

2. Do you allow the perceptual space to be swapped out to allow others to be used or just stick with OkLCh?

I'm not going to push hard in any direction, but more want to get a feel for what is wanted in this project. I'll just mention what is done in ColorAide, but again, nothing I outline below needs to be done, it is more just describing what is done in ColorAide.

In ColorAide, we allow any perceptual space to be used. Results depend on the geometry of the space, but generally, the approach translates to other spaces as well.

Each gamut bound color space that is being mapped into will usually specify an appropriate RGB equivalent gamut (preferably linear, but not absolutely required), and if they are RGB, they can specify a linear variant. This is to make it so the user doesn't have to think too hard about what they want, but we could require a user must specify an RGB space. Some spaces, that do not have a direct RGB space to reference are converted into a cube shape (HPLuv as an example) and used as the gamut, or we could simply not support such gamuts.

This could be taken as far or not as far as is liked.

EdgeSeeker is another is a whole other beast. We would need to determine how many gamuts it can support, when we generate the loots, do we store or does the user store them and pass them around?

facelessuser:
I have a draft here that adds the Ray Trace GMA: #715.

Right now, it is super "dumb". It will take a space or infer the space from the current origin. If that space is not an RGB gamut, it will throw an error. There's no logic to figure out what the linear version of the RGB space is, so if you want the linear version, you have to specify the specific linear version.

I'm keeping it pretty strict to the spec, so if you give it an HDR RGB space, it will return white or black if it is beyond the SDR range, and it will clamp the bounding box for ray tracing to an SDR range. I'm not sure exactly how we to handle HDR stuff in this context within Color.js, so I'm open to doing things differently.

[...]

PR #715, opened 2026-01-31
Add the CSS Ray Trace GMA

facelessuser:
- Add the Ray Trace GMA
- Add `rgbGamut` for bounded spaces to specify a compatible RGB gamut.
- Add `linearGamut` for non-linear spaces to declare a linear version (if one is available).
- Update spaces to declare appropriate gamuts so that Ray Trace can properly act upon them.

Related #713

### DIFF
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -354,7 +354,7 @@ function raytrace_box (start, end, bmin = [0, 0, 0], bmax = [1, 1, 1]) {
 		direction.push(d);
 
 		// Non parallel cases
-		if (Math.abs(d) > 1e-14) {
+		if (Math.abs(d) > 1e-12) {
 			const inv_d = 1 / d;
 			const t1 = (bn - a) * inv_d;
 			const t2 = (bx - a) * inv_d;


facelessuser:
Needs testing before a proper review.

facelessuser:
The more I think about it, the more I'm of the mind that raytrace should match the CSS MINDE and clamp to SDR lightness, then check if they are in gamut and abandon if further processing if it is in gamut. The two algorithms should be interchangeable, provide similar results, and have the same intent.

On paper, this shouldn't change anything in Color.js, as there is a hard check for anything in-gamut before the chroma reduction is executed, but there are a few cases where a difference is noticeable.

The exceptions are in spaces like the HDR ones that don't handle negative values very well. Those have many colors that will show up as out-of-gamut, but once you round-trip them through OkLCh, they come out as in gamut. This yields a slight difference with MINDE because MINDE abandons chroma compression, while Ray Trace will find a spot on the gamut surface.

```js
> new Color('rec2100hlg', [0.85655, -0.63822, -0.28243]).to('oklch').to('rec2100hlg').coords
[ 0.8565500000000001, 0.6128665103944593, 0.2824300000000001 ]
> new Color('rec2100hlg', [0.85655, -0.63822, -0.28243]).toGamut({method: 'css'}).coords
[ 0.8565500000000001, 0.6128665103944593, 0.2824300000000001 ]
> new Color('rec2100hlg', [0.85655, -0.63822, -0.28243]).toGamut({method: 'raytrace'}).coords
[ 0.876192567898056, 0.5699099283758461, 0 ]
```

Regardless of which is perceived as better or worse, I think they should at least behave similarly, with the same intent. For this reason, I'll match the CSS MINDE behavior as a final change. I guess the spec should be tweaked to represent this as well.

src/toGamut.js
```diff
@@ -99,8 +100,11 @@ export default function toGamut (
 	if (method === "css") {
 		spaceColor = toGamutCSS(color, { space });
 	}
+	else if (method === "raytrace") {
+		spaceColor = toGamutRayTrace(color, { space });
+	}
 	else {
-		if (method !== "clip" && !inGamut(color, space)) {
```
facelessuser: This was doing nothing as we have a gamut check with 0 epsilon at the start. This was just a cleanup, but can revert if desired. I realize this is out of scope.

PR #716, opened 2026-01-31
Fix HCT tests and fix best XYZ not being used if target accuracy cannot be achieved

facelessuser:
- Test values were wrong for two tests.
- Best XYZ was not being used when approximation accuracy could not be achieved, only the last. This did not affect the tests.
- Save best XYZ instead of best J used to get the best XYZ so we can save a calculation at the end. This also did not affect the tests.

### DIFF
--- a/src/spaces/hct.js
+++ b/src/spaces/hct.js
@@ -90,7 +90,7 @@ function fromHct (coords, env) {
 
 	// We could not acquire the precision we desired,
 	// return our closest attempt.
-	return xyz;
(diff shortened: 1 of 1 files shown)


facelessuser:
These tests were verified against the ColorAide implementation. Not sure why the expected values were wrong here.

svgeesus:
In addition to the changes to the tests, there are also changes to the HCT implementation - could you explain why those changes were made?

PR #719, opened 2026-02-06
fix(docs): typo in procedural API

drwpow:
- Changed `space` to `spaceId`
- (Optional, feel free to reject) Added comments showing output, just to make the example a little clearer to follow


LeaVerou:
IIRC either `space` or `spaceId` should work, and it's simpler to just use `space` everywhere. Did it not work that way? If so, that's probably a bug.

PR #734, opened 2026-05-14
Avoid implied color space dependencies for procedural usage

aduth:
Updates utility functions to use color space objects directly rather than retrieving the color space by a string lookup, which infers a dependency that those color spaces are already registered.

This improves usability of [the procedural API](https://colorjs.io/docs/procedural). Currently, if someone opts-in to the procedural API and wants to use these utilities, the consumer has to ensure that the color space is registered, which is not obvious. 

Using the color space object directly ensures these functions are independently self-contained and don't have any implied dependencies.

I noticed similar implied dependencies in [`lighten`](https://github.com/color-js/color.js/blob/acf4475326cde617d53dd0b5f38abfa16d42a04c/src/variations.js#L12) and [`darken`](https://github.com/color-js/color.js/blob/acf4475326cde617d53dd0b5f38abfa16d42a04c/src/variations.js#L23) functions, but given the "alternatives" fallbacks used there, it's less clear to me how best to handle those color spaces. I'm happy to update those functions as well if there's a direction that makes sense.

### DIFF
--- a/src/toGamut.js
+++ b/src/toGamut.js
@@ -7,6 +7,7 @@ import inGamut from "./inGamut.js";
 import to from "./to.js";
 import get from "./get.js";
 import oklab from "./spaces/oklab.js";
+import oklch from "./spaces/oklch.js";
 import set from "./set.js";
 import clone from "./clone.js";
 import getColor from "./getColor.js";
@@ -251,13 +252,12 @@ export function toGamutCSS (origin, { space } = {}) {
 	}
 
 	space = ColorSpace.get(space);
-	const oklchSpace = ColorSpace.get("oklch");
 
 	if (space.isUnbounded) {
 		return to(origin, space);
 	}
 
-	const origin_OKLCH = to(origin, oklchSpace);
+	const origin_OKLCH = to(origin, oklch);
 	let L = origin_OKLCH.coords[0];
 
 	// return media white or black, if lightness is out of range
@@ -415,8 +415,7 @@ export function toGamutRayTrace (origin, { space } = {}) {
 	}
 
 	// Get the OkLCh coordinates.
-	const oklchSpace = ColorSpace.get("oklch");
-	let oklchOrigin = to(origin, oklchSpace);
+	let oklchOrigin = to(origin, oklch);
 	let [lightness, chroma, hue] = oklchOrigin.coords;
 
 	// Return white or black if color's lightness exceeds the SDR range.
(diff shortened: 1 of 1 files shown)


src/toGamut.js
LeaVerou, line 36: Totally unrelated to this PR, but we shouldn't be having colorspace-specific stuff in a generic function like this. Especially for non-essential color spaces. This should live together with the HCT color space!

LeaVerou:
I need to look more closely but it seems that the oklch space is only used in the `toGamutCSS()` path? If so, this makes tree shaking _worse_, since it essentially turns a peer dependency that is rarely needed into a regular dependency.

It's probably not a huge deal since most people _do_ need oklch, and it's pretty small but thought I'd flag it and see what others think.

That said, I'm not even sure we _should_ have `toGamutCSS()` in core. CSS itself is moving away from this! We should probably move the whole thing to a plugin…

aduth:
If my read of the code is accurate, `toGamutCSS` would be the default path for all `toGamut` usage unless `method` is overridden to something other than `'css`' ([`gamut_mapping` defaults to `'css'`](https://github.com/color-js/color.js/blob/acf4475326cde617d53dd0b5f38abfa16d42a04c/src/defaults.js#L3) which [`toGamut` directs to `toGamutCSS`](https://github.com/color-js/color.js/blob/acf4475326cde617d53dd0b5f38abfa16d42a04c/src/toGamut.js#L100-L101))?

lloydk:
> I need to look more closely but it seems that the oklch space is only used in the `toGamutCSS()` path? If so, this makes tree shaking _worse_, since it essentially turns a peer dependency that is rarely needed into a regular dependency.
> 
> It's probably not a huge deal since most people _do_ need oklch, and it's pretty small but thought I'd flag it and see what others think.
> 

`toGamutRayTrace()` also uses oklch.


> That said, I'm not even sure we _should_ have `toGamutCSS()` in core. CSS itself is moving away from this! We should probably move the whole thing to a plugin…

I would like to see some kind of plugin system so that all of the gamut mapping methods aren't in a single file and additional gamut mapping methods could be added e.g. https://bottosson.github.io/posts/gamutclipping/

lloydk:
I haven't looked that closely but couldn't this PR be as simple as: 

- `import oklchSpace from "./spaces/oklch.js";`
- delete the calls to `ColorSpace.get("oklch")`

facelessuser:
I generally agree that as the gamut mapping options grow, making gamut mapping methods into plugins probably makes more sense, though that is tangential to this PR.

LeaVerou (approved):
Thanks for contributing! Oof, `toGamut()` needs a general overhaul because there are so many different methods (see https://apps.colorjs.io/gamut-mapping/ ) each with their own dependencies.
But this is an improvement over the status quo meanwhile.

PR #736, opened 2026-06-12
Add gamut-relative OKLCh & LCH color spaces (GamutRelativeColorSpace)

LeaVerou:
## Summary

Adds **gamut-relative** polar color spaces, where chroma is normalized to an RGB gamut so that `c = 1` is the most colorful in-gamut color for a given lightness and hue, and `c = 0.5` is half that. Lightness and hue behave exactly like the base space; only chroma is rescaled. This makes `c` an intuitive "how colorful, for this gamut" axis (great for pickers, gamut-aware gradients, generating maximally-saturated colors).

```js
new Color("oklch-p3", [0.7, 1, 30]);    // most saturated P3 color at L=0.7, H=30
new Color("red").to("oklch-p3").coords;  // [0.628, 0.896, 29.2] — sRGB red is ~0.9 of P3
new Color("oklch-p3", [0.7, 0.8, 200]).toString(); // color(--oklch-p3 70% 80% 200)
```

### New color spaces (6)

`oklch-p3`, `oklch-srgb`, `oklch-rec2020`, `lch-p3`, `lch-srgb`, `lch-rec2020`.

### New class: `GamutRelativeColorSpace`

A general, base-agnostic class (exported alongside `RGBColorSpace`). Given a polar `base` and a `gamutSpace`, it inherits the base's coordinates, rescales the chroma coordinate to a 0–1 fraction of the gamut's max chroma, and derives the serialization format. The chroma coordinate is configurable via `chroma` (default `"c"`), so it works for OKLCh, LCH, and any polar space with a chroma axis. Each of the 6 spaces above is a four-line instance.

The LCH spaces are included partly to prove the class bakes in no OKLCh-specific assumptions: LCH keeps its `[0,100]` lightness (serialized as a number), while OKLCh's `[0,1]` lightness serializes as a percentage.

## Semantics: `c = 1` is the highest *contiguous* in-gamut chroma

`maxChroma(l, h)` returns the most chroma reachable in a straight line **from neutral** — i.e. the boundary of the gamut region connected to the achromatic axis. This is well-defined and deterministic: the chroma search is seeded from the gamut's most-saturated corner (so it's out of gamut at every `l`/`h`, no per-gamut constant), and converges on the first boundary crossing out from `c = 0`.

Because OKLab is **non-convex near blue** (CIELAB too, elsewhere), some genuinely in-gamut colors lie beyond a gap in chroma and therefore read as `c > 1`. This is rare and by design — e.g. the blue primary's chroma exceeds the contiguous boundary at its lightness/hue, so `color(srgb 0 0 1) → oklch-srgb` gives `c ≈ 1.18`. Measured prevalence of in-gamut `c > 1`: ~0.01% (sRGB), ~0.14% (Rec.2020), 0% (P3, effectively convex). The alternative (a continuous "filled" boundary) was rejected because the gamut boundary genuinely steps there — no boundary definition is continuous across the fold.

## Implementation notes

- **`maxChroma`** is a method on `GamutRelativeColorSpace`. It seeds the reduced coordinate with `oogChroma` (a chroma that is out of gamut at every lightness/hue, precomputed once as the max of the gamut corners' base-space chroma + 1% headroom) and reduces it with `toGamut` (`method` e.g. `"oklch.c"`, `deltaEMethod: "OK"`, `jnd: 0`) to find the contiguous in-gamut boundary.
- **Caching removed for the MVP.** An earlier iteration lived in its own `src/maxChroma.js`, memoized results per (space, gamut) in a two-level digit trie (lightness → hue), and added `min`/`max` search-bound options to `toGamut()`. All of that was dropped to keep the MVP focused: `maxChroma` is now a plain method that computes on demand, and `toGamut` is unmodified. Memoization can be reintroduced later behind the same method if profiling warrants it.
- **Serialization**: each space declares a `cssId` (e.g. `--oklch-p3`); chroma and any 0–1 lightness serialize as percentages.
- `ColorSpace` keeping a subclass's own `toBase`/`fromBase` when not overridden via options landed separately in #746 (now on `main`).

## Tests & accuracy

Conversion tests for all 6 spaces (primaries → `c = 1`, interior → `c < 1`, achromatic → `c = 0`/`h = none`, blue-primary `c > 1` documented). For P3, accuracy vs. a geometric reference is within the 1e-4 design tolerance; round-trips are exact to ΔE ~1e-13.

> [!NOTE]
> `docs/images/spaces.d2` is regenerated from `spaces.njk`; `docs/images/spaces.svg` still needs regenerating with the `d2` tool (not available in CI here).

https://claude.ai/code/session_01Pg44HttqmX4k1HxsUqZcYy


LeaVerou:
## Why this stays P3-only: gamut-relative OKLCh doesn't generalize to sRGB / Rec.2020

I investigated extending this to `oklch-srgb` and `oklch-rec2020`. They build fine (identical to `oklch-p3` bar the gamut), but two findings make them not worth shipping. Sharing the data for the record.

### 1. The chroma-boundary slopes are *not* similar across gamuts

Ran `scripts/oklch-gamut-bounds.js <gamut>` (now parameterized over the gamut) for all three:

| Gamut | max \|dC/dh\| per° | max \|dC/dl\| (l ≤ 0.87) | max \|dC/dl\| (l > 0.87) |
|---|---|---|---|
| **P3** | 0.0531 | 3.504 | 19.288 |
| **sRGB** | **9.4455** | 2.980 | 15.674 |
| **Rec.2020** | **8.8714** | 3.972 | **36.301** |

The hue slope for sRGB/Rec.2020 is **~170× P3's**. Those steep slopes are confined to the cusps, though, so the trie's `STEPS = 1e6` is still adequate — sweeping `STEPS` from 1e6 → 1e9 left the residual error flat. Resolution is not the bottleneck.

### 2. The real blocker: sRGB and Rec.2020 are non-convex in OKLab

The "max in-gamut chroma" model assumes the in-gamut chroma set along a constant-(l, h) ray is a single interval `[0, cMax]`. That holds for P3 (effectively convex in OKLab) but **not** for sRGB/Rec.2020 near blue, where the gamut folds into disconnected **islands**. Example, sRGB at l = 0.42, h = 264:

```
in-gamut c-intervals = [0.000, 0.248]  ∪  [0.288, 0.290]
```

Colors in the outer island are genuinely in-gamut but the boundary search finds the main body's edge (0.248), so they normalize to **c ≈ 1.17**. This is topological, not a precision issue — for a non-convex gamut you cannot have both "c ≤ 1 ⟹ in gamut" and "c = 1 ⟹ on the boundary". Measured prevalence (with a seed fix, see below):

| Gamut | colors with c > 1 | worst c |
|---|---|---|
| **P3** | 0% | 1.0000 |
| **sRGB** | ~0% (extreme outliers only) | 1.16 |
| **Rec.2020** | ~0.14% | 1.16 |

P3 is clean precisely because it's convex enough; the others aren't.

### Bonus finding: a latent seed clamp (would matter if we ever add wider gamuts)

`maxChroma` seeds its `toGamut` search from OKLCh's `c` refRange max (**0.4**). Rec.2020 reds/greens reach ~0.47 OKLCh chroma, so they'd clamp at 0.4. P3 is unaffected (its max OKLCh chroma < 0.4), so no change is needed here now — noting it in case a wider gamut-relative space is ever considered.

### Conclusion

Keeping the PR scoped to **`oklch-p3`** only. Gamut-relative OKLCh is well-behaved for ~convex P3; sRGB/Rec.2020 would ship with rare-but-real `c > 1` colors that can't be fixed by tuning. The script generalization (`scripts/oklch-gamut-bounds.js <gamut>`) is included so these numbers are reproducible.

---
_Generated by [Claude Code](https://claude.ai/code)_

LeaVerou:
## Update: shipping `oklch-srgb` and `oklch-rec2020` after all, with a defined semantics for the fold

Superseding my earlier "P3-only" comment. After establishing that the blue-cusp issue is a genuine boundary discontinuity (no choice of boundary definition is continuous — both first-exit and outermost step ~0.05 over ~0.5°), we settled on a definition rather than smoothing:

**`c = 1` is the highest *contiguous* in-gamut chroma** — the most chroma reachable in a straight line from neutral at that lightness and hue. Disconnected gamut islands (the non-convex blue fold) sit beyond it, so colors there read **`c > 1` by design**. This keeps `c = 1` meaning the same thing everywhere, computed deterministically.

### Changes
- **`oklch-srgb`, `oklch-rec2020`** added next to `oklch-p3` (registered, accessors, conversion tests, spaces.d2).
- **`maxChroma` seeds the chroma search from the gamut's most-saturated corner** (its primaries are its chroma maxima) plus headroom, instead of OKLCh's nominal 0.4. Since the seed is out of gamut at every `l`/`h`, `toGamut`'s binary search converges on the **first** crossing out from `c = 0` → deterministic, contiguous boundary. This also removes the Rec.2020 clamp (its chroma exceeds 0.4) **without hardcoding a per-gamut constant**. `STEPS` stays `1e6`; P3 is unchanged (its existing tests still pass).

### Behavior
- In-gamut colors with `c > 1` are rare and **confined to blue**: ~0.01% (sRGB), ~0.14% (Rec.2020), 0% (P3, effectively convex).
- The **blue primary itself** is in the island: `color(srgb 0 0 1)` → `c ≈ 1.18`, `color(rec2020 0 0 1)` → `c ≈ 1.16`. This is the famous OKLab blue non-convexity, encoded explicitly in the tests.
- Primaries at the contiguous boundary → `c = 1`; in-gamut interior → `c < 1`; achromatic → `c = 0`, `h = none`.

Tests: 719/791 pass (0 failures; +11 for the new spaces). `docs/images/spaces.svg` needs regenerating from `spaces.d2` (no `d2` in this environment).

---
_Generated by [Claude Code](https://claude.ai/code)_

src/GamutRelativeColorSpace.js
```diff
@@ -0,0 +1,85 @@
...
+
+		// Inherit the base's coords; chroma becomes a 0–1 fraction of the gamut's max chroma
+		let coords = Object.fromEntries(
+			Object.entries(base.coords).map(([id, meta]) => [id, { ...meta }]),
+		);
+		coords[chroma].range = [0, 1];
+		delete coords[chroma].refRange;
+		options.coords ??= coords;
+
+		// Serialize chroma (always 0–1) and any 0–1 lightness as percentages; hue as number/angle
```
LeaVerou:
```suggestion
		// Serialize chroma and any 0–1 lightness as percentages; hue as number/angle
```

LeaVerou:
Reviewed by @svgeesus on my laptop

lloydk:
A few comments:

Performance could be improved if the `gamut.id/chromaCoord` function was cached in the `GamutRelativeColorSpace`. Right now the `{ space, gamut, chromaCoord }` and cache key objects are being created on every conversion.

The cache is unbounded and in certain cases could take up a lot of memory. I simulated generating a 300px square lightness vs chroma color area with a hue slider that had a step of 1 degree. The cache size was around 85MB under Node and 40MB under Bun for all lightness/hue combinations (108,000). It might be useful to be able to clear the cache.

There are no tests for the `isNone` behavior.

lloydk:
You may want to consider simplifying the cache to use nested Maps. Here's a copy of maxChroma.js that uses nested maps. Coded by Claude, completely unverified. In my color picker simulation memory dropped from 85MB to 6.5MB under Node and 40MB to 20MB under Bun. Warmed cache conversion was 2.3 times faster on my machine and a nice bonus is that the `digitsOf` function which was allocating an array every conversion is now gone.
<details>
<summary>maxChroma.js</summary>

```javascript
import ColorSpace from "./ColorSpace.js";
import toGamut from "./toGamut.js";
import { constrain } from "./angles.js";

/** @import { ColorSpace as ColorSpaceType } from "./types.js" */

// Cells per coordinate. Calibrated so a cell spans < 1e-4 of chroma for OKLCh + P3
// (see scripts/oklch-gamut-bounds.js); finer would not change the result.
const STEPS = 1e6;

// Each built function closes over a two-level cache keyed by lightness then hue, both quantized to
// integer cells (see makeMaxChroma).
/** @typedef {((l: number, h: number) => number) & { clear: () => void }} MaxChromaFn */

/**
 * Maximum chroma within an RGB gamut for a given lightness and hue, in the polar space's chroma
 * units (the highest *contiguous* in-gamut chroma — see {@link makeMaxChroma}). Builds a fresh
 * lookup each call; for repeated queries keep a {@link makeMaxChroma} function around instead (which
 * is what the gamut-relative color spaces do).
 * @param {number} l Lightness, in the polar space's units
 * @param {number} h Hue angle, in degrees
 * @param {{ space: string | ColorSpaceType, gamut: string | ColorSpaceType, chromaCoord?: string }} options
 *        `space` is the polar space whose chroma is reduced; `gamut` is the RGB gamut to fit;
 *        `chromaCoord` is the id of the chroma coordinate (default `"c"`, as in OKLCh and LCH).
 * @returns {number}
 */
export default function maxChroma (l, h, { space, gamut, chromaCoord = "c" }) {
	return makeMaxChroma(space, gamut, chromaCoord)(l, h);
}

/**
 * Build a memoized maxChroma function for one (space, gamut, chromaCoord): the highest contiguous
 * in-gamut chroma at a given lightness and hue, in the polar space's chroma units. Results are
 * cached in a Map held by the returned function; call its `.clear()` to free that memory (it refills
 * on demand). Hold on to the function and reuse it across conversions.
 * @param {string | ColorSpaceType} space Polar space whose chroma is reduced
 * @param {string | ColorSpaceType} gamut RGB gamut to fit
 * @param {string} [chromaCoord] Id of the chroma coordinate (default `"c"`, as in OKLCh and LCH)
 * @returns {MaxChromaFn}
 */
export function makeMaxChroma (space, gamut, chromaCoord = "c") {
	space = ColorSpace.get(space);
	gamut = ColorSpace.get(gamut);

	if (!space.isPolar || !(chromaCoord in space.coords)) {
		throw new TypeError(
			`makeMaxChroma() needs a polar space with a "${chromaCoord}" coordinate, got ${space.id}`,
		);
	}

	// Identify coordinates by role
	let ids = Object.keys(space.coords);
	let hueId = ids.find(id => space.coords[id].type === "angle");
	let lightnessId = ids.find(id => id !== chromaCoord && id !== hueId);
	let range = id => space.coords[id].range ?? space.coords[id].refRange;
	let [lMin, lMax] = range(lightnessId);
	let [hMin, hMax] = range(hueId);
	let cIndex = ids.indexOf(chromaCoord);
	let method = `${space.id}.${chromaCoord}`;

	// Memoized results, keyed by lightness cell then hue cell. Each coordinate is quantized to an
	// integer in [0, STEPS] so near-identical floats collapse to the same entry. Both cells stay within
	// V8's small-integer range, so the nested Maps key on unboxed integers — lighter and faster than a
	// flat Map keyed on a single (out-of-range) combined integer.
	/** @type {Map<number, Map<number, number>>} */
	let cache = new Map();

	// Seed from the gamut's most saturated corner (+ headroom) so the seed is out of gamut at every
	// l/h: the binary search then converges on the first crossing from c=0, i.e. the highest
	// *contiguous* in-gamut chroma (non-convex islands beyond it read as c>1). Deriving the seed from
	// the gamut also avoids clamping wide gamuts (Rec.2020 exceeds OKLCh's nominal 0.4 chroma).
	// prettier-ignore
	let corners = /** @type {[number, number, number][]} */ (
		[[0, 0, 1], [0, 1, 0], [0, 1, 1], [1, 0, 0], [1, 0, 1], [1, 1, 0]]
	);
	let seedC = 1.1 * Math.max(...corners.map(rgb => space.from(gamut, rgb)[cIndex] || 0));

	let lookup = (l, h) => {
		if (l <= lMin || l >= lMax) {
			return 0;
		}

		let lCell = Math.floor(((l - lMin) / (lMax - lMin)) * STEPS);
		let hCell = Math.floor(((constrain(h) - hMin) / (hMax - hMin)) * STEPS);

		let byHue = cache.get(lCell);
		if (!byHue) {
			cache.set(lCell, (byHue = new Map()));
		}

		let value = byHue.get(hCell);
		if (value === undefined) {
			let coords = /** @type {[number, number, number]} */ (
				ids.map(id => (id === chromaCoord ? seedC : id === hueId ? h : l))
			);

			value = toGamut(
				{ space, coords },
				// deltaE OK stays sensitive near black, where deltaE 2000 is too forgiving of excess chroma
				{ space: gamut, method, deltaEMethod: "OK", jnd: 0 },
			).coords[cIndex];
			byHue.set(hCell, value);
		}

		return value;
	};

	return Object.assign(lookup, {
		clear () {
			cache.clear();
		},
	});
}
```

</details>

LeaVerou:
Huh, that's interesting. I'd have expected an array to be way cheaper than a Map!

lloydk:
> Huh, that's interesting. I'd have expected an array to be way cheaper than a Map!

I think the issue is that a lot of small arrays were being allocated per cache entry and there was probably very few "nice" numbers in my example so most of the leaf nodes were at a depth of 6.

src/GamutRelativeColorSpace.js
```diff
@@ -0,0 +1,119 @@
...
+					let [min, max] = base.coords[id].range ?? base.coords[id].refRange ?? [];
+					let pctFirst = id === chroma || (min === 0 && max === 1);
+					return pctFirst ? "<percentage> | <number>" : "<number> | <percentage>";
+				}),
+			},
+		};
+
+		super(options);
+
+		this.chromaIndex = ids.indexOf(chroma);
```
LeaVerou: This should be right after `ids` in L24.

src/GamutRelativeColorSpace.js
```diff
@@ -0,0 +1,120 @@
...
+		// A chroma that is out of gamut at every lightness and hue: the gamut's most saturated corner
+		// (its primaries/secondaries are its chroma maxima) plus headroom. Seeding the gamut search
+		// from here lands on the first in-gamut crossing — the highest contiguous in-gamut chroma —
+		// and avoids clamping wide gamuts (e.g. Rec.2020 exceeds OKLCh's nominal 0.4 chroma).
+		// prettier-ignore
+		let corners = /** @type {[number, number, number][]} */ (
+			[[0, 0, 1], [0, 1, 0], [0, 1, 1], [1, 0, 0], [1, 0, 1], [1, 1, 0]]
+		);
+		this.oogChroma =
+			1.1 *
```
LeaVerou:
```suggestion
			1.01 *
```

LeaVerou:
I decided to drop the cache entirely for this MVP, since the exact tradeoff of space and time is unclear. We can add it in a later iteration. What do you think?

svgeesus (approved):
LGTM

PR #740, opened 2026-06-17
[interpolation] Fix `steps` using `deltaEMethod` inconsistently

### LINKED ISSUE #632: `steps()` uses `deltaEMethod` inconsistently
The `steps()` function makes four calls to `deltaE()`, but only one of them passes the `deltaEMethod` option.
https://github.com/color-js/color.js/blob/ffa9c3f61323040a831cb28bec2270471ab6deac/src/interpolation.js#L86
https://github.com/color-js/color.js/blob/ffa9c3f61323040a831cb28bec2270471ab6deac/src/interpolation.js#L112
https://github.com/color-js/color.js/blob/ffa9c3f61323040a831cb28bec2270471ab6deac/src/interpolation.js#L127

I noticed this after calling
```js
steps(
	{ space: sRGB, coords: from },
	{ space: sRGB, coords: to },
	{ space, steps, outputSpace: sRGB, hue, deltaEMethod: 'OK' },
)
```
without having the `Lab` color space registered first.

Is this behavior intended?

TimonLukas:
Fixes #632 

I stumbled over the same issue. Due to my limited experience with color stuff I'm not sure this fix is correct, especially since [this comment](https://github.com/color-js/color.js/issues/632#issuecomment-3689743575) mentions that some parts might already have been fixed, which I can't find - but this change fixes the error, the resulting colors look roughly as I'd expect them to, and the same amount of tests pass with and without.

Hope this is useful!

### DIFF
--- a/src/interpolation.js
+++ b/src/interpolation.js
@@ -78,7 +78,7 @@ export function steps (c1, c2, options = {}) {
 		colorRange = range(c1, c2, rangeOptions);
 	}
 
-	let totalDelta = deltaE(c1, c2);
+	let totalDelta = deltaE(c1, c2, deltaEMethod);
 	let actualSteps =
 		maxDeltaE > 0 ? Math.max(steps, Math.ceil(totalDelta / maxDeltaE) + 1) : steps;
 	let ret = [];
@@ -120,7 +120,11 @@ export function steps (c1, c2, options = {}) {
 
 				let p = (cur.p + prev.p) / 2;
 				let color = colorRange(p);
-				maxDelta = Math.max(maxDelta, deltaE(color, prev.color), deltaE(color, cur.color));
+				maxDelta = Math.max(
+					maxDelta,
+					deltaE(color, prev.color, deltaEMethod),
+					deltaE(color, cur.color, deltaEMethod),
+				);
 				ret.splice(i, 0, { p, color: colorRange(p) });
 				i++;
 			}


facelessuser:
I feel like this should have a test representing what previously failed but is now working due to the fix.

TimonLukas:
Makes sense! I've been trying to write one, but I need some advice regarding the test runner.

Since the `steps()` issue happens when using the procedural API while the `Lab` space is not registered, the test needs to modify the `ColorSpace.registry`. To keep the global state changes minimal I've been trying to swap out the existing registry, but this makes the tests somewhat brittle, as it causes any parallel tests to fail. The test suite passes if this test is running on levels 1 or 5 (`"Interpolation Tests"` and `"first"` below), and the suite fails while it's running on a level between them.

I couldn't find an existing solution in the other tests, or an intended approach in the `htest.dev` docs. Do you maybe have any tips?

<summary>
WIP test code
<details>

```javascript
// test/interpolation.js
import { ColorSpace, OKLCH, steps } from "../src/index-fn.js";

let originalRegistry;

export default {
	name: "Interpolation Tests",
	tests: [
		{
			name: "steps",
			tests: [
				{
					name: "One more level",
					tests: [
						{
							name: "No error",
							beforeAll () {
								originalRegistry = ColorSpace.registry;
								ColorSpace.registry = {};
							},
							afterAll () {
								ColorSpace.registry = originalRegistry;
							},
							beforeEach () {
								ColorSpace.registry = {};
								this.data.colorSpaces.forEach(space => {
									ColorSpace.register(space);
								});
							},
							run (first, second) {
								steps(first, second);
							},
							tests: [
								{
									name: "first",
									data: {
										colorSpaces: [OKLCH],
									},
									args: [
										{ space: "oklch", coords: [1, 0, 360], alpha: 1 },
										{ space: "oklch", coords: [0, 0, 0], alpha: 1 },
									],
								},
							],
						},
					],
				},
			],
		},
	],
};

```

</details>
</summary>

facelessuser:
I'd probably just use something like this:

```js
// test/interpolation.js
import Color from "../src/index.js";
import * as check from "../node_modules/htest.dev/src/check.js";

export default {
	name: "Interpolation Tests",
	check: check.deep(check.proximity({ epsilon: 0.00001 })),
	tests: [
		{
			name: "steps",
			tests: [
				{
					name: "One more level",
					tests: [
						{
							name: "No error",
							run (first, second, options) {
								const colors = new Color(first).steps(first, options).map(c => {return c.coords});
								return colors;
							},
							tests: [
								{
									name: "first",
									args: [
										{ space: "oklch", coords: [1, 0, 360], alpha: 1 },
										{ space: "oklch", coords: [1, 0, 0], alpha: 1 },
										{ space: "oklch", steps: 3, hue: "longer", deltaEMethod: "OK" }
									],
									expect: [[1, 0, 0], [1, 0, 180], [1, 0, 360]]
								},
							],
						},
					],
				},
			],
		},
	],
};
```

TimonLukas:
That unfortunately doesn't test what this PR fixes. As I said, the issue only happens while the `Lab` colorspace isn't registered, but it has already been registered when the test executes, so your suggestion doesn't hit the problematic path in the current `main` branch. The registry **has** to be cleared before the test (and presumably should be restored after).

Here's a minimal reproduction:

```javascript
// example.mjs
import { ColorSpace, OKLCH, steps } from "colorjs.io/fn"

ColorSpace.register(OKLCH)

steps({ space: "oklch", coords: [1, 0, 360], alpha: 1 }, { space: "oklch", coords: [0, 0, 0], alpha: 1 }, {
  deltaEMethod: "OK2"
})
```

If you save this as `example.mjs` and execute `pnpm init && pnpm add colorjs.io && node example.mjs`, the `steps()` function throws:
```
file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/ColorSpace.js:313
                                throw new TypeError(`No color space found with id = "${space}"`);
                                      ^

TypeError: No color space found with id = "lab"
    at ColorSpace.get (file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/ColorSpace.js:313:11)
    at distance (file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/distance.js:14:21)
    at Object.deltaE76 (file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/deltaE/deltaE76.js:11:9)
    at deltaE (file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/deltaE.js:28:27)
    at steps (file://node_modules/.pnpm/colorjs.io@0.6.1/node_modules/colorjs.io/src/interpolation.js:81:19)
    at file://example.mjs:5:1
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at async node:internal/modules/esm/loader:633:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.16.0
```

facelessuser:
Ah, I guess I didn't look super close at the issue it's fixing.

TimonLukas:
No problem! :) I still think a regression test would make sense if there's some clean way to do the override, but relying on nesting levels seems too brittle.

If someone can recommend a solution, I'll be happy to add the test case!

facelessuser:
Yeah, I'll take a closer look when I get a chance, assuming someone doesn't answer your question before I do.

svgeesus (approved):

PR #744, opened 2026-06-18
[ci] Fix install command

MysteryBlokHed:
No longer works as-is due to the removal of the lockfile from the repository; `npm ci` expects a lockfile

Opened as a PR instead of committing directly so that the workflow could be tested first

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
               with:
                   node-version: "lts/*"
 
-            - run: npm ci
+            - run: npm install
             - run: npm run lint:ci
             - name: Detect changes
               id: changes


MysteryBlokHed:
CI failure is related to a test rather than this change:

```
Error: /home/runner/work/color.js/color.js/types/test/display.ts:20:18
ERROR: 20:18  expect  TypeScript@5.3 compile error: 
Object literal may only specify known properties, and 'random' does not exist in type 'DisplayOptions'.

    at testTypesVersion (/home/runner/work/color.js/color.js/node_modules/@definitelytyped/dtslint/dist/index.js:194:15)
    at async runTests (/home/runner/work/color.js/color.js/node_modules/@definitelytyped/dtslint/dist/index.js:170:13)
    at async main (/home/runner/work/color.js/color.js/node_modules/@definitelytyped/dtslint/dist/index.js:79:9)
ERROR: "dtslint" exited with 1.
Error: Process completed with exit code 1.
```

PR #749, opened 2026-06-19
Expose color space transformation matrices via ColorSpace.M

### LINKED ISSUE #748: Color spaces should export any matrices they use
A lot of consumer code needs to duplicate matrix data because not all matrices are exported.

This should be a first-class concept.

`RGBColorSpace` already passes `toXYZ_M` and `fromXYZ_M` as options, but does not expose them.
Other color spaces, e.g. `OKLab` define a ton of matrices and never export them.
E.g. many of the algos in the GMA playground repeat matrices because of this.

I propose: `ColorSpace` has a `M` property that holds any matrices relevant to that color space. E.g. instead of `options.toXYZ_M`, the color space would refer to it via `this.M.toXYZ`.

This is a big lift, but hopefully a straightforward change. About to try it if no objections!

cc @svgeesus @lloydk @MysteryBlokHed @facelessuser

LeaVerou:
Closes #748

## Summary
This PR adds a new `M` property to color spaces that exposes the transformation matrices they use internally. This allows consumer code to reuse these matrices instead of duplicating the data, improving maintainability and reducing code duplication.

## Key Changes

- **ColorSpace API**: Added an `M` property to the `ColorSpace` class that holds the matrices relevant to that space as a `Record<string, Matrix3x3>`. It is initialized from the `M` option and defaults to an empty object for spaces that use no matrices.

- **RGBColorSpace**: The default `toBase()`/`fromBase()` are now regular prototype methods (shared across instances rather than recreated per space) that read the matrices from `this.M.toXYZ` / `this.M.fromXYZ`. Matrices can be supplied either via the existing `toXYZ_M` / `fromXYZ_M` options or via the generic `M` object.

- **Matrix exposure**: Color spaces that use transformation matrices now expose them via `M`:
  - `oklab`: `XYZtoLMS`, `LMStoXYZ`, `LMStoLab`, `LabtoLMS`
  - `cam16`: `cat16`, `cat16Inv`, `m1`
  - `ictcp`: `XYZtoLMS`, `LMStoXYZ`, `LMStoIPT`, `IPTtoLMS`
  - `jzazbz`: `XYZtoCone`, `ConetoXYZ`, `ConetoIab`, `IabtoCone`
  - `helmlab`, `helmgen`, `okhsl`, `okhsv`, and all RGB spaces (`M.toXYZ` / `M.fromXYZ`)

- **Removed redundant exports**: Now that matrices are accessible via `M`, the standalone matrix exports were removed (`oklab` `LMStoLab_M`/`LabtoLMS_M`, `srgb-linear` `fromXYZ_M`, `okhsl` `toLMS`/`toSRGBLinear`, `helmlab` `CAT_TO_HELM`/`CAT_FROM_HELM`). The `_M` suffix was dropped from the matrix consts so the `M` definitions can use object shorthand. Internal consumers (`okhsl`, `okhsv`, `hsluv`, `hpluv`) now read these matrices from the relevant space's `M`.

- **TypeScript definitions**: Added the `M` property to `SpaceOptions` and the `ColorSpace` class.

- **Tests**: Added `test/matrices.js` verifying that RGB and non-RGB spaces expose their matrices, that matrix-free spaces have an empty `M`, and that custom matrices passed via options are exposed.

https://claude.ai/code/session_01RP8psTHh2b5o9MQEAgqY2j


src/spaces/cam16.js
```diff
@@ -398,6 +398,12 @@ export default new ColorSpace({
 
 	base: xyz_d65,
 
+	M: {
+		cat16: cat16,
+		cat16Inv: cat16Inv,
+		m1: m1,
+	},
```
LeaVerou:
```suggestion
	M: { cat16, cat16Inv, m1 },
```

(and elsewhere)

src/spaces/ictcp.js
```diff
@@ -103,6 +103,12 @@ export default new ColorSpace({
 	},
 
 	base: XYZ_Abs_D65,
+	M: {
+		XYZtoLMS: XYZtoLMS_M,
+		LMStoXYZ: LMStoXYZ_M,
+		LMStoIPT: LMStoIPT_M,
+		IPTtoLMS: IPTtoLMS_M,
+	},
```
LeaVerou: Drop `_M` prefix so it can be compact?

src/RGBColorSpace.js
```diff
@@ -37,9 +37,16 @@ export default class RGBColorSpace extends ColorSpace {
...
-		if (options.toXYZ_M && options.fromXYZ_M) {
+		// Accept matrices either as dedicated options or via the generic `M` object
+		let toXYZ_M = options.toXYZ_M ?? options.M?.toXYZ;
+		let fromXYZ_M = options.fromXYZ_M ?? options.M?.fromXYZ;
+
+		if (toXYZ_M && fromXYZ_M) {
+			// Expose the matrices on the color space (via `this.M`) so consumers can reuse them
+			options.M = { ...options.M, toXYZ: toXYZ_M, fromXYZ: fromXYZ_M };
+
 			options.toBase ??= rgb => {
```
LeaVerou: Default `toBase` and `fromBase` can now be a regular method, no need to define it inline here (which creates a new instance every time)

src/RGBColorSpace.js
```diff
@@ -37,9 +37,16 @@ export default class RGBColorSpace extends ColorSpace {
...
 		}
 
-		if (options.toXYZ_M && options.fromXYZ_M) {
+		// Accept matrices either as dedicated options or via the generic `M` object
+		let toXYZ_M = options.toXYZ_M ?? options.M?.toXYZ;
+		let fromXYZ_M = options.fromXYZ_M ?? options.M?.fromXYZ;
+
+		if (toXYZ_M && fromXYZ_M) {
+			// Expose the matrices on the color space (via `this.M`) so consumers can reuse them
+			options.M = { ...options.M, toXYZ: toXYZ_M, fromXYZ: fromXYZ_M };
```
LeaVerou: Just use two assignments instead of cloning the whole object. Don't set/modify `options.M`, just set `this.M.fromXYZ`/`this.M.toXYZ` directly.

src/spaces/cam16.js
LeaVerou: Done — switched all the `M` definitions to object shorthand.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/spaces/ictcp.js
LeaVerou: Done — dropped the `_M` suffix here (and in `oklab`/`jzazbz`) so the shorthand is compact. Per your other note, since that breaks the named exports anyway I also removed the now-redundant standalone matrix exports (they're reachable via `M`); internal consumers (`okhsl`, `okhsv`, `hsluv`, `hpluv`) now read them from the relevant space's `M`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/RGBColorSpace.js
LeaVerou: Done — `toBase`/`fromBase` are now regular prototype methods on `RGBColorSpace` that read `this.M.toXYZ`/`this.M.fromXYZ`, so they're shared across instances instead of recreated per space.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/RGBColorSpace.js
LeaVerou: Done — no longer cloning; the matrices are assigned directly via `this.M.toXYZ` / `this.M.fromXYZ` after `super()`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/RGBColorSpace.js
```diff
@@ -37,26 +38,35 @@ export default class RGBColorSpace extends ColorSpace {
...
-				return xyz;
-			};
+		super(options);
 
-			options.fromBase ??= xyz => {
-				xyz = adapt(this.base.white, this.white, xyz);
-				return multiply_v3_m3x3(xyz, options.fromXYZ_M);
-			};
+		// Expose the matrices on the color space (via `this.M`) so consumers can reuse them.
+		// The default `toBase()` / `fromBase()` methods read them from here.
```
LeaVerou:
```suggestion
```

PR #752, opened 2026-06-22
Update gamut surface threshold and document values for 32 bit vs 64 bit

facelessuser:
For 64-bit values, 1e-12 can give better resolution in some specific cases. 1e-6 is better for 32-bit values.

As far as the CSS spec is concerned, the stated value of 1e-6 is fine, assuming implementation with 32-bit values. A note is likely needed to clarify this, not only for the gamut surface distance threshold, but for the ray trace delta threshold, which we also updated the comment for in this PR.


lloydk (approved):
LGTM

PR #753, opened 2026-07-11
Exclude hue angle from alpha premultiplication

svgeesus:
## Summary

Alpha premultiplication in `Color.range()` multiplied **every** coordinate by alpha. That's correct for rectangular spaces but wrong for polar ones: a hue angle must not be scaled by alpha (the old code even carried a `// not coping with polar spaces yet` note).

This finds the hue coordinate via the `isAngle` flag every space already declares (`angles.range`) and excludes it from both the premultiply and undo steps. Rectangular spaces are unaffected — `findIndex` returns `-1`, which matches no coordinate index.

## Testing

Exercised the real `Color.range(...)` path under Node:

- **HSL, opaque `hsl(30 100% 50%)` → transparent `hsl(90 100% 50%)`, premultiplied** → hue = **60°** (a plain midpoint). The old code would have premultiplied the angle and produced a wrong **30°**.
- **sRGB, opaque red → transparent blue, premultiplied** → `[1, 0, 0]` (transparent endpoint contributes nothing) — unchanged.
- **sRGB, plain interpolation** → `[0.5, 0, 0.5]` — unchanged.
- No `NaN` leaked from the skip/divide logic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/interpolation.js
+++ b/src/interpolation.js
@@ -147,10 +147,14 @@ Color.range = function(color1, color2, options = {}) {
 		[color1[space.id].hue, color2[space.id].hue] = angles.adjust(arc, [color1[space.id].hue, color2[space.id].hue]);
 	}
 
+	// Find the hue coordinate (if any) so it's excluded from premultiplication:
+	// an angle must not be scaled by alpha.
+	let coordNames = Object.keys(space.coords);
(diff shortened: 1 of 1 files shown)


svgeesus:
redoing without conflicts

PR #754, opened 2026-07-11
Premultiply alpha correctly in polar spaces, and handle none/zero alpha

svgeesus:
Premultiplication multiplied every coordinate by alpha, including the hue angle of polar spaces, which is meaningless. Per CSS Color 4 13.3, the hue angle is never premultiplied, and un-premultiplication must not divide by a zero or none alpha.

- Add ColorSpace#hueId and ColorSpace#hueIndex, so the hue coordinate is looked up from the space definition rather than assumed to be named h. isPolar is now expressed in terms of hueId.
- Skip the hue angle when premultiplying and when undoing premultiplication.
- Leave the coords untouched when undoing premultiplication with an alpha of zero, rather than dividing by zero and producing NaN.
- Carry a none alpha forward from the other color before premultiplying, per CSS Color 4 13.2, and leave alpha as none if both colors are missing it.
- Fix util.interpolate, which tested isNaN for missing components. none is null here, not NaN, so a none endpoint was coerced to 0 instead of taking the other endpoint's value. As a result, a none alpha used to premultiply every coordinate to zero, turning the color black.

Routing the hue lookup through hueId also fixes hue interpolation in Jzczhz, whose hue coordinate is named hz and so never matched the old check.

### DIFF
--- a/src/ColorSpace.js
+++ b/src/ColorSpace.js
@@ -159,14 +159,30 @@ export default class ColorSpace {
 		return this.formats?.color?.id || this.id;
 	}
 
-	get isPolar () {
+	/**
+	 * The id of this space's hue coordinate, or null if the space is not polar
+	 * @returns {string | null}
+	 */
+	get hueId () {
 		for (let id in this.coords) {
 			if (this.coords[id].type === "angle") {
-				return true;
+				return id;
 			}
 		}
 
-		return false;
+		return null;
+	}
+
+	/**
+	 * The index of this space's hue coordinate within its coords array, or -1 if the space is not polar
+	 * @returns {number}
+	 */
+	get hueIndex () {
+		return Object.keys(this.coords).findIndex(id => this.coords[id].type === "angle");
+	}
+
+	get isPolar () {
+		return this.hueId !== null;
 	}
 
 	/**
--- a/src/ColorSpace.js
+++ b/src/ColorSpace.js
@@ -495,7 +495,7 @@ export default class ColorSpace {
 function getPath (space) {
 	let ret = [space];
 
-	for (let s = space; (s = s.base); ) {
+	for (let s = space; (s = s.base);) {
 		ret.push(s);
 	}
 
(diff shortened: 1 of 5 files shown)


svgeesus:
Note that as a side effect, `interpolate`'s return type widened from `number` to `number | null`, which may need the type declarations regenerated.

facelessuser:
A test to verify this behavior would be good to ensure we don't ever break this in the future.

svgeesus:
> A test to verify this behavior would be good to ensure we don't ever break this in the future.

Agreed. Note that this PR already fixes some silent breakage that we hadn't noticed and wasn't tested:

- hue component in JzCzhz is called `hz` but some code was assuming all hues are called `h`
- old code was checking for `NaN` but should have been checking for `null`
- hues were being premultiplied (!) so 60 degrees with alpha 0.5 came out at 30 degrees 

There are a bunch of [examples in CSS Color 4](https://drafts.csswg.org/css-color-4/#interpolation), I was thinking of adapting them to our test framework.

svgeesus:
> A test to verify this behavior would be good to ensure we don't ever break this in the future.

Tests added, all the polar ones failed against the existing color.js and pass with this patch.

svgeesus:
@facelessuser @LeaVerou tests added, plus ESLint and Prettier run (just whitespace changes)

LeaVerou (approved):
LGTM. Note that it is baking an assumption in that any angular coordinate = hue, but that does seem to be true in practice.
I wonder if we should have an option to disable premultiplication? 🤔

svgeesus:
> LGTM. Note that it is baking an assumption in that any angular coordinate = hue, but that does seem to be true in practice.

Yes. I'm not aware of any real space with a non-hue angle. (Of course I could invent one, trivially).

> I wonder if we should have an option to disable premultiplication? 🤔

Since the goal is full compliance with CSS Color 4 and 5, which does not make premultiplication optional, I don't see the benefit.

Merge?

PR #761, opened 2026-07-27
Enable the disabled chromatic adaptation tests

### LINKED ISSUE #397: [chore] Port missing Chromatic adaptation tests
There are 13 tests in the old Chromatic adaptation test suite and 2 in the new test suite

soren-creates:
Closes https://github.com/color-js/color.js/issues/397.

### Background

The 11 commented-out sub-tests in test/adapt.js take their expected values from [Lindbloom's published tables](http://www.brucelindbloom.com/index.html?Eqn_ChromAdapt.html). As @svgeesus mentioned in [this thread](https://github.com/color-js/color.js/issues/397#issuecomment-1926080340), those are rounded to 7 decimals and computed from slightly different D50/D65 definitions, so they can't pass at a tolerance tight enough to be a useful regression test. Lindbloom's C→D50 Bradford matrix differs from what CATs.js computes by 7.1e-05, outside the suite's 5e-05 epsilon, which is why they were commented out rather than fixed.

### What this PR does

This recalculates those values at higher precision and uses those as the expected values, as @svgeesus suggested in that thread.

Every value was computed two independent ways, neither of which uses CATs.js's own arithmetic:

- The von Kries formula, written from scratch in Python/numpy, fed the whitepoints in src/adapt.js and the cone matrices in src/CATs.js.

- [colour-science](https://www.colour-science.org/)'s matrix_chromatic_adaptation_VonKries, given the same whitepoints.

The two agree to 2.7e-16 (worst case) across all 13 cases.

### The result

The suite epsilon goes from 5e-05 to 1e-13. The worst disagreement between these values and what CATs.js computes is 4.4e-16 — one ulp on the largest element — so 1e-13 sits ~225× above the noise floor while being ~5e+08× tighter than before. If we want to loosen that, 1e-12 would work as well.

Also fixes a name inherited from the old suite: the CAT16 C→D50 test was called "CAT02 D50 to D65", copy-pasted down from a preceding group. The two previously-enabled tests keep their values, which already agree with the recomputation to 2.2e-16.

npm test: 794/866 passing, 72 skipped — up from 783/855, i.e. the 11 newly enabled tests and nothing else.

(Drafted with the help of Claude.)

### DIFF
--- a/test/adapt.js
+++ b/test/adapt.js
@@ -1,3 +1,8 @@
+// The expected matrices here are computed at full double precision from the
+// whitepoints and cone matrices in src/, rather than taken from Lindbloom's
+// published tables. Lindbloom rounds to 7 decimals and uses slightly different
+// D50/D65 definitions, which is why these tests were previously commented out:
+// they cannot pass at a tolerance tight enough to catch a regression. See #397.
 import { WHITES } from "../src/adapt.js";
 import { adapt } from "../src/CATs.js";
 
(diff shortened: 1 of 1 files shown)


svgeesus (approved):
Excellent, thanks for doing this

PR #762, opened 2026-08-01
[spaces/xyy] Add CIE xyY color space

### LINKED ISSUE #724: Add CIE xyY
Even though [CIE xyY](https://en.wikipedia.org/wiki/CIE_1931_color_space#CIE_xyY_color_space) is a simple conversion from CIE XYZ, it would be handy to have it as it's own color space option.

tomatotomata:
## What changed

Closes #724.

- Added the CIE 1931 `xyY` color space with D65 XYZ as its base.
- Added the standard XYZ to xyY and xyY to XYZ conversions, including a black-color guard where chromaticity is undefined.
- Registered the space in the built-in space index and generated coordinate accessors.
- Added tests for D65 white, XYZ round-tripping, black, and `color(xyy ...)` parsing.

## Verification

- `npm test` passes: 787/859 passed, 72 skipped.
- `npx htest ./test/xyy.js` passes: 4/4.
- ESLint passes for the changed JavaScript files.
- `git diff --check` passes.

This contribution was prepared with AI assistance and reviewed and tested locally. The commit includes the required co-author trailer.

### DIFF
--- a/src/space-coord-accessors.d.ts
+++ b/src/space-coord-accessors.d.ts
@@ -52,10 +52,12 @@ declare class SpaceAccessors {
 	rec2100pq: SpaceAccessor;
 	srgb: SpaceAccessor;
 	srgb_linear: SpaceAccessor;
+	xyy: SpaceAccessor;
 	xyz: SpaceAccessor;
 	xyz_abs_d65: SpaceAccessor;
 	xyz_d50: SpaceAccessor;
 	xyz_d65: SpaceAccessor;
+	Y: number | null;
 	a: number | null;
 	az: number | null;
 	b: number | null;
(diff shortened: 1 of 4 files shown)


facelessuser:
Personally, I would return the white point's chromaticity points for x, y when Y is zero. I understand that some don't and some do, but I'm kind of in the camp that would.

tomatotomata:
I updated the black conversion to use the D65 white point chromaticity (`x = 0.3127`, `y = 0.329`) whenever the XYZ sum is zero, while preserving Y. I also changed the regression expectation and ran `npx htest ./test/xyy.js` with 4/4 passing. The commit is `6bca17e`. Do you think this matches the behavior you had in mind?

facelessuser:
It now matches more what I would personally expect. Not sure what's up with the lint action, but I don't think it's failing because of what you did.

PR #766, opened 2026-09-03
[types] Move from dtslint to ESLint

MysteryBlokHed:
dtslint after v0.0.182 is tailored to Definitely Typed and doesn't work well on arbitrary projects. This replaces it with an ESLint plugin to accomplish the same thing.

Claude was used to do the migration, but I looked the results over myself.


src/keywords.js
```diff
@@ -8,7 +8,7 @@
  *
  *  @type {Record<string, [number, number, number]>}
  */
-export default {
+const keywords = {
```
LeaVerou: We should not have to change perfectly valid code to appease a linting plugin. If it doesn't support this, it should be reported as a bug.

LeaVerou:
Thanks for working on this! LGTM except the comment I left. Linters forcing you change perfectly valid code is one of my pet peeves.

src/keywords.js
MysteryBlokHed: Fixed by changing one of the `$ExpectType` comments in the typing tests

PR #767, opened 2026-09-12
[parse] Reject unexpected punctuation in color function arguments

### LINKED ISSUE #614: new Color() accepts serialized coordinates with right paren as separator
Not sure if the title is the correct cause or way to phrase it.  I'm submitting this based on [this Q&A comment](https://github.com/color-js/color.js/discussions/611#discussioncomment-11477307) and @LeaVerou's response in the next comment.

This creates a valid `Color` instance using the right paren as an arg separator for `color()`:
```js
new Color("color(--lchuv 86.8% 42.9)230)")
```

For reference, this throws:
```js
new Color("color(--lchuv 86.8% 42.9)")
```

Are `!@#$%^&*[]{}(=` legitimate separators? (*that's not necessarily a complete list*)  You can chain as many of those as you like after a number and the coordinates are properly converted.  For example, this works:
```js
new Color("color(--lchuv 86.8{[]}(#!42.9 230)")
```

ScribleSean:
Fixes #614.

`color(--lchuv 86.8% 42.9)230)` currently parses successfully because argument extraction silently discards unmatched punctuation. Reject the function when unmatched content contains anything other than commas or whitespace.

Adds 19 invalid-input regressions and four valid-input controls, including alpha, scientific notation, `none`, and `calc(NaN)`. Also removes an extra closing parenthesis in an existing gamut test's expected value, exposed by the stricter parser. This change is limited to unconsumed punctuation; it does not attempt full CSS grammar validation.

Validation on base `19eb89f8995d2139436b6d38c7042944bb23a718`:
- All 19 new rejection cases fail before the source fix and pass afterward.
- Full suite: 817 passed, 72 existing skipped, zero failures.
- Full build, changed-file ESLint, parser/test formatting and diff whitespace checks pass. The gamut file retains an unrelated pre-existing formatting warning.

Prepared with OpenAI Codex assistance using GPT-6 Astra (`gpt-6-astra`), medium reasoning. I review all contributions before submission and remain responsible for the code and this message. The commit includes Codex co-author attribution.

### DIFF
--- a/src/parse.js
+++ b/src/parse.js
@@ -266,6 +266,11 @@ export function parseFunction (str) {
 			return "";
 		});
 
+		// Argument matching must not silently discard unexpected punctuation.
+		if (/[^,\s]/.test(separators)) {
+			return;
+		}
+
 		return {
 			name,
 			args,
--- a/test/gamut.js
+++ b/test/gamut.js
@@ -273,7 +273,7 @@ export default {
 				},
 				{
 					args: ["color(display-p3 0 1 1)"],
-					expect: "rgb(0% 76.098% 75.455%))",
+					expect: "rgb(0% 76.098% 75.455%)",
 				},
 				{
 					args: ["color(display-p3 1 0 1)"],
--- a/test/parse.js
+++ b/test/parse.js
@@ -11,6 +11,56 @@ const tests = {
 	run: parse,
 	tests: [
 		{
+			name: "Unexpected function argument separators (#614)",
+			throws: TypeError,
+			tests: [
+				{ arg: "color(--lchuv 86.8% 42.9)230)" },
+				{ arg: "color(--lchuv 86.8{[]}(#!42.9 230)" },
+				...[
+					"!",
+					"@",
+					"#",
+					"$",
+					"^",
+					"&",
+					"*",
+					"[",
+					"]",
+					"{",
+					"}",
+					"(",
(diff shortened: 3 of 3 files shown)


src/parse.js
```diff
@@ -266,6 +266,11 @@ export function parseFunction (str) {
 			return "";
 		});
 
+		// Argument matching must not silently discard unexpected punctuation.
+		if (/[^,\s]/.test(separators)) {
+			return;
```
lloydk: **[suggestion]** Rejecting unexpected punctuation with a bare `return` makes `parseFunction` look like a non-function string. `parse()` then walks custom formats (hex/keyword, which cannot match `rgb(...)` / `color(...)`) and throws `Could not parse ${str} as a color. Missing a plugin?`. For the motivating input `color(--lchuv 86.8% 42.9)230)` that message is misleading: `--lchuv` is a prefixed space, and the failure is leftover `)`, not a missing plugin. The same generic error is used for `rgb(0 0 0 /)` (dangling slash) even though the function name and coordinates were already recognized.

**Suggestion:** Throw a syntax `TypeError` here (or have `parse()` special-case “matched `regex.function` but `parseFunction` returned void”) so the message names unexpected argument content instead of a missing plugin, e.g. include the leftover `separators` in the error. Tests only assert `TypeError`, so they would still pass.

